PR #109, opened 2021-08-11
Install rsync on kube-burner container

Harshith-umesh:
### Fixes
Allows copying of pprof folder from the pod onto the machine.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,5 +1,5 @@
 FROM registry.fedoraproject.org/fedora-minimal:latest
-
+RUN microdnf install rsync
 COPY kube-burner /bin/kube-burner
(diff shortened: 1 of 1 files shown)


Containerfile
```diff
@@ -1,5 +1,5 @@
 FROM registry.fedoraproject.org/fedora-minimal:latest
-
+RUN microdnf install rsync
```
rsevilla87, line 2:
```suggestion
RUN microdnf install rsync && rm -Rf /var/cache/yum
```

rsevilla87:
Thanks @Harshith-umesh !, I think is a good idea, since rsync is required for `kubectl rsync` subcommand

PR #133, opened 2021-10-14
Add subtract function for dynamic template

smalleni:
Signed-off-by: Sai Sindhur Malleni <smalleni@redhat.com>


smalleni:
Closing this out as we now have the `sprig` library

PR #148, opened 2022-01-26
Add support to KubeVirt CRD, waiting for the VM/VMI ready condition and collecting detailed latency information 

marceloamaral:
### Description
The popularity of VMs in Kubernetes is growing, sparking great interest in improving performance and scalability from the KubeVirt community. Performing KubeVirt control plane performance analysis is challenging due to the complexity of configuring representative workload settings, measuring performance, and selecting which metrics to consider. In addition, there is no consensus on which benchmarking tool should be used, commonly each project uses its own internal tool.

Therefore, this PR extends kube-burner to support the KubeVirt CRD, waiting for the VM/VMI ready condition and collecting detailed latency information at different stages of the VM/VMI creation process.

This change will be used for performance analysis of OpenShift CNV and probably also on the KubeVirt CI/CD system.

### Key Changes/Improvements
- Changed common structures, moving functions from `pod_latency.go` to `metrics.go` and `watcher.go` to be also used by `vmi_latency.go`.
- `vmi_latency.go` configures the k8s client to understand the KubeVirt CRD to create an Informer.
- Added waiters for kubevirt VM objects in `waiters.go`.
- Created types to store CRD information and the structure to unpack CRD objects, before this structure was in `waiters.go` and it was specific for OpenShift build CRD.
- Added examples for density test creating KubeVirt VMs, key Prometheus metrics and Grafana dashboard.


cmd/kube-burner/kube-burner.go
```diff
@@ -327,6 +327,8 @@ func steps(uuid string, p *prometheus.Prometheus, alertM *alerting.AlertManager)
 		case config.CreationJob:
 			job.Cleanup()
 			measurements.Start()
+			log.Infof("Waiting 10s to initilize the measurement watchers to do not miss any event")
+			time.Sleep(10 * time.Second)
```
ksimon1, line 331: Isn't here a different way how to find out, that everything is ready?

pkg/measurements/metrics/metrics.go
```diff
@@ -0,0 +1,104 @@
...
+	log.Info("Evaluating latency thresholds")
+	for _, phase := range thresholds {
+		for _, pq := range quatiles {
+			if phase.ConditionType == pq.(LatencyQuantiles).QuantileName {
+				// Required to acccess the attribute by name
+				r := reflect.ValueOf(pq.(LatencyQuantiles))
+				v := r.FieldByName(phase.Metric).Int()
+				if v > phase.Threshold.Milliseconds() {
+					log.Warnf("%s %s latency (%dms) higher than configured threshold: %v", phase.Metric, phase.ConditionType, v, phase.Threshold)
+					rc = 1
```
ksimon1, line 95: Do you need to continue in loops after this assignment? If not maybe add break to not continue in loops.

cmd/kube-burner/kube-burner.go
marceloamaral: The watchers synchronize the cache, which should actually do this.
However, it was intermittently having problems for very small tests.
The sleep 10s workaround has fixed the issue for now, but we can investigate this further later.

pkg/measurements/metrics/metrics.go
marceloamaral: Thanks @ksimon1 for the review!

I just moved this code from `pod_latency.go`, so this logic was already there.
Anyway, I think it makes sense to continue in the loop as it will print a warning of all latencies that have been violated.

cmd/kube-burner/kube-burner.go
rsevilla87: @ksimon1 is right, I think we're running into a race condition here because the measurements are triggered in their own goroutine and it's possible that the cache is being synced while kube-burner creates the first objects, so we can miss some events during this time frame.
I think we could create a semaphore to ensure we don't create any object before the cache is synchronised.

I can fix it in a different PR, the change shouldn't be very big.

marceloamaral: In that case, I can remove this logic of sleep 10s from this PR!

pkg/burner/types/types.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+// Condition contains details for the current condition of this pod.
+type Condition struct {
+	Type   string `json:"type"`
+	Status string `json:"status"`
+}
+
+// UnstructuredContent minimum unstructured object content to unmarshal the status phase of a CRD
+type UnstructuredContent struct {
```
rsevilla87, line 38: Good idea :)

pkg/measurements/metrics/metrics.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}
+	return nil
+}
+
+func CheckThreshold(thresholds []types.LatencyThreshold, quatiles []interface{}) int {
+	var rc int
+	log.Info("Evaluating latency thresholds")
+	for _, phase := range thresholds {
+		for _, pq := range quatiles {
```
rsevilla87, line 88:
```suggestion
		for _, pq := range quantiles {
```

rsevilla87:
Hey @marceloamaral, it needs some rebase since I merged the race condition fix.

marceloamaral:
@rsevilla87, I merged the PR.
So, could you please review it again?

rsevilla87 (approved):
LGTM!

PR #152, opened 2022-02-02
Add CNI focused kubelet-density variants

smalleni:
This uses a curl for the readinessprobe of the client pod which can
help measure CNI latency. We also no longer use a postgres
pod on the server side - that is replaced by a simple node.js
app. Also adds networkpolicy.

Signed-off-by: Sai Sindhur Malleni <smalleni@redhat.com>

### DIFF
--- a/examples/workloads/kubelet-density-cni-networkpolicy/kubelet-density-cni-networkpolicy.yml
+++ b/examples/workloads/kubelet-density-cni-networkpolicy/kubelet-density-cni-networkpolicy.yml
@@ -1,0 +1,50 @@
+---
+global:
+  writeToFile: true
+  metricsDirectory: collected-metrics
+  indexerConfig:
+    enabled: true
+    esServers: [http://elastic-elk.apps.rsevilla.kube-burner.com]
+    insecureSkipVerify: true
+    defaultIndex: kube-burner
+    type: elastic
+  measurements:
+    - name: podLatency
+      esIndex: kube-burner
+
+jobs:
+  - name: deny-all-policy
+    jobIterations: 1
+    qps: 1
+    burst: 1
+    namespacedIterations: false
+    namespace: kubelet-density-cni-networkpolicy
+    jobPause: 1m
+    objects:
+
+      - objectTemplate: templates/deny-all.yml
+        replicas: 1
+
+  - name: kubelet-density-cni-networkpolicy
+    jobIterations: 100
+    qps: 25
+    burst: 25
+    namespacedIterations: false
+    namespace: kubelet-density-cni-networkpolicy
+    waitWhenFinished: true
+    podWait: false
+    preloadImages: true
+    preLoadPeriod: 2m
+    objects:
+
+      - objectTemplate: templates/webserver-deployment.yml
+        replicas: 1
+
+      - objectTemplate: templates/webserver-service.yml
+        replicas: 1
+
+      - objectTemplate: templates/allow-http.yml
+        replicas: 1
+
+      - objectTemplate: templates/curl-deployment.yml
+        replicas: 1
--- a/examples/workloads/kubelet-density-cni-networkpolicy/templates/allow-http.yml
+++ b/examples/workloads/kubelet-density-cni-networkpolicy/templates/allow-http.yml
@@ -1,0 +1,12 @@
+kind: NetworkPolicy
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: allow-{{.Replica}}-{{.Iteration}}
+spec:
+  podSelector:
+    matchLabels:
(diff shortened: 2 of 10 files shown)


smalleni:
@rsevilla87 PTAL

smalleni:
Only question/concern I have is networkpolicy might not be propagated fast enough, in which case we need to have that in a separate job and then introduce a jobpause.

examples/workloads/kubelet-density-medium/templates/curl-deployment.yml
```diff
@@ -0,0 +1,40 @@
...
+  template:
+    metadata:
+      labels:
+        name: curl-{{.Replica}}-{{.Iteration}}
+    spec:
+      nodeSelector:
+        node-role.kubernetes.io/worker: ""
+      containers:
+      - name: curlapp
+        image: quay.io/smalleni/curl:latest
```
dry923, line 15: can this be in the cloud-bulldozer quay instead?

examples/workloads/kubelet-density-medium/templates/webserver-deployment.yml
```diff
@@ -0,0 +1,30 @@
...
+  template:
+    metadata:
+      labels:
+        name: webserver-{{.Replica}}-{{.Iteration}}
+    spec:
+      nodeSelector:
+        node-role.kubernetes.io/worker: ""
+      containers:
+      - name: webserver
+        image: quay.io/smalleni/sampleapp:latest
```
dry923, line 15: same as the other. can we move this to cloud-bulldozers quay?

rsevilla87:
> Only question/concern I have is networkpolicy might not be propagated fast enough, in which case we need to have that in a separate job and then introduce a jobpause.

The pods won't be ready until all NetworkPolicies are configured, thus the job will block until all pods are ready. This is the desired behavior right?

smalleni:
> > Only question/concern I have is networkpolicy might not be propagated fast enough, in which case we need to have that in a separate job and then introduce a jobpause.
> 
> The pods won't be ready until all NetworkPolicies are configured, thus the job will block until all pods are ready. This is the desired behavior right?

Raul,

I don't believe OVNK checks for all flows before setting pod to be ready - there's only a few flows in some tables it checks for specifically. Do you know for certain, that it waits for NetworkPolicy to be applied?

rsevilla87:
> > > Only question/concern I have is networkpolicy might not be propagated fast enough, in which case we need to have that in a separate job and then introduce a jobpause.
> > 
> > 
> > The pods won't be ready until all NetworkPolicies are configured, thus the job will block until all pods are ready. This is the desired behavior right?
> 
> Raul,
> 
> I don't believe OVNK checks for all flows before setting pod to be ready - there's only a few flows in some tables it checks for specifically. Do you know for certain, that it waits for NetworkPolicy to be applied?

Right, however kube-burner will wait for the deployments created to be ready, and that happens when its pods pass all readiness probes which requires network policies to be propagated.

smalleni:
> > > > Only question/concern I have is networkpolicy might not be propagated fast enough, in which case we need to have that in a separate job and then introduce a jobpause.
> > > 
> > > 
> > > The pods won't be ready until all NetworkPolicies are configured, thus the job will block until all pods are ready. This is the desired behavior right?
> > 
> > 
> > Raul,
> > I don't believe OVNK checks for all flows before setting pod to be ready - there's only a few flows in some tables it checks for specifically. Do you know for certain, that it waits for NetworkPolicy to be applied?
> 
> Right, however kube-burner will wait for the deployments created to be ready, and that happens when its pods pass all readiness probes which requires network policies to be propagated.

The networkpolicy just allows http. Even before the network policy was applied http would have been allowed, so how do we know if it was appied? I'm wondering if we should have a  job that first does deny all and then apply this networkpolicy along with creating pods in a different yaml?

rsevilla87:
> st does deny all and then apply this networkpolicy along with creatin

oh, right. Didn't understand you at first read. Yeah we can create a deny-job and wait for some time to ensure the deny NP was propagated, not very fancy though there's no way to verify from the client side a networkpolicy has been applied w/o creating a pod.

numansiddique:
Why can't the job (or kube-burner)  first create a network policy - default deny (for both Egress and Ingress)  when the namespaces is created  ?

smalleni:
> Why can't the job (or kube-burner) first create a network policy - default deny (for both Egress and Ingress) when the namespaces is created ?

Yep did that. PTAL

smalleni:
@rsevilla87 @mohit-sheth Let me know if these two variants look good and I can add them to the operator/e2e then.

rsevilla87 (approved):
LGTM!

PR #163, opened 2022-02-22
add access to watchers to get the number of Ready objects without using list or get

marceloamaral:
### Description
The waiters use several List and Get requests which can be very expensive when creating a large number of objects.
For example, in one of my experiments, creating 20k objects in a large cluster, some list request latency took a few minutes at the 99th percentile.

Therefore, List and Get requests from waiters can affect the entire performance of the experiment. Instead of Lists and Gets, we can use watchers to monitor and wait for the state of the desired object.

Fortunately, we already have watcher logic to monitor pod and vm latency. So, we can leverage these watchers to count the number of objects in the desired state and namespace.

I've generalized the logic from the watchers to a package to use for pod and vm latency. Also, the watcher is always initialized when creating a Pod, VM, or VMI object. The watcher maintains a map of resources, namespace, and conditions, which is used by waiters.

### Fixes
Removed List requests when waiting for Pod, VM, or VMI objects.


pkg/measurements/pod_latency.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 The Kube-burner Authors.
+// Copyright 2021 The Kube-burner Authors.
```
RobertKrawitz, line 1: 2022

RobertKrawitz:
I am not familiar enough with the innards of kube-burner to review in a lot of detail, but the approach certainly looks like the right one, so

/lgtm

other than one copyright date that should be fixed.

pkg/measurements/pod_latency.go
marceloamaral: Thanks, It makes sense!
I will update it.

PR #165, opened 2022-02-28
CNI workload: Change the probe from readinessProbe to startupProbe.

numansiddique:
This patch changes the probe in curl-deployment.yml from
readinessProbe to startupProbe for the following reasons:

 - readinessProbe is periodic and with default period interval of 1
   second, it significantly loads kubelet effecting the pod
   readiness as the scale increases.

 - startupProbe is enough to check if CNI has done its job of wiring
   the networking of the pods.

 - Below are the tests on AWS 120 node cluster with 4000 iterations
   of kubelet-density-cni

   With readinessProbe - ContainerReady
   -----------------------------------
    "P99": 80928,
    "P95": 66161,
    "P50": 9526,
    "max": 671045,
    "avg": 19321,

   With startupProbe - ContainerReady
   -----------------------------------
    "P99": 11872,
    "P95": 9609,
    "P50": 4862,
    "max": 18487,
    "avg": 5329,

Signed-off-by: Numan Siddique <nusiddiq@redhat.com>


examples/workloads/kubelet-density-cni-networkpolicy/templates/curl-deployment.yml
```diff
@@ -22,13 +22,15 @@ spec:
...
         securityContext:
           privileged: false
-        readinessProbe:
+        startupProbe:
           exec:
             command: 
               - "/bin/sh"
               - "-c"
-              - "curl ${WEBSERVER_HOSTNAME}:${WEBSERVER_PORT}"
+              - "curl ${WEBSERVER_HOSTNAME}:9090"
```
jtaleric, line 30: any reason you removed `WEBSERVER_PORT` here?

jtaleric (commented):
Small q/nit

examples/workloads/kubelet-density-cni-networkpolicy/templates/curl-deployment.yml
numansiddique: Sorry.  I forgot to clean it up.  I'll rever this change.

numansiddique: Done

jtaleric (approved):
LGTM

examples/workloads/kubelet-density-cni-networkpolicy/templates/curl-deployment.yml
```diff
@@ -22,13 +22,15 @@ spec:
...
           privileged: false
-        readinessProbe:
+        startupProbe:
           exec:
             command: 
               - "/bin/sh"
               - "-c"
               - "curl ${WEBSERVER_HOSTNAME}:${WEBSERVER_PORT}"
           periodSeconds: 1
+          timeoutSeconds: 60
```
smalleni, line 32: Isn't timeoutSeconds too big here? My understanding is that the probe waits for 60s before timing out, perhaps this should be more granular?

numansiddique: I think it can be lowered.  But I am not sure what is the ideal values to have.

In my testing with the original values with readinessProbe,  I noticed that curl was timing out.  So definitely 1 second seems not enough to me.

In my testing I just set it to 60 seconds.  I guess 10 seconds seems reasoble ?  what do you think ?

smalleni: Thinking about it more, it probably doesn't matter all that much. Did you check that the containers never go into ready if the startup probe doesn't succeed?

numansiddique: Yes.  I tested that scenario.  And the containers never go into ready state if the curl is not successful on all retries.

numansiddique:
I can test it out.

Do you know how http probe works ?   If kubelet runs the probe outside of the curl pod (my understanding is that http probe is run by kubelet by not execing to the pod,  I could be wrong though),  then I don't think we will be testing the actual networking from the curl pod to the service and it will be only testing if the server pod is listening on the server port or not.

smalleni:
> I can test it out.
> 
> Do you know how http probe works ? If kubelet runs the probe outside of the curl pod (my understanding is that http probe is run by kubelet by not execing to the pod, I could be wrong though), then I don't think we will be testing the actual networking from the curl pod to the service and it will be only testing if the server pod is listening on the server port or not.

You are right, we are stuck with exec probes then. Exec probes do tend to be expensive for kubelet/cri-o as well.

smalleni (approved):

numansiddique:
We probably could find other ways.

But that would probably require changes in kube-burner.   Maybe each curl pod can log the timestamp when it first successfully curled.  And later kube-burner has to scrape through all the logs and measure the time it took for the curl pod to come up.
This definitely would be very invasive and I'm not sure if it's even  a good idea to worth exploring it :)

Until then "Exec" it is.

PR #185, opened 2022-08-31
Adding privileged namespaces by default to namespaces

paigerube14:
### Description
This is adding all necessary namespace labels to any created namespace by kube-burner
Left the setting of namespace labels from [PR](https://github.com/cloud-bulldozer/kube-burner/pull/184) to after the setting I do so we can overwrite if needed

### Fixes
PSS Issues in 4.12


paigerube14:
going to just go off of the PR Raul added instead of this one

PR #194, opened 2022-10-21
included a var for pause before scraping metrics

mukrishn:
### Description
Adding a new optional argument `--pause` to have a configurable pause before scrapping metrics from prometheus. This helps in hypershift testing when running kube-burner across 100s of hostedcluster at the same time and let it scrape with a delay to keep the prometheus/thanos from responding too many API queries at a time.

### DIFF
--- a/cmd/kube-burner/kube-burner.go
+++ b/cmd/kube-burner/kube-burner.go
@@ -82,6 +82,7 @@ func initCmd() *cobra.Command {
 	var url, metricsProfile, alertProfile, configFile string
 	var username, password, uuid, token, configMap, namespace string
 	var skipTLSVerify bool
+	var scrapePause time.Duration
 	var prometheusStep time.Duration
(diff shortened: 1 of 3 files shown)


rsevilla87:
> ### Description
> 
> Adding a new optional argument `--pause` to have a configurable pause before scrapping metrics from prometheus. This helps in hypershift testing when running kube-burner across 100s of hostedcluster at the same time and let it scrape with a delay to keep the prometheus/thanos from responding too many API queries at a time.
> ### Fixes

Hey @mukrishn!, thanks for the contribution, however I think this flag is not necessary as you can replicate a similar behavior with the parameter `jobPause`, documented at https://kube-burner.readthedocs.io/en/latest/configuration/#jobs

What do you think?

mukrishn:
@rsevilla87 yes you are right, `jobPause` is working as expected, so far no issue with that. I will re-open if needed.

Issue #195, opened 2022-10-27
[RFE] kube-burner openshift wrapper


rsevilla87:
**Is your feature request related to a problem? Please describe.**
Having an OpenShift wrapper for kube-burner will simplify and make more stable the process of running workloads in OpenShift based clusters

PR #199, opened 2022-11-07
Openshift wrapper

rsevilla87:
### Description

Very opinionated wrapper meant to simplify the way we trigger kube-burner against OpenShift clusters.

Tasks:

- [x] Code base as a separate binary: kube-burner-ocp
- [x] Merge kube-burner-ocp into a new `kube-burner ocp'  subcommand
- [x] node-density logic (calculate number of pods to deploy across the cluster)
- [x] Prometheus endpoint and token automatic discovery
- [x] Integrated three basic workloads. node-density, cluster-density and node-density-heavy
- [x] Better integration in makefile
- [x] Documentation
- [x] CI tests

Example output:

```shell
rsevilla@wonderland ~/labs/kube-burner (openshift-wrapper) $  bin/amd64/kube-burner-ocp -h

        This tool is a kube-burner wrapper meant to be used against OpenShift clusters and serve as a shortcut to trigger well-known workloads

Usage:
  kube-burner-ocp [command]

Available Commands:
  cluster-density    Runs cluster-density workload
  completion         Generates completion scripts for bash shell
  help               Help about any command
  node-density       Runs node-density workload
  node-density-heavy Runs node-density-heavy workload
  version            Print the version number of kube-burner-ocp

Flags:
      --burst int          Kube-burner Burst (default 20)
      --es-index string    Elastic Search index
      --es-server string   Elastic Search endpoint
  -h, --help               help for kube-burner-ocp
      --indexing           Elastic Search indexing (default true)
      --log-level string   Kube-burner log level (default "info")
      --qps int            kube-burner QPS (default 20)
      --uuid string        Benchmark UUID (default "3ebef78a-48d9-4e87-9d74-237cf585c0d2")

Use "kube-burner-ocp [command] --help" for more information about a command.

```
- Fixes: #195


jtaleric:
There has been some discussions wrt node-density-heavy and the relevance of that workload. I haven't done nearly the investigations as @afcollins so I will let him speak to if he feels we need to either change the workload or dismiss it totally.

pkg/discovery/discovery.go
```diff
@@ -0,0 +1,197 @@
...
+		if err != nil {
+			return podCount, err
+		}
+		podCount += len(podList.Items)
+	}
+	return podCount, nil
+}
+
+// GetInfraDetails returns cluster anme and platform
+func GetInfraDetails() (infraObj, error) {
```
jtaleric, line 112: This always bites me too.. s/G/g

jtaleric (commented):
Really excited about the Discovery work here to reduce the need to provide prom data... Great improvement!

afcollins:
>There has been some discussions wrt node-density-heavy and the relevance of that workload.

node-density-heavy is a fine workload, if used correctly. I think it is unique from others and serves its role as one workload option for us with regular probes and pod-to-svc communication.  My issue is in trying to "black-box" run node-density-heavy at 245ppn and expect it will not wreak havoc on a cluster when the only place I've seen it run alright is on the beefiest bare metal servers we have.

afcollins:
I usually run kube-burner workloads tweaking some of the config parameters (e.g. PODS_PER_NODE/JOB_ITERATIONS, QPS/BURST, PRELOAD_IMAGES). Does this allow me to do that? env parameters running through e2e-benchmarking are what I currently use to achieve that.

jtaleric:
> I usually run kube-burner workloads tweaking some of the config parameters (e.g. PODS_PER_NODE/JOB_ITERATIONS, QPS/BURST, PRELOAD_IMAGES). Does this allow me to do that? env parameters running through e2e-benchmarking are what I currently use to achieve that.

The idea of this is to reduce the complexity of modifying the vars. If you need do modify vars you can just the traditional kube-burner cli. IMHO

jtaleric:
> > There has been some discussions wrt node-density-heavy and the relevance of that workload.
> 
> node-density-heavy is a fine workload, if used correctly. I think it is unique from others and serves its role as one workload option for us with regular probes and pod-to-svc communication. My issue is in trying to "black-box" run node-density-heavy at 245ppn and expect it will not wreak havoc on a cluster when the only place I've seen it run alright is on the beefiest bare metal servers we have.

With that said, does it make sense to have a more sane default here? https://github.com/cloud-bulldozer/kube-burner/pull/199/files#diff-0d30f4f9accca6d61345de555771e02724733c29b1d6adf78f8c2530b16092c8R47 

Again, the idea here is to reduce the params the user has to pass, so the more sane the defaults the better. 

I wouldn't think we want to assume either extreme of the worker size (meaning, don't assume we are testing baremetal or super small CI instances).

rsevilla87:
> I usually run kube-burner workloads tweaking some of the config parameters (e.g. PODS_PER_NODE/JOB_ITERATIONS, QPS/BURST, PRELOAD_IMAGES). Does this allow me to do that? env parameters running through e2e-benchmarking are what I currently use to achieve that.

Yes, this wrapper allows to modify via flags some of those parameters that are often tweaked. 

i.e, with the latest approach, where the wrapper is integrated within the original kube-burner binary, looks like:
```shell
rsevilla@wonderland ~/labs/kube-burner (openshift-wrapper) $ bin/amd64/kube-burner ocp node-density -h
Runs node-density workload

Usage:
  kube-burner ocp node-density [flags]

Flags:
      --container-image string         Container image (default "gcr.io/google_containers/pause:3.1")
  -h, --help                           help for node-density
      --pod-ready-threshold duration   Pod ready timeout threshold (default 5s)
      --pods-per-node int              Pods per node (default 245)

Global Flags:
      --burst int          Burst (default 20)
      --es-index string    Elastic Search index
      --es-server string   Elastic Search endpoint
      --indexing           Enable Elastic Search indexing (default true)
      --log-level string   Allowed values: trace, debug, info, warn, error, fatal (default "info")
      --qps int            QPS (default 20)
      --uuid string        Benchmark UUID (default "ef63a981-9213-4525-bd13-988f3883ad63")

```

config/node-density-heavy/app-deployment.yml
```diff
@@ -0,0 +1,61 @@
...
+          timeoutSeconds: 60
+          initialDelaySeconds: 30
+        livenessProbe:
+          httpGet:
+            path: /health
+            port: 8080
+          periodSeconds: {{ .livenessPeriod }}
+          failureThreshold: 1
+          timeoutSeconds: 15
+          initialDelaySeconds: 30
```
afcollins, line 32: I have tweaked all these probe timeouts in cni and heavy workloads in one environment based on what the servers can handle. Might be useful to at least allow overrides.  However, I see how that could quickly lead to parameter explosion for niche use-cases that might be better served simply modifying the config files.

config/cluster-density/cluster-density.yml
```diff
@@ -0,0 +1,57 @@
...
+jobs:
+  - name: cluster-density
+    namespace: node-density
+    jobIterations: {{.JOB_ITERATIONS}}
+    qps: {{.QPS}}
+    burst: {{.BURST}}
+    namespacedIterations: true
+    podWait: false
+    waitWhenFinished: true
+    preLoadImages: true
```
afcollins, line 22: Having preLoadImages as a variable is helpful and I use it frequently.

config/node-density-heavy/node-density-heavy.yml
```diff
@@ -0,0 +1,44 @@
...
+        - conditionType: Ready
+          metric: P99
+          threshold: {{.POD_READY_THRESHOLD}}
+jobs:
+  - name: node-density-heavy
+    namespace: node-density-heavy
+    jobIterations: {{.JOB_ITERATIONS}}
+    qps: {{.QPS}}
+    burst: {{.BURST}}
+    namespacedIterations: false
```
afcollins, line 23: having a variable for namespacedIterations can be helpful

config/node-density-heavy/app-deployment.yml
rsevilla87: That's correct, I added this parameter as flag for now though, we can add the future PRs if really needed

PR #231, opened 2022-12-13
Churning improvements

rsevilla87:
### Description

Improve churning code and increase churning delay
The default churning delay of 30s is very short, the cluster-density workload on this repo is more complex in terms of networking requirements, and needs a higher value (2 mins) to prevent excessive SDN delays.


dry923 (approved):
LGTM!

PR #232, opened 2022-12-15
Cluster-density network-policies flag and increase default timeout

rsevilla87:
### Description

Add flag to opt-out network policies from the cluster-density workload
Pass the `--network-policies=false` argument to disable them

And increase default timeout to 4h

### DIFF
--- a/cmd/kube-burner/kube-burner.go
+++ b/cmd/kube-burner/kube-burner.go
@@ -138,7 +138,7 @@ func initCmd() *cobra.Command {
 	cmd.Flags().StringVarP(&alertProfile, "alert-profile", "a", "", "Alert profile file or URL")
 	cmd.Flags().BoolVar(&skipTLSVerify, "skip-tls-verify", true, "Verify prometheus TLS certificate")
 	cmd.Flags().DurationVarP(&prometheusStep, "step", "s", 30*time.Second, "Prometheus step size")
-	cmd.Flags().DurationVarP(&timeout, "timeout", "", 2*time.Hour, "Benchmark timeout")
+	cmd.Flags().DurationVarP(&timeout, "timeout", "", 4*time.Hour, "Benchmark timeout")
 	cmd.Flags().StringVarP(&configFile, "config", "c", "", "Config file path or URL")
 	cmd.Flags().StringVarP(&configMap, "configmap", "", "", "Configmap holding all the configuration: config.yml, metrics.yml and alerts.yml. metrics and alerts are optional")
 	cmd.Flags().StringVarP(&namespace, "namespace", "", "default", "Namespace where the configmap is")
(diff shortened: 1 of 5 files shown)


cmd/kube-burner/kube-burner.go
```diff
@@ -138,7 +138,7 @@ func initCmd() *cobra.Command {
 	cmd.Flags().StringVarP(&alertProfile, "alert-profile", "a", "", "Alert profile file or URL")
 	cmd.Flags().BoolVar(&skipTLSVerify, "skip-tls-verify", true, "Verify prometheus TLS certificate")
 	cmd.Flags().DurationVarP(&prometheusStep, "step", "s", 30*time.Second, "Prometheus step size")
-	cmd.Flags().DurationVarP(&timeout, "timeout", "", 2*time.Hour, "Benchmark timeout")
+	cmd.Flags().DurationVarP(&timeout, "timeout", "", 4*time.Hour, "Benchmark timeout")
```
jtaleric, line 141: If we want to have a strict timeout, possibly the user manages this outside of kube-burner? The issue with a hard timeout is the user passes the `--iterations` which could be 4000, but if they set `qps` and `burst` to 1 it might take > 4 hours and they might not consider the hard timeout, so they just lost 4 hours of testing.

jtaleric (commented):
Since this CLI is more catered to OCP, we should consider breaking each of these workloads and timeouts out... 

If the user overrides any of the parameters that can impact the time it takes to execute the test, kube-burner spits out a warning that the timeout might not be sufficient... but we provide a timeout that we know will be OK for the default params. 

We should consider having multiple options with cluster-density...
- `cluster-density`
- `cluster-density-network-pol`
- `cluster-density-churn`
- `cluster-density-network-pol-churn`

Thoughts?

cmd/kube-burner/kube-burner.go
rsevilla87: That's a good point, we can have some mechanism to "automatically calculate a proper timeout" if the user doesn't pass it.

It's not simple, because it would depend on several factors like the number of objects to create, churning, etc.

rsevilla87:
> Since this CLI is more catered to OCP, we should consider breaking each of these workloads and timeouts out...
> 
> If the user overrides any of the parameters that can impact the time it takes to execute the test, kube-burner spits out a warning that the timeout might not be sufficient... but we provide a timeout that we know will be OK for the default params.
> 
> We should consider having multiple options with cluster-density...
> 
>     * `cluster-density`
> 
>     * `cluster-density-network-pol`
> 
>     * `cluster-density-churn`
> 
>     * `cluster-density-network-pol-churn`
> 
> 
> Thoughts?

This is a different and easier approach than the one I proposed above. Sounds reasonable, I think having 4 different entrypoints won't be necessary though

PR #249, opened 2023-02-03
Delete namespaces with timeout

rsevilla87:
### Description
With the old implementation, when a job times out the created namespaces are not garbage collected.


afcollins (approved):
LGTM. Thank you!

PR #257, opened 2023-02-22
Cluster density v1

rsevilla87:
### Description

Adds cluster-density v1 and reduces code by adding the extract functionality to the parent PersistentPreRun.

cc: @paigerube14 @jtaleric


jtaleric (approved):
lgtm! Glad we are making the distention. Not going to block this, but I am curious if we should have churn disabled (by default)  with v1?

rsevilla87:
> lgtm! Glad we are making the distention. Not going to block this, but I am curious if we should have churn disabled (by default) with v1?

We're already running our tests with churning enabled by default, so I think that should be the default.

PR #258, opened 2023-02-28
Index kube-burner version information along with workload metadata info

rsevilla87:
### Description

Index kube-burner version information along with workload metadata info


jtaleric (approved):
lgtm

PR #272, opened 2023-03-14
Fix OCP command for 1.4.3

mtodor:
### Description

This PR is fixing two problems discovered in version 1.4.3.

1. Adding indexer option to define if indexer should be used or not (default: false)
2. Setting correct metrics configuration file

### Fixes

**1. Local indexer**

If we are using a local indexer, the problem is that the indexer instance will not be created, and `kube-burner` will panic.
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x230f9b4]

goroutine 78 [running]:
github.com/cloud-bulldozer/kube-burner/pkg/burner.indexjobSummaryInfo(0x0, {0x7ff7bfeff635, 0x38}, 0x4040ede1d2178f69, {0x14, 0x0, 0x0, {0xc00057ced0, 0xf}, {0xc0003bc000, ...}, ...}, ...)
	/home/runner/work/kube-burner/kube-burner/pkg/burner/metadata.go:51 +0x214
github.com/cloud-bulldozer/kube-burner/pkg/burner.Run.func1()
	/home/runner/work/kube-burner/kube-burner/pkg/burner/job.go:168 +0x1449
created by github.com/cloud-bulldozer/kube-burner/pkg/burner.Run
	/home/runner/work/kube-burner/kube-burner/pkg/burner/job.go:87 +0x3ed
```

The reason for this problem is that usage of the indexer is determined by the existence of the Elastic endpoint option. And that option does not exist for local indexer.

The proposed solution is an independent command option that allows us to enable/disable the indexer per test execution.

**2. Empty metrics profile**
 
With some changes in 1.4.3, information about the default metrics profile is not propagated, and the metrics profile is always empty.

Added changes will propagate the metrics profile provided in a function call.

### DIFF
--- a/cmd/kube-burner/ocp.go
+++ b/cmd/kube-burner/ocp.go
@@ -41,6 +41,7 @@ func openShiftCmd() *cobra.Command {
 	var wh workloads.WorkloadHelper
 	esServer := ocpCmd.PersistentFlags().String("es-server", "", "Elastic Search endpoint")
 	esIndex := ocpCmd.PersistentFlags().String("es-index", "", "Elastic Search index")
+	indexing := ocpCmd.PersistentFlags().Bool("indexing", true, "Enable indexing")
 	metricsEndpoint := ocpCmd.PersistentFlags().String("metrics-endpoint", "", "YAML file with a list of metric endpoints")
 	alerting := ocpCmd.PersistentFlags().Bool("alerting", true, "Enable alerting")
 	uuid := ocpCmd.PersistentFlags().String("uuid", uid.NewV4().String(), "Benchmark UUID")
@@ -53,17 +54,16 @@ func openShiftCmd() *cobra.Command {
 	ocpCmd.MarkFlagsRequiredTogether("es-server", "es-index")
 	ocpCmd.PersistentPreRun = func(cmd *cobra.Command, args []string) {
 		rootCmd.PersistentPreRun(cmd, args)
-		indexing := *esServer != ""
 		envVars := map[string]string{
 			"ES_SERVER": strings.TrimSuffix(*esServer, "/"),
 			"ES_INDEX":  *esIndex,
 			"QPS":       fmt.Sprintf("%d", *qps),
 			"BURST":     fmt.Sprintf("%d", *burst),
-			"INDEXING":  fmt.Sprintf("%v", indexing),
+			"INDEXING":  fmt.Sprintf("%v", *indexing),
 			"GC":        fmt.Sprintf("%v", *gc),
 		}
 		discoveryAgent := discovery.NewDiscoveryAgent()
-		wh = workloads.NewWorkloadHelper(envVars, *alerting, OCPConfig, discoveryAgent, indexing, *timeout, *metricsEndpoint, *userMetadata)
+		wh = workloads.NewWorkloadHelper(envVars, *alerting, OCPConfig, discoveryAgent, *indexing, *timeout, *metricsEndpoint, *userMetadata)
 		wh.Metadata.UUID = *uuid
 		if *extract {
 			if err := wh.ExtractWorkload(cmd.Name(), workloads.MetricsProfileMap[cmd.Name()]); err != nil {
--- a/cmd/kube-burner/ocp.go
+++ b/cmd/kube-burner/ocp.go
@@ -41,7 +41,7 @@ func openShiftCmd() *cobra.Command {
 	var wh workloads.WorkloadHelper
 	esServer := ocpCmd.PersistentFlags().String("es-server", "", "Elastic Search endpoint")
 	esIndex := ocpCmd.PersistentFlags().String("es-index", "", "Elastic Search index")
-	indexing := ocpCmd.PersistentFlags().Bool("indexing", false, "Enable indexing")
+	indexingFlag := ocpCmd.PersistentFlags().Bool("indexing", false, "Enable indexing")
 	metricsEndpoint := ocpCmd.PersistentFlags().String("metrics-endpoint", "", "YAML file with a list of metric endpoints")
 	alerting := ocpCmd.PersistentFlags().Bool("alerting", true, "Enable alerting")
 	uuid := ocpCmd.PersistentFlags().String("uuid", uid.NewV4().String(), "Benchmark UUID")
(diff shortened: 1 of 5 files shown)


cmd/kube-burner/ocp.go
```diff
@@ -41,6 +41,7 @@ func openShiftCmd() *cobra.Command {
 	var wh workloads.WorkloadHelper
 	esServer := ocpCmd.PersistentFlags().String("es-server", "", "Elastic Search endpoint")
 	esIndex := ocpCmd.PersistentFlags().String("es-index", "", "Elastic Search index")
+	indexing := ocpCmd.PersistentFlags().Bool("indexing", false, "Enable indexing")
```
jtaleric, line 44: we should not change the default behavior of `indexing` into Elastic Search, so I would recommend the default behavior to be true here.

test/run-ocp.sh
```diff
@@ -16,7 +16,7 @@ die() {
 UUID=$(uuidgen)
 ES_SERVER="https://search-perfscale-dev-chmf5l4sh66lvxbnadi4bznl3a.us-west-2.es.amazonaws.com/"
 ES_INDEX="kube-burner-ocp"
-COMMON_FLAGS="--es-server=${ES_SERVER} --es-index=${ES_INDEX} --alerting=true --uuid=${UUID} --qps=5 --burst=5"
+COMMON_FLAGS="--indexing=true --es-server=${ES_SERVER} --es-index=${ES_INDEX} --alerting=true --uuid=${UUID} --qps=5 --burst=5"
```
jtaleric, line 19: We should have a test for local and es.

pkg/workloads/helpers.go
```diff
@@ -221,6 +221,7 @@ func (wh *WorkloadHelper) run(workload, metricsProfile string) {
 		metricsEndpoints = append(metricsEndpoints, prometheus.MetricEndpoint{
 			Endpoint:     wh.prometheusURL,
 			AlertProfile: alertsProfile,
+			Profile:      metricsProfile,
```
jtaleric, line 224: I am not clear why this is needed?

jtaleric (changes requested):
Recommend we flip to assume we are indexing (true), and I am not clear on the `metricProfile` change

cmd/kube-burner/ocp.go
mtodor: Unfortunately, this is not possible. The reason for that is that currently, indexing is implicitly enabled if an elastic server is provided.

If we introduce an explicit flag that will enable indexing and set it to true by default, the basic command (`kube-burner ocp cluster-density --iterations=1`) will fail if an elastic server is not provided.

Suppose we want to keep the previous functionality and support the local indexer. In that case, we must keep implicit indexing based on the elastic server and explicit indexing flag that is not set by default.

I have changed the implementation to apply that.

Just one consideration. If you look at that solution from the user's perspective, it's very confusing.

pkg/workloads/helpers.go
mtodor: Let me try to explain.

We can set the metrics profile by providing a metrics endpoint. An example: https://github.com/cloud-bulldozer/kube-burner/blob/master/test/metrics-endpoints.yaml - and **this will work for the OCP command**.

If we don't provide metrics endpoint configuration, then the default metrics profile will be used. That is defined in this map: https://github.com/cloud-bulldozer/kube-burner/blob/master/pkg/workloads/workloads.go

This matrics profile will be propagated when the workload is executed: https://github.com/cloud-bulldozer/kube-burner/blob/a575df584a6b520a45e2fe7903e608a34e722e5f/pkg/workloads/cluster-density.go#L42

Now if we look at this part of the code from line 218 to 226: https://github.com/cloud-bulldozer/kube-burner/blob/a575df584a6b520a45e2fe7903e608a34e722e5f/pkg/workloads/helpers.go#L218

You can see that if the metrics endpoint is not provided, we will create a "default" metrics endpoint, but that default metrics endpoint will have empty metrics profile. And that metrics profile will be used in line 230:
https://github.com/cloud-bulldozer/kube-burner/blob/a575df584a6b520a45e2fe7903e608a34e722e5f/pkg/workloads/helpers.go#L230

To use default metrics profiled defined for a workload, we have to use the value provided in the function call.

So the problem is that, with some changes in 1.4.3, information about the default metrics profile is not propagated, and the metrics profile is always empty.

And added changes will propagate the metrics profile provided in a function call.

test/run-ocp.sh
mtodor: Added test for local indexer. It's simple and checks only if the tar file exists in the end.

cmd/kube-burner/ocp.go
jtaleric: Hm - I would think we could add a check to see if `--indexing` is not passed and `--es-server` is empty, it doesn't attempt to run the workload, noting that `--es-server` is required if `--indexing` isn't set to false?

mtodor: I don't understand what exactly you mean here. Do you want to simply stop execution if both of these options are not provided?

You also have to consider that `es-server` can be provided in the workload configuration directly. There is also a flag for indexing. But that is ignored for the `ocp` command because the decision of whether indexing is enabled or not is defined before workload configuration is used.

I would suggest major refactoring for the `ocp` command. We should use provided workload configuration as a base, and if some option is provided via the command option, then that option will override the value provided in the configuration. Kick out using environment variables because they are just adding unnecessary complexity. But that is out of scope here and what we are trying to do is just fix the usage of local indexer.

jtaleric: I wonder if how you are using the `ocp-wrapper` how it isn't intended to be used? Maybe you might be better off running workloads outside of the `ocp-wrapper`?

pkg/workloads/helpers.go
rsevilla87: Good catch, this is a bug, have to add a test case to detect it

rsevilla87 (changes requested):
Hey @mtodor, wrt to the _indexing_ flag I understand your point of view, that flag is needed in case we want to enable _local_ indexing, however it will affect the current command execution and will require an extra flag that is not needed at the moment as @jtaleric suggests.

I think we can address the issue by replacing the _indexing_ flag by a _local-indexing_  flag, that would mean that once this flag is passed, the OCP wrapper would assume that Elastic indexing is disabled and enable local indexing, thoughts?

On the other hand, I cherry-picked some of the changes on this PR to fix the empty metrics-profile bug to and opened https://github.com/cloud-bulldozer/kube-burner/pull/274  so we can have it merged sooner.

mtodor:
@rsevilla87 Tnx for taking the fix and merging it.

I have switched to the `init` command instead of `ocp`. It's possible to go around the problem for the local indexer by using trivial values `--es-index=test --es-server=test` - to pass the check for both options, and that indexing is enabled.

Both proposed solutions with new options additionally complicate the command. If we can come up with a generic clean solution, then we should implement it.

One idea I had was to have all options for `indexerConfig` filled. In example:
```
indexerConfig:
  enabled: {{.INDEXING}}
  type: {{.INDEXER_TYPE}}
  esServers: ["{{.ES_SERVER}}"]
  insecureSkipVerify: true
  defaultIndex: {{.ES_INDEX}}
  metricsDirectory: collected-metrics
  createTarball: true
  tarballName: collected-metrics.tar.gz
```

An to create a new option in the `ocp` command `--indexer-type`. By default, it will be `elastic`. Possible options could be:
- elastic
- local
- none

For `none` - indexing is disabled.

In that case, you can directly run the `ocp` command with elastic/local/no-indexing without any modification on configuration.

If we are not sure about the solution yet, then we can open an issue in GitHub to keep track and have it documented if someone else encounters a similar issue. And then get to it when we have better ideas.

rsevilla87:
@mtodor , I just merged #278 that should address the issue you had with local-indexing with the OCP wrapper, let me know if it works for you, thanks!

mtodor:
@rsevilla87 Awesome work! I looked at PR, and it addresses the issue with the local indexer of `ocp` outlined here and beyond that. Tnx!

I have switched to using the `init` command. So, I'm not sure when and if I'll test it. But if you want to ensure that option works, you can add one test similar to what I have added in `test/run-ocp.sh` for this PR.

Thanks once again! I'll close this PR because all problems are addressed in other PRs.

rsevilla87:
> @rsevilla87 Awesome work! I looked at PR, and it addresses the issue with the local indexer of `ocp` outlined here and beyond that. Tnx!
> 
> I have switched to using the `init` command. So, I'm not sure when and if I'll test it. But if you want to ensure that option works, you can add one test similar to what I have added in `test/run-ocp.sh` for this PR.
> 
> Thanks once again! I'll close this PR because all problems are addressed in other PRs.

Great!, glad it worked. I'm not updating the tests very much lately, the current test framework is not very reliable, it needs a full refactor as I noticed at https://github.com/cloud-bulldozer/kube-burner/issues/252

Anyhow, I'll take into account your test case once I start with this refactor.

PR #273, opened 2023-03-14
Add prometheus related metrics to ocp wrapper

smalleni:
With every release we seem to be adding more metrics and as a result more timeseries, samples and datapoints per minute to prometheus. It would be beneficial to maintain this data for historical comparisons and RCAing prometheus issues.


rsevilla87 (approved):
thanks!, lgtm

PR #293, opened 2023-04-13
Early garbage collection

rsevilla87:
### Description

Initialize namespace garbage collection in the background before collecting metrics. This should help to save some benchmark time as the namespace cleanup would run in parallel with the metrics collection


pkg/burner/job.go
```diff
@@ -113,7 +113,7 @@ func Run(configSpec config.Spec, uuid string, prometheusClients []*prometheus.Pr
 				if job.Config.Cleanup {
 					ctx, cancel := context.WithTimeout(context.Background(), timeout)
 					defer cancel()
-					CleanupNamespaces(ctx, job.selector.ListOptions)
+					CleanupNamespaces(ctx, job.selector.ListOptions, true)
```
vishnuchalla, line 116: Why did we decide to wait until all the namespaces to get deleted. Why aren't we just triggering the deletion and moving forward. Same for all the `cleanupWait: true` cases?

vishnuchalla:
Code changes look good to me, but why are we deleting some namespaces asynchronously and some synchronously? Why not do all the deletions in asynchronous manner? Is there any breaking dependency?

pkg/burner/job.go
rsevilla87: The original implementation waits for all namespaces to be delete before moving forward/finishing the benchmark. I added this flag just to remove them asynchronously before initialize metric/alert collection/indexing, so we the namespaces are being deleted in the background while these tasks are running in foreground.

pkg/burner/job.go
```diff
@@ -165,11 +165,9 @@ func Run(configSpec config.Spec, uuid string, prometheusClients []*prometheus.Pr
 				indexjobSummaryInfo(indexer, uuid, elapsedTime, job.Config, job.Start, metadata)
 			}
 		}
-		// Update end time of last job
-		jobList[len(jobList)-1].End = time.Now().UTC()
-		if len(prometheusClients) > 0 {
-			log.Infof("Waiting %v extra before scraping prometheus endpoints", prometheusClients[0].Step)
-			time.Sleep(prometheusClients[0].Step)
+		// We initialize cleanup as soon as the benchmark finishes
+		if configSpec.GlobalConfig.GC {
```
rsevilla87, line 169: @vishnuchalla  note that cleanupWait is false here

pkg/burner/job.go
```diff
@@ -165,11 +165,9 @@ func Run(configSpec config.Spec, uuid string, prometheusClients []*prometheus.Pr
...
+			go CleanupNamespaces(context.TODO(), v1.ListOptions{LabelSelector: fmt.Sprintf("kube-burner-uuid=%v", uuid)}, false)
```
rsevilla87, line 170: @vishnuchalla note that cleanupWait is false here

vishnuchalla: Yes, Understood. My thought was, why can't we have it `false` for all the remaining `cleanupWait: true` cases as well?

rsevilla87: For design reasons I wanted to wait for cleanup to finish before exiting kube-burner, otherwise we would have to verify that all the created assets have been pruned externally (oc or kubectl).

rsevilla87: Also, when churning is enabled, we must wait for the namespaces to be gone before re-creating them.

vishnuchalla: Make sense. Thanks for the clarification.

vishnuchalla (approved):

PR #295, opened 2023-04-14
Log file and line number

rsevilla87:

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -15,7 +15,10 @@
 package log
 
 import (
+	"fmt"
 	"os"
+	"path"
+	"runtime"
 
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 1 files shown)


pkg/log/log.go
```diff
@@ -0,0 +1,22 @@
+package log
```
vishnuchalla, line 1: Having a separate `log` package for this small `init` function doesn't seem right. Since it is only being used by `kube-buner.go` as of now, we can simply have this as a private method in that file and in a long term we can simply move it to `go-commons`. Please let me know your thoughts on this. Thanks!

rsevilla87: Just moved the logic to `kube-burner.go` and removed the log package

vishnuchalla: LGTM!

vishnuchalla (approved):

PR #299, opened 2023-04-26
Implements Iterations per Namespace

### LINKED ISSUE #201: [RFE] Allow configurable namespace buckets
kube-burner currently has a flag "namespacedIterations", which is a boolean that switches between only 1) all-in-one namespace for all workload iterations or 2) a distinct namespace for each iteration.
There is control plane overhead with creating a namespace, and this also means the creation rate of resources is rate-limited by the control plane, rather than the client QPS settings of kube-burner.

I would like a knob that allows me to batch iterations into a fixed number of namespaces.
Could be implemented in many ways, but two ways to implement the knob might be: 1) Specifying a desired number of namespaces that the iterations are created in, or 2) Specifying the number of iterations per namespace to create.

This increased granularity added to namespacedIterations would also allow us to expand our understanding of cluster limits beyond the limits of a single namespace to the effects of multiple namespaces at the identified limits.

Extra credit:
Also, I may not always want to include the namespace creation in the workload measurement period.
There could be a separate flag to "prime" the namespaces, such that kube-burner creates all namespaces at once up front, then creates the workloads at the desired QPS rate.
This improvement allows us to have more desirable scale rate than what is currently limited by the instantiation time of a namespace.

afcollins:
### Description
Adds a new field that allows further granularity of the namespacedIterations boolean field by specifying a cap on number of iterations in a single namespace before creating a new namespace.

This solves the 5k service-per-namespace limitation and also allows us to minimize compute overhead, like node-density-heavy@245ppn, which does not run smoothly on an 8-core machine in a single namespace, but purely by setting iterationsPerNamespace=1000, the CPU demand on a node drops from 12c to 3c.

### Fixes
#201


afcollins:
I know it needs documentation, but I want to get the code itself reviewed. I have been testing it and it works reliably with and without namespacedIterations.

cc: @rsevilla87

pkg/burner/namespaces.go
```diff
@@ -38,6 +38,7 @@ func createNamespace(clientset *kubernetes.Clientset, namespaceName string, nsLa
 			return false, err
 		}
 		if errors.IsAlreadyExists(err) {
+			// Getting unnecessary logs here. Would need to implement a GET first so we aren't issuing creates. Should just use an in-memory cache for creation.
```
rsevilla87, line 41: Why are you seeing too many logs here? I assume the `namespacesCreated` logic takes care of only calling this function when the namespace is not present in the map

rsevilla87 (commented):
I like the approach @afcollins, it's simple and effective.

pkg/burner/namespaces.go
afcollins: Oh, that was a note to myself to find where to make changes. Using a map means the logs don't happen. I can remove this comment. Thanks!

pkg/burner/create.go
```diff
@@ -38,6 +38,11 @@ import (
 	"k8s.io/client-go/dynamic"
 )
 
+var (
+	namespacesCreated = make(map[string]bool)
+	namespacesWaited  = make(map[string]bool)
+)
+
```
rsevilla87, line 45: I think the tests are failing b/c one the RunCreateJob function finishes you need to flush these maps (some of the tests have churning enabled, so when it tries to recreate a churn-ed namespace it doesn't create anything since it already exists in this map)

Best solution is to move these maps as a local variable from `RunCreateJob`

pkg/burner/create.go
```diff
@@ -123,18 +128,29 @@ func (ex *Executor) RunCreateJob(iterationStart, iterationEnd int) {
...
+					continue
+				}
+				namespacesCreated[ns] = true
 			}
 		}
 		for objectIndex, obj := range ex.objects {
 			ex.replicaHandler(objectIndex, obj, ns, i, &wg)
 		}
 		if ex.Config.PodWait {
+			// TODO Only need one waiter per namespace, not one per namespace per iterations.
```
rsevilla87, line 144: I think you can remove this comment as well :)

pkg/burner/create.go
```diff
@@ -123,18 +128,29 @@ func (ex *Executor) RunCreateJob(iterationStart, iterationEnd int) {
 		}
 		log.Debugf("Creating object replicas from iteration %d", i)
 		if ex.Config.NamespacedIterations {
-			ns = fmt.Sprintf("%s-%d", ex.Config.Namespace, i)
-			if err = createNamespace(ClientSet, ns, nsLabels); err != nil {
-				log.Error(err.Error())
-				continue
+			ns = fmt.Sprintf("%s-%d", ex.Config.Namespace, i/ex.Config.IterationsPerNamespace)
```
rsevilla87, line 131: Can you add a simple comment explaining the logic of this? I think it can be useful in the future

pkg/burner/create.go
```diff
@@ -149,7 +165,12 @@ func (ex *Executor) RunCreateJob(iterationStart, iterationEnd int) {
 		sem := make(chan int, int(ClientSet.RESTClient().GetRateLimiter().QPS())*2)
 		for i := iterationStart; i <= iterationEnd; i++ {
 			if ex.Config.NamespacedIterations {
-				ns = fmt.Sprintf("%s-%d", ex.Config.Namespace, i)
+				ns = fmt.Sprintf("%s-%d", ex.Config.Namespace, i/ex.Config.IterationsPerNamespace)
+				if namespacesWaited[ns] {
```
vishnuchalla, line 169: do we need an `else` here?
```suggestion
				if namespacesWaited[ns] {
				continue
				}
				namespacesWaited[ns] = true
```

pkg/burner/create.go
```diff
@@ -123,18 +128,29 @@ func (ex *Executor) RunCreateJob(iterationStart, iterationEnd int) {
...
+					log.Error(err.Error())
+					continue
+				}
+				namespacesCreated[ns] = true
 			}
 		}
 		for objectIndex, obj := range ex.objects {
 			ex.replicaHandler(objectIndex, obj, ns, i, &wg)
 		}
 		if ex.Config.PodWait {
```
vishnuchalla, line 143: Can we condense this block like this, please?
```suggestion
		if ex.Config.PodWait {
                        if !ex.Config.NamespacedIterations || !namespacesWaited[ns] {
                            log.Infof("Waiting up to %s for actions to be completed in namespace", ex.Config.MaxWaitTimeout, ns)
                            ex.waitForObjects(ns)
                            namespacesWaited[ns] = true
                        }
                }
```

pkg/burner/create.go
vishnuchalla: Yes moving them as local variables to `RunCreateJob()` should help the maps flushed out as a part of garbage collection during the end of function execution.

pkg/burner/create.go
```diff
@@ -123,18 +128,29 @@ func (ex *Executor) RunCreateJob(iterationStart, iterationEnd int) {
 		}
 		log.Debugf("Creating object replicas from iteration %d", i)
 		if ex.Config.NamespacedIterations {
```
vishnuchalla, line 130: Can we rewrite this logic as below, please?
```suggestion
		if ex.Config.NamespacedIterations {
                        nsIndex := i / ex.Config.IterationsPerNamespace
                        ns := fmt.Sprintf("%s-%d", ex.Config.Namespace, nsIndex)
                        if namespacesCreated[ns] {
                            continue
                        }
                        if err := createNamespace(ClientSet, ns, nsLabels); err != nil {
                            log.Error(err.Error())
                            continue
                        }
                        namespacesCreated[ns] = true
               }

```

pkg/burner/create.go
afcollins: Not unless you think that's a code smell. if continue implies else condition.

vishnuchalla:
LGTM!

vishnuchalla (approved):

rsevilla87 (approved):
LGTM!

PR #300, opened 2023-04-27
Remove vendor directory

rsevilla87:
### Description

Vendor directory is becoming huge, we better get rid of it and leave the the heavy lifting on golang's dependency resolution.


vishnuchalla (commented):
LGTM!

PR #302, opened 2023-04-28
Deprecate custom template functions

rsevilla87:
The custom template functions are already provided by the sprig library


afcollins (approved):
LGTM!

PR #313, opened 2023-05-16
Remove incorrect job duration message

rsevilla87:
### Description

Removing job duration log message from `create.go` and print rounded duration in seconds rather than seconds with decimals

Fixes: https://github.com/cloud-bulldozer/kube-burner/issues/312


vishnuchalla (approved):
Looks good to me.

Issue #324, opened 2023-05-25
[BUG] Unable to create a NetworkAttachmentDefinition resource from kube burner


josecastillolema:
I am trying to create a **NetworkAttachmentDefinition** resource using Kube-burner.
Below is the job item:
```
  - name: create-networks-job
    jobType: create
    jobIterations: 1
    qps: 10
    burst: 10
    namespacedIterations: true
    cleanup: false
    namespace: serving-ns
    podWait: false
    waitWhenFinished: false
    verifyObjects: true
    errorOnVerify: false
    jobIterationDelay: 1s
    jobPause: 0s
    preLoadImages: false
    objects:
      - objectTemplate: objectTemplates/macvlan_network.yml
        replicas: 1
```

```
apiVersion: k8s.cni.cncf.io/v1
kind: NetworkAttachmentDefinition
metadata:
  name: sriov-net
  namespace: serving-ns-1
spec:
  config: |-
    {
      "cniVersion": "0.3.1",
      "name": "internal-net",
      "plugins": [
        {
          "type": "macvlan",
          "master": "breth0",
          "mode": "bridge",
          "ipam": {
            "type": "static"
          }
        },
        {
          "capabilities": {
            "mac": true,
            "ips": true
          },
          "type": "tuning"
        }
      ]
    }
```

But the creation fails with the error:
```
ERRO[2023-05-23 18:09:28] Error creating object NetworkAttachmentDefinition/sriov-net-1 in namespace serving-ns-1: the server could not find the requested resource 
```

Manually creating the resource works, i.e.: `kubectl apply -f macvlan_network.yml`

Kube-burner version 1.6 (tested 1.3 as well).

rsevilla87:
I managed to reproduce it locally too, I haven't found a fix for this issue though.
It seems like the discovered gvr for NetworkAttachDeffinitions is not correct.

https://github.com/cloud-bulldozer/kube-burner/blob/be7b493d6b6f9301525a329133117f101ffc3eeb/pkg/burner/create.go#LL76C18-L76C43


The root of the issue is that the object Kind is `NetworkAttachmentDefinition` and the previous function returns `networkattachmentdefinitions`. The implementation of `UnsafeGuessKindToResource`  is unsafe and it has some exceptions like this one. It should have returned `network-attachment-definitions` (with dashes), as per api-resources return.

```shell
$ oc api-resources | grep network-attachment
[...]

PR #331, opened 2023-05-29
Shifting from UnsafeGuessKindToResource to RestMapper

rsevilla87:
### Description

Resource discovery is now based on the rest API mappings, rather than  the result from `UnsafeGuessKindToResource`, which turned out to fail in certain cases where the resource naming is different from the object kind, this happens in networkAttachDeffinitions where the kind is `NetworkAttachDeffinition` and the resource name is `network-attach-deffinitions`

Includes some code cleanup:
- `selector.go` is quite useless, the function it contains is only called once, and similar functionality has been implemented with `fmt.Sprintf`
- Use embed config for Executor

cc: @josecastillolema
Fixes: #324


josecastillolema:
Tested successfully against a fairly complex config file including **NetworkAttachmentDefinitons**, pods, deployments, configMaps, serviceAccounts, etc.

vishnuchalla (approved):
lgtm!

PR #349, opened 2023-06-09
Default iterations per namespace for cni heavy

afcollins:
### Fixes ...

... failures running node-density-heavy and node-density-cni @245ppn on clusters with more than 24 worker nodes.

### DIFF
--- a/cmd/kube-burner/ocp-config/node-density-cni/node-density-cni.yml
+++ b/cmd/kube-burner/ocp-config/node-density-cni/node-density-cni.yml
@@ -14,7 +14,8 @@ jobs:
     jobIterations: {{.JOB_ITERATIONS}}
     qps: {{.QPS}}
     burst: {{.BURST}}
-    namespacedIterations: false
+    namespacedIterations: true
+    iterationsPerNamespace: 1000
     podWait: false
     waitWhenFinished: true
     preLoadImages: true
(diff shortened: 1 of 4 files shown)


jtaleric:
Thoughts on renaming the workload to `cni-heavy-v2`? I fear people might just compare the results when they shouldn't really be compared...

afcollins:
My thoughts on that are: this is intended to be a replacement. I don't think the current configuration (245ppn in 1 namespace) should be run at all going forward.

Maybe we have a cutover period in the next readouts that run old & new config, but honestly the current config doesn't provide meaningful results, hence trying to erase it.

jtaleric:
> My thoughts on that are: this is intended to be a replacement. I don't think the current configuration (245ppn in 1 namespace) should be run at all going forward.

ack. ➕  My concern is people (not in our team) taking this new test and comparing it to prior runs. 

> 
> Maybe we have a cutover period in the next readouts that run old & new config, but honestly the current config doesn't provide meaningful results, hence trying to erase it.

afcollins:
I consulted the two known consumers (QE and ACS), and neither are using ocp wrapper to run cni and heavy that they would be affected by this change.

The e2e-benchmarking parameters from kube-burner-ocp-wrapper use the defaults defined in the wrapper. I learned from Paige they have not migrated the jobs with cni or heavy to the e2e ocp-wrapper, so I proposed  to wait until these changes are merged and can work at that time to (hopefully) get more meaningful results out of the test.

I added missing documentation for the config variable, and a note to clarify the behavior.

If we want to add some note to the next release notes or something to call it out, I am happy to do that.

afcollins:
Looks like a flake to me. I can't find a re-run option anywhere.

rsevilla87 (approved):
lgtm!

PR #359, opened 2023-06-15
Add aggregated workers memory utilization and remove memory total/ava…

rsevilla87:
…ilable metrics

### Description

Memory Total and Available can be calculated from the worker hardware specs and memory utilization respectively.


vishnuchalla (approved):
lgtm!

PR #362, opened 2023-06-26
Add iterations-per-namespace as cli option

venkataanil:
This commit adds namespaced-iterations and iterations-per-namespace
cli options to configure namespacedIterations and iterationsPerNamespace
respectively for node-density-cni and node-density-heavy workloads

### DIFF
--- a/cmd/kube-burner/ocp-config/node-density-cni/node-density-cni.yml
+++ b/cmd/kube-burner/ocp-config/node-density-cni/node-density-cni.yml
@@ -14,8 +14,8 @@ jobs:
     jobIterations: {{.JOB_ITERATIONS}}
     qps: {{.QPS}}
     burst: {{.BURST}}
-    namespacedIterations: true
-    iterationsPerNamespace: 1000
+    namespacedIterations: {{.NAMESPACED_ITERATIONS}}
+    iterationsPerNamespace: {{.ITERATIONS_PER_NAMESPACE}}
     podWait: false
     waitWhenFinished: true
     preLoadImages: true
--- a/cmd/kube-burner/ocp-config/node-density-heavy/node-density-heavy.yml
+++ b/cmd/kube-burner/ocp-config/node-density-heavy/node-density-heavy.yml
@@ -18,8 +18,8 @@ jobs:
     jobIterations: {{.JOB_ITERATIONS}}
     qps: {{.QPS}}
     burst: {{.BURST}}
-    namespacedIterations: true
-    iterationsPerNamespace: 1000
+    namespacedIterations: {{.NAMESPACED_ITERATIONS}}
+    iterationsPerNamespace: {{.ITERATIONS_PER_NAMESPACE}}
     podWait: false
     waitWhenFinished: true
     preLoadImages: true
--- a/pkg/workloads/node-density-cni.go
+++ b/pkg/workloads/node-density-cni.go
@@ -26,6 +26,8 @@ import (
 // NewNodeDensity holds node-density-cni workload
 func NewNodeDensityCNI(wh *WorkloadHelper) *cobra.Command {
 	var podsPerNode int
+	var namespacedIterations bool
+	var iterationsPerNamespace int
 	cmd := &cobra.Command{
 		Use:          "node-density-cni",
 		Short:        "Runs node-density-cni workload",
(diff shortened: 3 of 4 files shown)


pkg/burner/create.go
```diff
@@ -187,6 +187,9 @@ func (ex *Executor) RunCreateJob(iterationStart, iterationEnd int) {
 // of iterations before the next namespace is created.
 func (ex *Executor) generateNamespace(iteration int) string {
 	nsIndex := iteration / ex.IterationsPerNamespace
+	if iteration % ex.IterationsPerNamespace == 0 {
+		nsIndex = nsIndex - 1
+	}
```
rsevilla87, line 192: This change is breaking some workloads, since the namespace count starts with 0 rather than 1 as it was before, ie:

```
$ oc get ns                            
NAME                                               STATUS   AGE   
cluster-density-v2-0                               Active   13m   
cluster-density-v2-1                               Active   13m 
``

Comment (rsevilla87) [reply pkg/burner/create.go:192]: This is relevant because the template variable `{{.Iteration}}` must match the namespace index and now is different

File: cmd/kube-burner/ocp-config/node-density/node-density.yml [hunk 2]
```diff
@@ -18,7 +18,8 @@ jobs:
     jobIterations: {{.JOB_ITERATIONS}}
     qps: {{.QPS}}
     burst: {{.BURST}}
-    namespacedIterations: false
+    namespacedIterations: true
```
rsevilla87, line 21: Do we have a team agreement on this change? cc: @dry923 @jtaleric @mukrishn

pkg/burner/create.go
rsevilla87: I realized that the same was happening in node-density-heavy workloads before this patch, but it wasn't affecting the workloads b/c the {{.Iteration}} variable is not used.

cmd/kube-burner/ocp-config/node-density/node-density.yml
mukrishn: I have never had any problem running node-density before but I'm good with this. 
Trying to find the namespace limit(if any) per cluster.

paigerube14, line 21: Can we make this configurable? Is there a case where we would want to run both namespacedIterations=true and namespacedIterations=false versions ?

rsevilla87: This is already possible by extracting the workload and customizing it. https://cloud-bulldozer.github.io/kube-burner/v1.7.1/ocp/#customizing-workloads

rsevilla87: Also, I don't think we should touch the current node-density workload

vishnuchalla (approved):
lgtm except for the existing comments.

pkg/burner/create.go
vishnuchalla: @venkataanil why did we make changes to the namespace index calculation logic? while the previous code was stable: https://github.com/cloud-bulldozer/kube-burner/pull/299

vishnuchalla: Looks like the namespace index issue got fixed in this PR: https://github.com/cloud-bulldozer/kube-burner/pull/364. Tested and verified the change. Thank you!

cmd/kube-burner/ocp-config/node-density/node-density.yml
venkataanil: I want this to be consistent across all node-density variants. With this change, namespacedIterations can be used in all node-density workloads. 

namespacedIterations=false is nothing but using single namespace for all iterations which can be achieved by not setting a value of iterationsPerNamespace (i.e as explained in commit message, if the user doesn't provide a value, the default value will be taken from the job_iterations parameter, maintaining the behavior of creating all pods in a single namespace as before.)

For example, "kube-burner ocp node-density" command will create all pods in a single namespace.

venkataanil: Please check my above comment for namespacedIterations=false and also why are touching node-desnity workload. "kube-burner ocp node-density" command will create all pods in a single namespace i.e current behaviour, with this patch.

pkg/burner/create.go
venkataanil: I ran node-density-cni and observed index starting at 1 while pushing this patch (i.e before PR https://github.com/cloud-bulldozer/kube-burner/pull/364 ) Please see 
https://github.com/cloud-bulldozer/kube-burner/pull/364/commits/1df75b14e13f17b448217c34eccd340091251df7#diff-bc606f2fd54d740859086073970bd68ea15b70016af24d3ac7a446741932b1edL124

 So when the index starts at 1, "iterationsPerNamespace" implementation in #299  creates one less iteration than what user has asked in iterationsPerNamespace. For example, when iterationsPerNamespace is set to 1000, only 999 iterations (starting from 1 to 999) are created in a namespace instead of 1000 iterations. This patch resolves this issue and ensures that the correct number of iterations is created i.e 1000 iterations.

Another scenario is when user doesn't provide iterationsPerNamespace, we want to create all the iterations in a single namespace. Existing implementation is creating them in 2 namespaces as iterations are starting from 1.

I will retest this patch as all workloads start from iteration 0 now because of #364

vishnuchalla: Understood. Thanks for the detailed explanation @venkataanil .

afcollins:
While you are adding CLI arguments, would you please also add one for namespacedIterations? "editing the config" isn't a good response for changing the default because it is more complicated to script. If you are introducing one for iterationsPerNamespace, it makes sense to also add one for the other related config.

pkg/burner/create.go
rsevilla87: I think these lines are not required anymore because of #364 , can you verify it @venkataanil ?

jtaleric (approved):
lgtm, shouldn't break current functionality

afcollins (changes requested):
Commit message still says it updates node-density. Amend the commit to keep it accurate, and LGTM

PR #373, opened 2023-07-10
Keeping timeout constant across stages

vishnuchalla:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

- Related Issue #
- Closes #

### DIFF
--- a/cmd/kube-burner/ocp.go
+++ b/cmd/kube-burner/ocp.go
@@ -46,7 +46,7 @@ func openShiftCmd() *cobra.Command {
 	metricsEndpoint := ocpCmd.PersistentFlags().String("metrics-endpoint", "", "YAML file with a list of metric endpoints")
 	alerting := ocpCmd.PersistentFlags().Bool("alerting", true, "Enable alerting")
 	uuid := ocpCmd.PersistentFlags().String("uuid", uid.NewV4().String(), "Benchmark UUID")
-	timeout := ocpCmd.PersistentFlags().Duration("timeout", 3*time.Hour, "Benchmark timeout")
+	timeout := ocpCmd.PersistentFlags().Duration("timeout", 4*time.Hour, "Benchmark timeout")
 	qps := ocpCmd.PersistentFlags().Int("qps", 20, "QPS")
 	burst := ocpCmd.PersistentFlags().Int("burst", 20, "Burst")
 	gc := ocpCmd.PersistentFlags().Bool("gc", true, "Garbage collect created namespaces")
--- a/pkg/burner/job.go
+++ b/pkg/burner/job.go
@@ -84,7 +84,7 @@ func Run(configSpec config.Spec, prometheusClients []*prometheus.Prometheus, ale
 	go func() {
 		var innerRC int
 		measurements.NewMeasurementFactory(configSpec, indexer, metadata)
-		jobList := newExecutorList(configSpec, uuid)
+		jobList := newExecutorList(configSpec, uuid, timeout)
 		// Iterate job list
 		for jobPosition, job := range jobList {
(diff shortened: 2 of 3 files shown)


rsevilla87:
@vishnuchalla , the problem with your approach of using the --timeout value across different stages will be that, if using defaults, the benchmark and garbage collection timeout will be aggregated rather than combined, that's to say the benchmark could last up to 4h and the garbage collection an additional 4h, making a total timeout of 8h  which could be misleading with the flag description.

Ideally I think we should combine the benchmark and the garbage collection stages under the same timeout value

rsevilla87:
> > @vishnuchalla , thanks for the PR, the problem with this approach of using the --timeout value across different stages will be that, if using defaults, the benchmark and garbage collection timeout will be aggregated rather than combined, that's to say the benchmark could last up to 4h and the garbage collection an additional 4h, making a total timeout of 8h which could be misleading with the flag description.
> > Ideally I think we should combine the benchmark and the garbage collection stages under the same timeout value (meaning that the benchmark+garbage collection shouldn't last more than the configured timeout value), however I found that a bit difficult with the current logic since the job execution is not using any kind of context at the moment, but there must be more alternatives for sure, they just didn't come up to my mind yet :-)
> 
> Yes I agree with your concern. But, what I feel is by setting the timeouts this way we are just creating the extra room for the run, and the run need not necessarily happen till the timeout set. And also when it comes to bechmark+grabage collection, I agree that the benchmark can go upto the timeout set, but the garbage collection need not necessarily execute till its timeout (as we are also triggering parallel deletes right after the benchmark [here](https://github.com/cloud-bulldozer/kube-burner/blob/master/pkg/burner/job.go#L174)). So this change shouldn't have any negative impact is what I believe.
> 
> And also if we want our program to exit right after the benchmark timed out, we can make use of a flag to check for timeout and immediately exit skipping rest of the activity (i.e garbage collection) after the timeout.
> 
> I am open to other people thoughts too! Thank you.

In that case I think we should add a "garbage-collection timeout"  parameter (with a sane default, like 1h or so), so the user can customize it. Thoughts?

vishnuchalla:
> > > @vishnuchalla , thanks for the PR, the problem with this approach of using the --timeout value across different stages will be that, if using defaults, the benchmark and garbage collection timeout will be aggregated rather than combined, that's to say the benchmark could last up to 4h and the garbage collection an additional 4h, making a total timeout of 8h which could be misleading with the flag description.
> > > Ideally I think we should combine the benchmark and the garbage collection stages under the same timeout value (meaning that the benchmark+garbage collection shouldn't last more than the configured timeout value), however I found that a bit difficult with the current logic since the job execution is not using any kind of context at the moment, but there must be more alternatives for sure, they just didn't come up to my mind yet :-)
> > 
> > 
> > Yes I agree with your concern. But, what I feel is by setting the timeouts this way we are just creating the extra room for the run, and the run need not necessarily happen till the timeout set. And also when it comes to bechmark+grabage collection, I agree that the benchmark can go upto the timeout set, but the garbage collection need not necessarily execute till its timeout (as we are also triggering parallel deletes right after the benchmark [here](https://github.com/cloud-bulldozer/kube-burner/blob/master/pkg/burner/job.go#L174)). So this change shouldn't have any negative impact is what I believe.
> > And also if we want our program to exit right after the benchmark timed out, we can make use of a flag to check for timeout and immediately exit skipping rest of the activity (i.e garbage collection) after the timeout.
> > I am open to other people thoughts too! Thank you.
> 
> In that case I think we should add a "garbage-collection timeout" parameter (with a sane default, like 1h or so), so the user can customize it. Thoughts?

Added GCTimeout param as suggested.

PR #375, opened 2023-07-10
Adding delete and wait functinality for non-namespaced resources

vishnuchalla:
## Type of change

- [x] Refactor
- [x] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

- Related Issue #
- Closes # https://github.com/cloud-bulldozer/kube-burner/issues/333

## Testing
Tested building a binary on top of the source code on self-managed AWS cluster. Steps to test non-namespaced resources is as below.
* Deploy this CRD on to your cluster
```
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: myclusterresources.example.com
spec:
  group: example.com
  names:
    plural: myclusterresources
    singular: myclusterresource
    kind: MyClusterResource
  scope: Cluster
  versions:
    - name: v1
      served: true
      storage: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              # Add your desired properties here
```
* Create a resource with the CRD as below
```
apiVersion: example.com/v1
kind: MyClusterResource
metadata:
  name: my-cluster-resource-5
  labels:
    kube-burner-uuid: tes
```
* Extract any of the workloads and add this resource as one of the objects (i.e namespaced:false). For example `node-density-heavy`
```
    objects:

      - objectTemplate: mycustomresource.yml
        replicas: 1
        namespaced: false

      - objectTemplate: postgres-deployment.yml
        replicas: 1

      - objectTemplate: app-deployment.yml
        replicas: 1
        inputVars:
          probesPeriod: {{.PROBES_PERIOD}}

      - objectTemplate: postgres-service.yml
        replicas: 1

```
* Now run the workload using command similar to below
```
kube-burner ocp node-density-heavy --timeout 1h --metrics-endpoint ~/metrics-endpoints.yaml --pods-per-node=500 --log-level=debug 2>&1 | tee -a kube-burner-$(date +"%F_%H-%M-%S")
```
The same works with the `destroy`  sub command as well.
Full length testing logs here: http://pastebin.test.redhat.com/1104569


vishnuchalla:
@darkmuggle Requesting review. Thank you.

pkg/config/types.go
```diff
@@ -84,7 +84,7 @@ type Object struct {
 	// LabelSelector objects with this labels will be removed
 	LabelSelector map[string]string `yaml:"labelSelector" json:"labelSelector,omitempty"`
 	// Namespaced this object is namespaced
-	Namespaced bool `yaml:"-" json:"-"`
+	Namespaced bool `yaml:"namespaced" json:"namespaced"`
```
rsevilla87, line 87: What's the purpose if this change?

rsevilla87: The Namespaced field was removed from the YAML on purpose as it's discovered automatically.

vishnuchalla: Was trying to understand its significance while testing and forgot to revert back. Doing it now. Thank you!

pkg/config/config.go
```diff
@@ -69,6 +69,7 @@ func (j *Job) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	type rawJob Job
 	raw := rawJob{
 		Cleanup:                true,
+		Namespace:              "default",
```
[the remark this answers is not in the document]
vishnuchalla: Looks like all the current runs mandate a namespace, or we will run into a validation error while parsing the config. I did this in order to avoid that. As per our discuss offline as this change is dangerous, reverting it back as it is expected to fail when there is not namespace specified. Thank you!

pkg/burner/namespaces.go
```diff
@@ -74,6 +77,45 @@ func CleanupNamespaces(ctx context.Context, l metav1.ListOptions, cleanupWait bo
 	}
 }
 
+// Cleanup non-namespaced resources with the given selector
+func CleanupNonNamespacedResources(ctx context.Context, l metav1.ListOptions, cleanupWait bool) {
```
rsevilla87, line 81: I think this function fits on the `destroy` approach since when a user runs `kube-burner destroy` does not pass any kind of configuration file, so we have to iterate over all available resources.

However, when the workload gets garbage collected, we know what resources it has created. Would it be possible to create an alternative function able to only iterate over (and try to delete) the non-namespaces resources created by the workload?

smalleni:
Teted this with a custom create-crd job I was playing with and works as expected

```
time="2023-07-26 11:56:14" level=info msg="Job create-crd took 29s" file="job.go:160"
time="2023-07-26 11:56:14" level=info msg="Finished execution with UUID: c424c0eb-7c95-4143-a622-fc0c33256408" file="job.go:194"
time="2023-07-26 11:56:14" level=info msg="Garbage collecting remaining namespaces" file="job.go:207"
time="2023-07-26 11:56:14" level=info msg="Deleting namespaces with label kube-burner-uuid=c424c0eb-7c95-4143-a622-fc0c33256408" file="namespaces.go:63"
time="2023-07-26 11:56:14" level=info msg="Deleting namespaces with label kube-burner-uuid=c424c0eb-7c95-4143-a622-fc0c33256408" file="namespaces.go:63"
time="2023-07-26 11:56:15" level=info msg="Waiting for namespaces to be definitely deleted" file="namespaces.go:120"
I0726 11:56:22.420508    4583 request.go:696] Waited for 1.018735708s due to client-side throttling, not priority and fairness, request: GET:https://api.saiocp.perfscale.devcluster.openshift.com:6443/apis/node.k8s.io/v1?timeout=15s
time="2023-07-26 11:56:28" level=info msg="Deleting non-namespace resources with label kube-burner-uuid=c424c0eb-7c95-4143-a622-fc0c33256408" file="namespaces.go:83"
```

pkg/burner/namespaces.go
vishnuchalla: Yes Raul, Its already being achieved with this same function. When a workload is ran, all the resources only specific to that workload will have `kube-burner-uuid=` label assigned. And in the `GC` step we are only cleaning up the resources specific to that workload by calling this function with UUID label.
```
CleanupNonNamespacedResources(ctx, v1.ListOptions{LabelSelector: fmt.Sprintf("kube-burner-uuid=%v", uuid)}, true)
```

rsevilla87: Right, I knew that already. What I meant was that in the way  the function`CleanupNonNamespacedResources` is coded, it tries to delete all the available resources in the API labeled with `kube-burner-uuid=<uuid>`.


My suggestion was to create an alternative function that tries to delete only resources a benchmark has created rather than iterating over all the available resources exposed in the API.

vishnuchalla: Ohh, Sorry my bad. I initially misunderstood this comment. Added a function to delete non namespaced resources using GVR. Thank you!

PR #377, opened 2023-07-11
Using only one deamonset to fetch all the preload images

vishnuchalla:
## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description
Currently kube-burner does deploy a demonset for each image in the preload step which will create `pods=number_of_worker_nodes` for each image. This approach uses only one single deamonset for the entire list of images as we only care about pulling them into our cluster rather then running them in the `preload` step.

- Related Issue #
- Closes #

## Testing
Tested on my self managed cluster by verifying the events. Attaching the logs below.
```
^C[vchalla@vchalla ~]$ oc get events
LAST SEEN   TYPE      REASON             OBJECT                   MESSAGE
14s         Normal    Scheduled          pod/preload2jpmt-hwcqc   Successfully assigned preload-kube-burner/preload2jpmt-hwcqc to ip-10-0-192-233.us-west-2.compute.internal
14s         Normal    AddedInterface     pod/preload2jpmt-hwcqc   Add eth0 [10.128.2.132/23] from ovn-kubernetes
10s         Normal    Pulling            pod/preload2jpmt-hwcqc   Pulling image "registry.redhat.io/rhel8/postgresql-10@sha256:4b912c80085b88a03309aeb7907efcc29dd3342fa3952b6ea067afb1914bfe53"
13s         Normal    Pulled             pod/preload2jpmt-hwcqc   Successfully pulled image "registry.redhat.io/rhel8/postgresql-10@sha256:4b912c80085b88a03309aeb7907efcc29dd3342fa3952b6ea067afb1914bfe53" in 1.254680227s (1.254692137s including waiting)
9s          Normal    Created            pod/preload2jpmt-hwcqc   Created container container-0
9s          Normal    Started            pod/preload2jpmt-hwcqc   Started container container-0
12s         Normal    Pulling            pod/preload2jpmt-hwcqc   Pulling image "quay.io/cloud-bulldozer/perfapp:latest"
11s         Normal    Pulled             pod/preload2jpmt-hwcqc   Successfully pulled image "quay.io/cloud-bulldozer/perfapp:latest" in 1.193207898s (1.193223358s including waiting)
11s         Normal    Created            pod/preload2jpmt-hwcqc   Created container container-1
11s         Normal    Started            pod/preload2jpmt-hwcqc   Started container container-1
```

### DIFF
--- a/pkg/burner/pre_load.go
+++ b/pkg/burner/pre_load.go
@@ -100,48 +100,54 @@ func createDSs(imageList []string, namespaceLabels map[string]string) error {
 	if err := createNamespace(preLoadNs, nsLabels); err != nil {
 		log.Fatal(err)
 	}
-	for i, image := range imageList {
-		dsName := fmt.Sprintf("preload-%d", i)
-		container := corev1.Container{
-			Name:            "kube-burner-rocks",
-			ImagePullPolicy: corev1.PullAlways,
-			Image:           image,
-		}
-		ds := appsv1.DaemonSet{
-			TypeMeta: metav1.TypeMeta{
-				Kind:       "DaemonSet",
-				APIVersion: "apps/v1",
+	dsName := "preload"
+	ds := appsv1.DaemonSet{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "DaemonSet",
+			APIVersion: "apps/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: dsName,
+		},
+		Spec: appsv1.DaemonSetSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{"app": dsName},
 			},
-			ObjectMeta: metav1.ObjectMeta{
-				GenerateName: dsName,
-			},
-			Spec: appsv1.DaemonSetSpec{
-				Selector: &metav1.LabelSelector{
-					MatchLabels: map[string]string{"app": dsName},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{"app": dsName},
 				},
-				Template: corev1.PodTemplateSpec{
-					ObjectMeta: metav1.ObjectMeta{
-						Labels: map[string]string{"app": dsName},
-					},
-					Spec: corev1.PodSpec{
-						TerminationGracePeriodSeconds: pointer.Int64(0),
-						InitContainers:                []corev1.Container{container},
-						// Only Always restart policy is supported
-						Containers: []corev1.Container{
-							{
-								Name:  "sleep",
-								Image: "gcr.io/google_containers/pause-amd64:3.0",
-							},
+				Spec: corev1.PodSpec{
+					TerminationGracePeriodSeconds: pointer.Int64(0),
+					InitContainers:                []corev1.Container{},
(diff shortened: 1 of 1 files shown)


pkg/burner/pre_load.go
```diff
@@ -55,9 +55,9 @@ func preLoadImages(job Executor) error {
 	time.Sleep(job.PreLoadPeriod)
 	log.Infof("Pre-load: Deleting namespace %s", preLoadNs)
 	// 5 minutes should be more than enough to cleanup this namespace
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
-	defer cancel()
-	CleanupNamespaces(ctx, metav1.ListOptions{LabelSelector: "kube-burner-preload=true"}, true)
+	//ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+	//defer cancel()
+	//CleanupNamespaces(ctx, metav1.ListOptions{LabelSelector: "kube-burner-preload=true"}, true)
```
rsevilla87, line 60: ?

pkg/burner/pre_load.go
```diff
@@ -100,48 +100,45 @@ func createDSs(imageList []string, namespaceLabels map[string]string) error {
...
+					TerminationGracePeriodSeconds: pointer.Int64(0),
+					Containers:                    []corev1.Container{},
 				},
 			},
+		},
+	}
+
+	// Add the list of containers using images
+	for i, image := range imageList {
+		container := corev1.Container{
```
rsevilla87, line 130: The containers with the workload images, should be appended to initContainers, if you use a regular container, it keeps restarting when it finishes.

pkg/burner/pre_load.go
vishnuchalla: Sorry about that. Forgot to uncomment them after testing. Corrected it now.

pkg/burner/pre_load.go
vishnuchalla: Acknowledged. When the containers are pulled and started successfully we do not see any restarts in case of adding those list of images to either of `containers` or `initcontainers`. But when there are is error in starting either of `containers` or `initcontainers` the restarts happens for both. Example for `initcontainers` restart

```
preload-0jkbzb-lvplw   0/1     Init:Error              1 (3s ago)   6s
preload-0jkbzb-5jpd6   0/1     Init:CrashLoopBackOff   1 (1s ago)   6s
preload-0jkbzb-lvplw   0/1     Init:CrashLoopBackOff   1 (2s ago)   7s
preload-0jkbzb-5jpd6   0/1     Init:Error              2 (14s ago)   19s
preload-0jkbzb-j577r   0/1     Init:Error              2 (16s ago)   20s

```
So this change shouldn't make any difference. The only advantage i see is when we have large number of worker nodes for example 250 and 3 images in our job, we can get all the images pulled using 250 pods each in every node instead of having 750 pods which the previous logic used to deploy. It will cut down the `total_number_of_pods` to `total_number_of_pods/no_of_images`. But open to other suggestions too. Thank you.

vishnuchalla: Thanks for the detailed explanation. Changed as suggested. Keeping only `sleep` as the main containers and moving the rest to init containers.

Issue #372, opened 2023-07-07
[RFE] Configuration file parsing


rsevilla87:
**Is your feature request related to a problem? Please describe.**

At the moment the current configuration parsing uses the standard YAML decoder plus some tweaks to establish sane defaults, it's not the best of the solutions so I think we could start working on other approaches.

**Describe the solution you'd like**
The [viper](https://github.com/spf13/viper) library provides some interesting features such as reading from environment variables, settings defaults and more. I proposed this library as an alternative b/c is the one I'm more familiar with, but there's not any hard requirement to use it.

smalleni:
@vishnuchalla This seems like a good task for Shashank, thoughts?

vishnuchalla:
Yes. Should to be decent one to explore viper.

shashank-boyapally:
By exploring viper, we see that we can use setDefault method to set defaults for the unset parameters, but looking deeper into it we understand we couldn't set defaults for the slice of objects using viper. In order to tackle this issue I'm checking if the param is set or not, and if not I'm assigning a default value which we need. This process could only be done after Unmarshaling to the configSpec. The following are the drawbacks which I felt:

- Since we are cannot use setDefault method, viper is only being used to read the config(this too partially) and unmarshal it to a struct.
- The major idea to shift from yamlv3 to viper is to set defaults which is not possible at the moment.
- Using viper for Unmarshaling and reading and performing the loop operation to check of set of param? is costly which inturn decreases our efficiency. 
- We cannot override the Unmarshal method of viper.

Since yamlv3 now preserves the ordering of the objects, and Unmarshal method  can be overridden where we could set defaults, I feel the present way is the most optimal approach for the time being. Please let me know any of your thoughts over this.

PR #380, opened 2023-07-13
Using viper to set Defaults for configuration file

shashank-boyapally:
## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
We intend to use viper to set defaults for the parameters which are required but are not set in the yml file. Since through viper we could not access the slice of objects in yml, the approach which was taken was to set default after unmarshaling to configSpec, by checking if the parameter was not set in the yml file. A simple example is shown below

```
if !cfg.IsSet(fmt.Sprintf("jobs.%d.cleanup", i)) {
			configSpec.Jobs[i].Cleanup = true
		}
```

- Related Issue #372 
- Closes #


chentex (approved):
LGTM

rsevilla87:
Thanks for your inputs and investigation @shashank-boyapally, I spoke yesterday with @vishnuchalla about this topic, and we concluded that given that we cannot leverage viper to set the defaults on the benchmark's jobs, which is the main intention of this effort, might be better to leave this PR behind. And focus in other issues/RFEs

What do you think?

shashank-boyapally:
Yes @rsevilla87, thank you for the response. I think it's better to close this PR and move ahead with other issues. I'm closing this PR.

Issue #288, opened 2023-04-05
[RFE] preload uses node selectors specified


afcollins:
**Is your feature request related to a problem? Please describe.**
If I specify node selectors for kube-burner to select certain nodes, I also want the capability for preload to use the same node selector. Currently, preload runs on all nodes of the cluster, which raises eyebrows with users looking to restrict the execution.

**Describe the solution you'd like**
I would like preload daemonset to be created with the node selector specified to kube-burner.

Or, as a stretch, to specify a node selector for kube-burner at the job or global level to apply to all workloads created within that group.

**Describe alternatives you've considered**
Exposing any node selectors to be templated into the workload manifests are at the config.jobs.objects level, but these have no influence on preload.

**Additional context**
The ability to run workloads comes from e2e-benchmarking, where the namespace label is exposed as a variable. I see that it is not exposed in the `kube-burner ocp` subcommand workloads, which assumes we are running on `node-role.kubernetes.io/worker` nodes.

afcollins:
Capturing more detail here after chatting with Vishnu more about this yesterday.

Preload daemonset has node selectors hard-coded in go code, and no means of configuration override.

The ocp-wrapper also does not expose variables for node selector labels, however, one could achieve this by extracting the workload and modifying the configs with specific labels.

The larger ask here is to have the ability to override nodeSelectors for both preload and all workload pods.

The use case is a user who wants to be able to run on a subset of nodes in a cluster, without scheduling to any of the other nodes in the cluster. With the current setup, if I modify the workloads with specific labels, preload will still run on all role=worker nodes, which may not include the subset of nodes.

vishnuchalla:
Added an option `preaLoadNodeLabels` at a job level to specify node selectors for preload resources. Keeping it at the lowest granular level possible to leave it with flexibility on specifying node selector options for mixed requirements. Feel free to let us know if you have any further questions. Thank you!

vishnuchalla:
[...]

PR #382, opened 2023-07-16
Adding node selector labels for preload step

vishnuchalla:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description
Added a parameter to specify node selectors in the preload step. This provides us with the flexibility to fully run kube-burner in desired set of nodes.

- Related Issue #
- Closes # https://github.com/cloud-bulldozer/kube-burner/issues/288

## Testing
Tested building a binary on my self managed cluster.
### Step: 1
Added labels `app=test` to one of worker nodes.
```
ip-10-0-183-53.us-west-2.compute.internal    Ready    worker                 82d   v1.25.8+27e744f   app=test,beta.kubernetes.io/arch=amd64,beta.kubernetes.io/instance-type=m5.4xlarge,beta.kubernetes.io/os=linux,failure-domain.beta.kubernetes.io/region=us-west-2,failure-domain.beta.kubernetes.io/zone=us-west-2b,kubernetes.io/arch=amd64,kubernetes.io/hostname=ip-10-0-183-53.us-west-2.compute.internal,kubernetes.io/os=linux,node-role.kubernetes.io/worker=,node.kubernetes.io/instance-type=m5.4xlarge,node.openshift.io/os_id=rhcos,topology.ebs.csi.aws.com/zone=us-west-2b,topology.kubernetes.io/region=us-west-2,topology.kubernetes.io/zone=us-west-2b 
```
### Step: 2
Modified the node-density-heavy config `jobs` section to use the label.
```
jobs:
  - name: node-density-heavy
    namespace: node-density-heavy
    jobIterations: {{.JOB_ITERATIONS}}
    qps: {{.QPS}}
    burst: {{.BURST}}
    namespacedIterations: true
    iterationsPerNamespace: 1000
    podWait: false
    waitWhenFinished: true
    preLoadImages: true
    preLoadPeriod: 15s
    preLoadNodeLabels:
      app: test
    namespaceLabels:
      security.openshift.io/scc.podSecurityLabelSync: false
      pod-security.kubernetes.io/enforce: privileged
      pod-security.kubernetes.io/audit: privileged
      pod-security.kubernetes.io/warn: privileged
    objects:

      - objectTemplate: postgres-deployment.yml
        replicas: 1
```
### Step: 3
Ran and verified if the pre-load pods are getting deployed only in the nodes that has specified labels.
```
NAME                   READY   STATUS                  RESTARTS     AGE   IP            NODE                                        NOMINATED NODE   READINESS GATES
preload-0wsp7m-64bdb   0/1     Init:CrashLoopBackOff   1 (7s ago)   12s   10.129.3.71   ip-10-0-183-53.us-west-2.compute.internal   <none>           <none>
preload-1czp49-f7m6z   0/1     Init:0/1                0            11s   10.129.3.72   ip-10-0-183-53.us-west-2.compute.internal   <none>           <none>
```

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -60,11 +60,14 @@ This section contains the list of jobs `kube-burner` will execute. Each job can 
 | `errorOnVerify`        | Set RC to 1 when objects verification fails                                      | Boolean | true    |
 | `preLoadImages`        | Kube-burner will create a DS before triggering the job to pull all the images of the job   | true    |
 | `preLoadPeriod`        | How long to wait for the preload daemonset                                       | Duration| 1m     |
+| `preloadNodeLabels`    | Add node selector labels for the resources created in preload stage              | Object  | {} |
 | `namespaceLabels`      | Add custom labels to the namespaces created by kube-burner                       | Object  | {} |
 | `churn`                | Churn the workload. Only supports namespace based workloads                      | Boolean | false |
 | `churnPercent`         | Percentage of the jobIterations to churn each period                             | Integer | 10 |
 | `churnDuration`        | Length of time that the job is churned for                                       | Duration| 1h |
 | `churnDelay`           | Length of time to wait between each churn period                                 | Duration| 5m |
+
+Our configuration files strictly follow YAML syntax. To clarify on List and Object types usage, they are nothing but the `Lists and Dictionaries` in YAML syntax like mentioned [here](https://gettaurus.org/docs/YAMLTutorial/#Lists-and-Dictionaries). Please feel free to refer YAML syntax more for details on a specific `Type` usage. 
 
 Examples of valid configuration files can be found at the [examples folder](https://github.com/cloud-bulldozer/kube-burner/tree/master/examples).
 
--- a/cmd/kube-burner/ocp-config/cluster-density-v2/build.yml
+++ b/cmd/kube-burner/ocp-config/cluster-density-v2/build.yml
@@ -4,6 +4,10 @@ apiVersion: build.openshift.io/v1
 metadata:
   name: {{.JobName}}-{{.Replica}}
 spec:
+  resources:
+    requests:
+      cpu: 10m
+      memory: "10Mi"
   nodeSelector:
     node-role.kubernetes.io/worker: ""
   serviceAccount: builder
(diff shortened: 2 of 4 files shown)


rsevilla87 (commented):
Not against this change, but honestly, I don't see much value in it, the preload stage of a benchmark is a simply an stage to trigger the image pulling from all available nodes across the cluster regardless of their role.

This pre-load stage doesn't affect to the workload results itself as the metrics or other measurements are not collected before pre-load finishes.


I'd like to hear others opinions too, what do you think @afcollins?

afcollins:
@rsevilla87 the intention here is if a user creates a workload to run on nodes with certain labels only (that don't have the `node-role.../worker` label), then the preload ds doesn't schedule to those nodes.

I am also not crazy about this implementation.
I would rather have a more generic capability to set/override nodeSelector labels in all things created by kube-burner (including the preload daemonset and all workload pods), which we had in e2e-benchmarking and no longer have in ocp-wrapper.

It's a bigger change than this tactical fix, but it has its applications if we want to increase adoption of the tool.

We can take the discussion to the #288.

docs/reference/configuration.md
```diff
@@ -59,6 +59,7 @@ This section contains the list of jobs `kube-burner` will execute. Each job can
 | `errorOnVerify`        | Set RC to 1 when objects verification fails                                      | Boolean | true    |
 | `preLoadImages`        | Kube-burner will create a DS before triggering the job to pull all the images of the job   | true    |
 | `preLoadPeriod`        | How long to wait for the preload daemonset                                       | Duration| 1m     |
+| `preloadNodeLabels`    | Add node selector labels for the resources created in preload stage              | Object  | {} |
```
afcollins, line 62: I think an example of what format to use to specify the labels would be helpful. I get confused when trying to specify inlined arrays, so I would like to know which braces and quote syntax I need to use.

rsevilla87: Right, also, what does leaving this field empty mean?

rsevilla87: ok, can you document it along with and example?

vishnuchalla: Updated in the docs as suggested!

rsevilla87 (approved):
lgtm!

PR #401, opened 2023-08-03
Cluster-Density-v2: Create services and router before client pods

smalleni:
This is required because the readiness probe of the client depends on being able to hit the service/route. It only makes sense to create the service and route before the client pod.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [X] Optimization
- [ ] Documentation Update

### DIFF
--- a/cmd/kube-burner/ocp-config/cluster-density-v2/cluster-density-v2.yml
+++ b/cmd/kube-burner/ocp-config/cluster-density-v2/cluster-density-v2.yml
@@ -30,16 +30,22 @@ jobs:
       pod-security.kubernetes.io/warn: privileged
     objects:
 
-      - objectTemplate: imagestream.yml
-        replicas: 1
+      - objectTemplate: service.yml
+        replicas: 5
 
-      - objectTemplate: build.yml
-        replicas: 1
+      - objectTemplate: route.yml
+        replicas: 2
 
       - objectTemplate: deployment-server.yml
         replicas: 3
         inputVars:
           podReplicas: 2
+
+      - objectTemplate: imagestream.yml
+        replicas: 1
+
+      - objectTemplate: build.yml
+        replicas: 1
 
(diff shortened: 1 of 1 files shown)


cmd/kube-burner/ocp-config/cluster-density-v2/cluster-density-v2.yml
```diff
@@ -41,6 +41,12 @@ jobs:
         inputVars:
           podReplicas: 2
 
+      - objectTemplate: service.yml
```
afcollins, line 44: If we move these, why not move all the way to the top? The build and server take some time that we could overlap with service & route creation.

smalleni: You are right.

rsevilla87 (commented):
Adding some thoughts:

- Objects within an iteration are not created in order. Every object replica gets created by its own goroutine, which doesn't guarantee any execution order https://github.com/cloud-bulldozer/kube-burner/blob/ced0d8a90d0641ef6aad84b6f78f123cb66558cc/pkg/burner/create.go#L195
- Each cluster-density-v2 iteration has less than 37 object replicas, so, in case of using a QPS/Burst rate of 20, the biggest time difference between one route/service and a pod creation would be less than 2 seconds. 

Rather than sorting the object list I'd configure a _initialDelaySeconds_ of at least two seconds in the pods in order to guarantee that the services/routes are available (which doesn't actually mean that is ready to use) when then the readinessProbe is triggered.

rsevilla87:
what do you think from the last comment @smalleni ?

rsevilla87:
This PR is not relevant due to the reasons I explained at https://github.com/cloud-bulldozer/kube-burner/pull/401#pullrequestreview-1630497833, closing

PR #412, opened 2023-08-08
Increase cluster-density-v2 build to 20m CPU request...

afcollins:
... because 10m isn't enough that nodes can still
 become overwhelmed by builds, which actually
 use around 15m.

## Type of change

- [x] Optimization


rsevilla87:
> ... because 10m isn't enough that nodes can still become overwhelmed by builds, which actually use around 15m.
> ## Type of change
> 
>     * [x]  Optimization
> 
> 
> ## Checklist before requesting a review
> 
>     * [x]  I have performed a self-review of my code.

have you tested it?

afcollins:
Yes, I tested it. Works as expected. Fewer builds are scheduled to the same nodes.

rsevilla87 (approved):
lgtm!

PR #432, opened 2023-08-14
Cleanup and code optimizations

rsevilla87:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

- Remove the `measurementStopFunctions` variable.
- Call `CleanupNonNamespacedResourcesUsingGVR` during cleanup stage rather than `CleanupNonNamespacedResources`
- Add locks to pod latency calculations, multi-job benchmarks are failing due a potential concurrent map access by the different goroutines
- Missing fields to VMI & VM latency calculation

- Related Issue #
- Closes #

### DIFF
--- a/pkg/burner/job.go
+++ b/pkg/burner/job.go
@@ -240,13 +240,14 @@ func newExecutorList(configSpec config.Spec, uuid string, timeout time.Duration)
 	}
 	discoveryClient = discovery.NewDiscoveryClientForConfigOrDie(restConfig)
 	for _, job := range configSpec.Jobs {
-		if job.JobType == config.CreationJob {
+		switch job.JobType {
+		case config.CreationJob:
 			ex = setupCreateJob(job)
-		} else if job.JobType == config.DeletionJob {
+		case config.DeletionJob:
 			ex = setupDeleteJob(&job)
-		} else if job.JobType == config.PatchJob {
+		case config.PatchJob:
 			ex = setupPatchJob(job)
-		} else {
+		default:
 			log.Fatalf("Unknown jobType: %s", job.JobType)
 		}
 		for _, j := range executorList {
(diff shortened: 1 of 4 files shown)


pkg/burner/job.go
```diff
@@ -174,22 +174,13 @@ func Run(configSpec config.Spec, prometheusClients []*prometheus.Prometheus, ale
 					log.Errorf("Failed measurements: %v", err.Error())
 					innerRC = 1
 				}
-			} else {
-				measurementStopFunctions = append(measurementStopFunctions, measurements.Stop)
 			}
 		}
 		if globalConfig.WaitWhenFinished {
 			runWaitList(globalWaitMap, executorMap)
-			endTime := time.Now().UTC()
```
vishnuchalla, line 183: I think we should keep those `measurementStopFunctions` as is. The idea behind them was, when have `WaitWhenFinished` enabled at global level, our workload creates all the resources at once and finally waits at the end of benchmark instead of waiting per each job. So at the end of the benchmark is when we finally wait for our resources and make sure they are in ready state and then calculate measurements. But if we have `stop` method called per each job, by that time resources might have been created and we might  end up in a situation where we call `measurement.stop()` even before making sure that our resources are reached to `ready` state which is only happened at the end once all the resources in given list of jobs are created (i.e only when `global.WaitWhenFinished` is enabled).

rsevilla87: I knew that, don't worry. Note that with suggested implementation, when `WaitWhenFinished` set enabled at global scope,  we only stop measurements after running all benchmark's jobs

vishnuchalla: Acknowledged. Looks good to me.

PR #435, opened 2023-08-15
Increase node-density podReadyThreshold to 15s.

### LINKED ISSUE #424: [RFE] Remove default pod-ready-latency=5s for ocp node-density

--pod-ready-latency is a feature, but not all environments will actually meet the p99 < 5s latency that is the default `kube-burner ocp node-density`.

I see failures in runs all over (i.e. QE CI, airflow playground, airflow CPT) that should proceed forward instead of halting the entire test flow because p99 was higher than this arbitrary value.

Also, latest 4.14 nightlies p99s are well above 5s, so I think we need to reconsider how we use this value.

Keep the flag. Remove the default value. It's a step towards a "more friendly to newcomers" user experience.

I could set the latency explicitly if I understand what a reasonable value is, but a user may not always know.

afcollins:
5s is too low for what we are seeing on 4.14.

Partially resolves #424, but does not unset by default.

## Type of change

- [x] Bug fix

## Description

node-density fails on unknowing users because OCP 4.14 introduces changes that regularly bring p99 Ready above 5s.  This adjusts our tool to respond to that behavior change.

- Related Issue #424


afcollins:
@vishnuchalla @rsevilla87 PTAL

pkg/workloads/node-density.go
```diff
@@ -49,7 +49,7 @@ func NewNodeDensity(wh *WorkloadHelper) *cobra.Command {
 		},
 	}
 	cmd.Flags().IntVar(&podsPerNode, "pods-per-node", 245, "Pods per node")
-	cmd.Flags().DurationVar(&podReadyThreshold, "pod-ready-threshold", 5*time.Second, "Pod ready timeout threshold")
+	cmd.Flags().DurationVar(&podReadyThreshold, "pod-ready-threshold", 15*time.Second, "Pod ready timeout threshold")
```
vishnuchalla, line 52: I am just curious, on how did we decide 15 seconds as threshold? I see for [node-density-heavy](https://github.com/cloud-bulldozer/kube-burner/blob/e5c8c1e224f76653158b00caad6af0c5edb74533/pkg/workloads/node-density-heavy.go#L55) it is 1 hour.

afcollins: Well, I would rather remove the default setting entirely. :)

In recent runs I see max 12s for p99 Ready, so 15s gives some grace period.

My serious preference is to remove it entirely, I just didn't think y'all would buy in to that.

rsevilla87: I configured 1 hour randomly for node-density-heavy, so don't take it into account :)

afcollins: Yes, OK. But this is still the default if no user sets it. If you don't want to keep changing it here, then we should just remove the default entirely.

vishnuchalla: Removing this completely would leave it unset and we will never see an alert on any value. So my suggestions is to have a default as suggested in this change and have an extra `ENV` added in `e2e` so that we need not visit here again and again.

afcollins: Sounds good. Please merge.

PR #437, opened 2023-08-16
Added replicaset tracking to metrics

nathan-weinberg:
See https://github.com/cloud-bulldozer/e2e-benchmarking/pull/615 for context

### DIFF
--- a/cmd/kube-burner/ocp-config/metrics-aggregated.yml
+++ b/cmd/kube-burner/ocp-config/metrics-aggregated.yml
@@ -124,6 +124,9 @@
 - query: sum(kube_node_status_condition{status="true"}) by (condition)
   metricName: nodeStatus
 
+- query: count(kube_replicaset_labels{})
+  metricName: replicaSets
+
 # Prometheus metrics
 
 - query: openshift:prometheus_tsdb_head_series:sum{job="prometheus-k8s"}
--- a/cmd/kube-burner/ocp-config/metrics-aggregated.yml
+++ b/cmd/kube-burner/ocp-config/metrics-aggregated.yml
@@ -125,7 +125,8 @@
   metricName: nodeStatus
 
 - query: count(kube_replicaset_labels{})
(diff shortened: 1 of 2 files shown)


rsevilla87:
ReplicaSets get created on a 1:1 ratio with deployments, are you sure we need this?

nathan-weinberg:
@memodi See above, does this satisfy the requirement you've mentioned?

cmd/kube-burner/ocp-config/metrics-aggregated.yml
```diff
@@ -124,6 +124,9 @@
 - query: sum(kube_node_status_condition{status="true"}) by (condition)
   metricName: nodeStatus
 
+- query: count(kube_replicaset_labels{})
+  metricName: replicaSets
```
rsevilla87, line 128: Can we make this query instant?, I think we don't need it timebased

nathan-weinberg (commented):
dsadas

cmd/kube-burner/ocp-config/metrics-aggregated.yml
rsevilla87, line 128: I think is better to make this query instant and update its name to be consistent with other object counts.

```suggestion
- query: count(kube_replicaset_labels{})
  metricName: replicaSetCount
  instant: true
```

cmd/kube-burner/ocp-config/metrics.yml
```diff
@@ -115,6 +115,9 @@
 - query: sum(kube_node_status_condition{status="true"}) by (condition)
   metricName: nodeStatus
 
+- query: count(kube_replicaset_labels{})
```
rsevilla87, line 118: Same as above

nathan-weinberg:
Thanks for the review @rsevilla87! Made the requested changes :+1:

vishnuchalla (approved):
lgtm!

PR #446, opened 2023-08-23
Adding audit functionality on latencies

vishnuchalla:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

* Finalizing V2 logic for latencies and removing V1.
* Keeping V2 as it is accurate, consistent and supportive for future development practices.
* Adding audit functionality based on V2 as it proved to be reliable. Please take a look at the attached issue for more context. 

- Related Issue # https://github.com/cloud-bulldozer/kube-burner/issues/439
- Closes #

## Testing
Tested on self managed cluster with below command
```
kube-burner ocp node-density-heavy --timeout 1h --metrics-endpoint ~/metrics-endpoints.yaml --pods-per-node=50 --log-level=trace 2>&1 --gc=true | tee -a kube-burner-$(date +"%F_%H-%M-%S")
```
Full logs: https://gist.github.com/vishnuchalla/85f670152ac46ac898779ec301459a1b

### DIFF
--- a/docs/measurements.md
+++ b/docs/measurements.md
@@ -77,7 +77,7 @@ Where `quantileName` matches with the pod conditions and can be:
 - `Ready`: The pod is able to service requests and should be added to the load balancing pools of all matching services.
 
 !!! note
-    We also log the confidence level and the errorRate of the latencies for user's understanding. Currently the threshold for the errorRate is 10% and we do not log latencies if the error is > 10% which indicates a problem with environment.(i.e system under test)
+    We also log the errorRate of the latencies for user's understanding. It indicates the percentage of pods out of all pods in the workload that got errored during the latency calculations. Currently the threshold for the errorRate is 10% and we do not log latencies if the error is > 10% which indicates a problem with environment.(i.e system under test)
 
 !!! info
     More information about the pod conditions can be found at the [kubernetes documentation site](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions).
--- a/docs/measurements.md
+++ b/docs/measurements.md
@@ -77,7 +77,7 @@ Where `quantileName` matches with the pod conditions and can be:
 - `Ready`: The pod is able to service requests and should be added to the load balancing pools of all matching services.
 
 !!! note
-    There are a V2 version of these latencies as well. Both are being kept with the intention of monitoring them over time (i.e precision vs accuracy problem) Therefore, if you notice a significant discrepancy and want to tell us about it, please feel free to do so. Or else prefer to stay with the non-v2 and everything will continue to be the same as it always has been.
+    We also log the confidence level and the errorRate of the latencies for user's understanding. Currently the threshold for the errorRate is 10% and we do not log latencies if the error is > 10% which indicates a problem with environment.(i.e system under test)
 
 !!! info
     More information about the pod conditions can be found at the [kubernetes documentation site](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions).
--- a/pkg/measurements/pod_latency.go
+++ b/pkg/measurements/pod_latency.go
@@ -39,28 +39,23 @@ const (
 )
 
 type podMetric struct {
-	Timestamp                time.Time `json:"timestamp"`
-	CreationTimestampV2      time.Time `json:"creationTimestampV2"`
-	scheduled                time.Time
-	SchedulingLatency        int `json:"schedulingLatency"`
-	SchedulingLatencyV2      int `json:"schedulingLatency_v2"`
-	initialized              time.Time
-	InitializedLatency       int `json:"initializedLatency"`
-	InitializedLatencyV2     int `json:"initializedLatency_v2"`
-	containersReady          time.Time
-	ContainersReadyLatency   int `json:"containersReadyLatency"`
-	ContainersReadyLatencyV2 int `json:"containersReadyLatency_v2"`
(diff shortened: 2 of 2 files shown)


docs/measurements.md
```diff
@@ -77,7 +77,7 @@ Where `quantileName` matches with the pod conditions and can be:
 - `Ready`: The pod is able to service requests and should be added to the load balancing pools of all matching services.
 
 !!! note
-    There are a V2 version of these latencies as well. Both are being kept with the intention of monitoring them over time (i.e precision vs accuracy problem) Therefore, if you notice a significant discrepancy and want to tell us about it, please feel free to do so. Or else prefer to stay with the non-v2 and everything will continue to be the same as it always has been.
+    We also log the confidence level and the errorRate of the latencies for user's understanding. Currently the threshold for the errorRate is 10% and we do not log latencies if the error is > 10% which indicates a problem with environment.(i.e system under test)
```
smalleni, line 80: I think the term confidence level may not be appropriate here or overloaded. If you are defining the errorRate as erroredPods/totalPods * 100, we may need to use some other term for 100-errorRate, because in statistics confidence level has a different meaning in the context of confidence intervals.

rsevilla87, line 80: Please, explain what errorRate is, and why is necessary

pkg/measurements/pod_latency.go
```diff
@@ -166,7 +159,11 @@ func (p *podLatency) start(measurementWg *sync.WaitGroup) {
 func (p *podLatency) stop() error {
 	var err error
 	p.watcher.StopWatcher()
-	p.normalizeMetrics()
+	errorRate := p.normalizeMetrics()
+	if errorRate > 10.00 {
+		log.Info("Latency errors beyond 10%. Hence invalidating the results")
+		return fmt.Errorf("Something is wrong with system under test. Pod latencies confidencelevel: %f, errorRate: %f", (100.00 - errorRate), errorRate)
```
rsevilla87, line 165:
```suggestion
		return fmt.Errorf("Pod latencies errorRate was: %2.f", errorRate)
```

docs/measurements.md
vishnuchalla: Removing this term altogether to avoid confusion.

vishnuchalla: Yes, errorRate should be able explain things. Changed as suggested.

pkg/measurements/pod_latency.go
```diff
@@ -346,3 +318,16 @@ func (p *podLatency) validateConfig() error {
 	}
 	return nil
 }
+
+// validatePod validates a pod based on job type and returns its timestamp for latency calculation.
```
vishnuchalla, line 322: Added this function as it validates on what pods to be considered and captures correct timestamps based on `jobType`, i.e both creation and the rest jobs.

pkg/measurements/pod_latency.go
```diff
@@ -166,7 +159,11 @@ func (p *podLatency) start(measurementWg *sync.WaitGroup) {
 func (p *podLatency) stop() error {
 	var err error
 	p.watcher.StopWatcher()
-	p.normalizeMetrics()
+	errorRate := p.normalizeMetrics()
+	if errorRate > 10.00 {
+		log.Info("Latency errors beyond 10%. Hence invalidating the results")
```
rsevilla87, line 164:
```suggestion
		log.Error("Latency errors beyond 10%. Hence invalidating the results")
```

docs/measurements.md
```diff
@@ -77,7 +77,7 @@ Where `quantileName` matches with the pod conditions and can be:
 - `Ready`: The pod is able to service requests and should be added to the load balancing pools of all matching services.
 
 !!! note
-    There are a V2 version of these latencies as well. Both are being kept with the intention of monitoring them over time (i.e precision vs accuracy problem) Therefore, if you notice a significant discrepancy and want to tell us about it, please feel free to do so. Or else prefer to stay with the non-v2 and everything will continue to be the same as it always has been.
+    We also log the errorRate of the latencies for user's understanding. It indicates the percentage of pods out of all pods in the workload that got errored during the latency calculations. Currently the threshold for the errorRate is 10% and we do not log latencies if the error is > 10% which indicates a problem with environment.(i.e system under test)
```
jtaleric, line 80: Thoughts on below: 

We also log the errorRate of the latencies. It indicates the percentage of pods that got errored during the latency calculations. The threshold for the errorRate is 10% and we do not log latencies if the error is > 10% which indicates a problem with environment.

pkg/measurements/pod_latency.go
```diff
@@ -166,7 +159,11 @@ func (p *podLatency) start(measurementWg *sync.WaitGroup) {
 func (p *podLatency) stop() error {
 	var err error
 	p.watcher.StopWatcher()
-	p.normalizeMetrics()
+	errorRate := p.normalizeMetrics()
+	if errorRate > 10.00 {
```
jtaleric, line 163: Do we ant to fix this in code, or allow the user to override? I can't think of why a user would want values with a higher error rate... Just a thought.

rsevilla87: Same, I think we can leave it hardcoded, 10 percent is a sane number IMO. I wouldn't expect this situation to be triggered very often.

vishnuchalla: Yes, +1 on leaving it hardcoded for now as this situation is not going to occur very often.

PR #451, opened 2023-08-29
Adding index subcommand for ocp wrapper

vishnuchalla:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

Adding index sub command for OCP wrapper as well.

- Related Issue #
- Closes # https://github.com/cloud-bulldozer/kube-burner/issues/385

## Testing
Full logs for testing: https://gist.github.com/vishnuchalla/cee74f8bd9349107c027c5a384d8c7c9


pkg/workloads/index.go
```diff
@@ -0,0 +1,96 @@
...
+		},
+		Run: func(cmd *cobra.Command, args []string) {
+			ocpMetadata := map[string]interface{}{
+				"platform":        wh.Metadata.Platform,
+				"ocpVersion":      wh.Metadata.OCPVersion,
+				"ocpMajorVersion": wh.Metadata.OCPMajorVersion,
+				"k8sVersion":      wh.Metadata.K8SVersion,
+				"totalNodes":      wh.Metadata.TotalNodes,
+				"sdnType":         wh.Metadata.SDNType,
+			}
```
rsevilla87, line 49: Why don't you execute something like

```shell
	ocpMetadata, err := ocpmetadata.NewMetadata(restConfig)
	if err != nil {
		log.Fatal(err.Error())
	}
    ocpMetadata.GetClusterMetadata()
```

pkg/workloads/index.go
```diff
@@ -0,0 +1,96 @@
...
+	"time"
+
+	"github.com/cloud-bulldozer/go-commons/indexers"
+	"github.com/cloud-bulldozer/kube-burner/pkg/util/metrics"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+// NewIndex orchestrates indexing for ocp wrapper
+func NewIndex(wh *WorkloadHelper) *cobra.Command {
```
rsevilla87, line 27: Let's avoid using WorkloadHelper here, as this is not actually a workload and can be confusing

pkg/workloads/index.go
```diff
@@ -0,0 +1,96 @@
...
+				"ocpMajorVersion": wh.Metadata.OCPMajorVersion,
+				"k8sVersion":      wh.Metadata.K8SVersion,
+				"totalNodes":      wh.Metadata.TotalNodes,
+				"sdnType":         wh.Metadata.SDNType,
+			}
+			if wh.envVars["ES_SERVER"] != "" && wh.envVars["ES_INDEX"] != "" {
+				configSpec.GlobalConfig.IndexerConfig = indexers.IndexerConfig{
+					Type:    indexers.ElasticIndexer,
+					Servers: []string{wh.envVars["ES_SERVER"]},
+					Index:   wh.envVars["ES_INDEX"],
```
rsevilla87, line 54: As I mentioned previously, we shouldn't use workloadHelper, we can access parent's cmd variables with

```golang
esServer, err:= cmd.Flags().GetString("es-server")
esIndex, err:= cmd.Flags().GetString("es-index")
```

rsevilla87:
ping @vishnuchalla

pkg/workloads/index.go
vishnuchalla: Changed as suggested

pkg/workloads/index.go
vishnuchalla: Yes, passing only variables that are required to the function.

pkg/workloads/index.go
vishnuchalla: Acknowledged

vishnuchalla:
> ping @vishnuchalla

Rewrote the logic as per suggestions. Had to expose some of the vars across the package to reuse them

pkg/workloads/index.go
```diff
@@ -0,0 +1,107 @@
...
+			if esServer != "" && esIndex != "" {
+				configSpec.GlobalConfig.IndexerConfig = indexers.IndexerConfig{
+					Type:    indexers.ElasticIndexer,
+					Servers: []string{esServer},
+					Index:   esIndex,
+				}
+			} else {
+				if metricsDirectory == "collected-metrics" {
+					metricsDirectory = metricsDirectory + "-" + uuid
+				}
```
rsevilla87, line 70: Let's remove these lines to be consistent with the regular "kube-burner index" subcommand, which doesn't appends the uuid to the directory name, this directory is also configurable

vishnuchalla: No. Even in the current kube-burner index subcommand, we have this overriden [here](https://github.com/cloud-bulldozer/kube-burner/blob/master/pkg/config/config.go#L152). It was done on a purpose to prevent metrics from being overridden across multiple runs: https://github.com/cloud-bulldozer/kube-burner/pull/442.

rsevilla87: It's not, as the index subcommand doesn't use config file. We can address that in another PR though. Thanks

rsevilla87 (approved):
lgtm

PR #454, opened 2023-09-01
Add new metrics for cpu seconds

afcollins:
## Type of change

- [X] New feature

## Description

I have found a lot of value in comparing the raw cpu_seconds aggregated at the few levels added here.

We currently index and report on the irate of these raw CPU seconds, but by using the raw seconds, we can report on the true cost of each change/component/namespace. This is just the first application of this kind of comparison.

They are important to have in all metrics profiles to make a meaningful comparison. They should be lighter than other metrics since we aren't retaining all nodes individually.

### DIFF
--- a/cmd/kube-burner/ocp-config/metrics-aggregated.yml
+++ b/cmd/kube-burner/ocp-config/metrics-aggregated.yml
@@ -138,3 +138,32 @@
 
 - query: openshift:prometheus_tsdb_head_samples_appended_total:sum{job="prometheus-k8s"}
   metricName: prometheus-ingestionrate
+
+# Retain the raw CPU seconds totals for comparison
+- query: sum( node_cpu_seconds_total and on (instance) label_replace(kube_node_role{role="worker",role!="infra"}, "instance", "$1", "node", "(.+)") ) by (mode)
+  metricName: nodeCPUSeconds-Workers
+  instant: true
+
+- query: sum( node_cpu_seconds_total and on (instance) label_replace(kube_node_role{role="master"}, "instance", "$1", "node", "(.+)") ) by (mode)
+  metricName: nodeCPUSeconds-Masters
+  instant: true
+
+- query: sum( node_cpu_seconds_total and on (instance) label_replace(kube_node_role{role="infra"}, "instance", "$1", "node", "(.+)") ) by (mode)
+  metricName: nodeCPUSeconds-Infra
+  instant: true
+
+- query: sum (  container_cpu_usage_seconds_total {  id =~ "/system.slice|/system.slice/kubelet.service|/system.slice/ovs-vswitchd.service|/system.slice/crio.service|/kubepods.slice" } and on (node) kube_node_role{ role = "worker",role != "infra" } )  by   (   id  )
+  metricName: cgroupCPUSeconds-Workers
+  instant: true
+
+- query: sum (  container_cpu_usage_seconds_total {  id =~ "/system.slice|/system.slice/kubelet.service|/system.slice/ovs-vswitchd.service|/system.slice/crio.service|/kubepods.slice" }  and on (node) kube_node_role{ role = "master" } )  by   (   id  )
+  metricName: cgroupCPUSeconds-Masters
+  instant: true
+
+- query: sum (  container_cpu_usage_seconds_total {  id =~ "/system.slice|/system.slice/kubelet.service|/system.slice/ovs-vswitchd.service|/system.slice/crio.service|/kubepods.slice" }  and on (node) kube_node_role{ role = "infra" } )  by   (   id  )
+  metricName: cgroupCPUSeconds-Infra
+  instant: true
+
+- query: sum( container_cpu_usage_seconds_total{container!~"POD|",namespace=~"openshift-.*"} )  by (namespace)
+  metricName: cgroupCPUSeconds-namespaces
+  instant: true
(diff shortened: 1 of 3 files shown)


cmd/kube-burner/ocp-config/metrics-aggregated.yml
```diff
@@ -135,3 +135,39 @@
...
+# Retain the raw CPU seconds totals for comparison
+- query: sum( node_cpu_seconds_total and on (instance) label_replace(kube_node_role{role="worker",role!="infra"}, "instance", "$1", "node", "(.+)") ) by (mode)
+  metricName: nodeCPUSeconds-Workers
+
+- query: sum( node_cpu_seconds_total  and on (instance) label_replace(kube_node_role{role="master"}, "instance", "$1", "node", "(.+)") ) by (mode)
+  metricName: nodeCPUSeconds-Masters
+
+- query: sum( node_cpu_seconds_total @end() and on (instance) label_replace(kube_node_role{role="worker",role!="infra"}, "instance", "$1", "node", "(.+)") ) by (mode)
+  metricName: nodeCPUSecondsEnd-Workers
+  instant: true
```
afcollins, line 148: I expect questions about this. The reason for this is that when metrics are indexed at a 30s step interval, this actually omits the metrics at the precise end time (`@end()`) if it does not aline on 30s from the previous measurement.

Can track it as a separate bug. But for now, this allows us to capture a reading from the precise moment the job stops.

vishnuchalla: I think we should have this implementation in the source code instead of specifying an extra query in the metrics profile.

afcollins: I agree. If we don't have a code fix for the implementation in a timely manner, then I would like to proceed to merge this so we can at least start collecting this data from runs being made in CPT.

vishnuchalla (approved):
lgtm for now. Need changes to add this logic into source code. May be we can track it through the RFE.

cmd/kube-burner/ocp-config/metrics-aggregated.yml
```diff
@@ -135,3 +135,39 @@
 
 - query: openshift:prometheus_tsdb_head_samples_appended_total:sum{job="prometheus-k8s"}
   metricName: prometheus-ingestionrate
+
+# Retain the raw CPU seconds totals for comparison
+- query: sum( node_cpu_seconds_total and on (instance) label_replace(kube_node_role{role="worker",role!="infra"}, "instance", "$1", "node", "(.+)") ) by (mode)
+  metricName: nodeCPUSeconds-Workers
+
+- query: sum( node_cpu_seconds_total  and on (instance) label_replace(kube_node_role{role="master"}, "instance", "$1", "node", "(.+)") ) by (mode)
```
rsevilla87, line 143: What's the point of this query?, we're already collecting similar data (and more) with:
https://github.com/cloud-bulldozer/kube-burner/blob/25730bdb355e23a4192a64e6cc2bb28a9bc62c28/cmd/kube-burner/ocp-config/metrics-aggregated.yml#L45

cmd/kube-burner/ocp-config/metrics-aggregated.yml
```diff
@@ -135,3 +135,39 @@
 
 - query: openshift:prometheus_tsdb_head_samples_appended_total:sum{job="prometheus-k8s"}
   metricName: prometheus-ingestionrate
+
+# Retain the raw CPU seconds totals for comparison
+- query: sum( node_cpu_seconds_total and on (instance) label_replace(kube_node_role{role="worker",role!="infra"}, "instance", "$1", "node", "(.+)") ) by (mode)
```
rsevilla87, line 140: This query will return the aggregated sum of all CPU modes from all worker nodes, is that correct?
I wonder if we can get something similar with 
https://github.com/cloud-bulldozer/kube-burner/blob/25730bdb355e23a4192a64e6cc2bb28a9bc62c28/cmd/kube-burner/ocp-config/metrics-aggregated.yml#L48
multiplied by the number of nodes

cmd/kube-burner/ocp-config/metrics-aggregated.yml
afcollins: The point is to capture the amount of CPU spent in the test. It could be considered a measure of the cost and energy consumed during the test period.

The irate query gives us the rate of change of this value (i.e. the derivative). This query is the integral of the irate query without having to integrate the rate, as that would introduce additional error unnecessarily when we can compute from the source data in prometheus.

I find that the irate can be helpful when watching metrics in real time to correlate spikes in activity with other symptoms, but they don't give us great quality or accuracy to measure our software. This metric is a step in the direction of measuring the cost and overhead of our platform, and gives us another absolute number that can be used for tracking regression over time.

I already use these queries in my analysis, notably to quantify improvements from recent OVN fixes, and the iterationsPerNamespace proposal. I add them here bc I want them collected from more runs outside of my own using custom metrics profiles, and collected in an automated way instead of me manually collecting from grafana for each time interval.

cmd/kube-burner/ocp-config/metrics-aggregated.yml
afcollins: If we did more calculation on the that avg(sum(irate)) metric, we would carry through whatever error and obfuscation happens by running that query.  My intention of capturing these raw cpu_seconds is to intentionally eliminate the irate and averaging, which hides data.

I want to capture the raw data. It gives us additional information that would be hidden by these other functions.

afcollins:
Alright. Changes are ready now and work as expected as instant queries.

afcollins:
Can we merge this please?

PR #465, opened 2023-09-19
PVC Density/Stress

### LINKED ISSUE #427: [RFE] Add storage contol plane related scenarios to ocp wrapper

It would be good to have some scenarios that create PVCs, bind them to PVs, mount them in pods and finally reclaim them.

PV/PVC related scenarios on OCP wrapper with appropriate checks that a storageclass exists in case dynamic provisioning is reqired.

**Additional context**

mfleader:
## Type of change

- [ ] Refactor
- [X] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add an experiment that creates a lot of Pods with corresponding PVCs that supports the default dynamic storage provisioning in a computing environment.

[gist of kube-burner standard output log](https://gist.github.com/mfleader/433acdac8b888e5bddccf0d842e720c1)

This software experiment would be aided by something like a PVC or PV latency calculation, similar to the pod latency calculation.

- Related Issue #427 
- Closes #427

### DIFF
--- a/cmd/kube-burner/ocp-config/pvc-density/pod.yml
+++ b/cmd/kube-burner/ocp-config/pvc-density/pod.yml
@@ -1,0 +1,34 @@
+kind: Pod
+apiVersion: v1
+metadata:
+  labels:
+    app: pvc-density-{{.Iteration}}
+  name: pod-{{.Iteration}}
+spec:
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+        - matchExpressions:
+          - key: node-role.kubernetes.io/worker
+            operator: Exists
+          - key: node-role.kubernetes.io/infra
+            operator: DoesNotExist
+          - key: node-role.kubernetes.io/workload
+            operator: DoesNotExist
+  tolerations:
+  - key: os
+    value: Windows
+    effect: NoSchedule
+  volumes:
+    - name: storage-stress-vlm-{{.Iteration}}
+      persistentVolumeClaim:
+        claimName: pvc-{{.Iteration}}
+  containers:
+  - image: {{.containerImage}}
+    name: pvc-density-container
+    resources:
+      requests:
+        memory: "10Mi"
+        cpu: "10m"
+    imagePullPolicy: IfNotPresent
--- a/pkg/workloads/pvc-density.go
+++ b/pkg/workloads/pvc-density.go
@@ -1,0 +1,64 @@
+package workloads
+
+import (
+	"fmt"
+	"os"
+	"regexp"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+var dynamicStorageProvisioners = map[string]string{
+	"aws":        "kubernetes.io/aws-ebs",
+	"cinder":     "kuberenetes.io/cinder",
+	"azure-disk": "kubernetes.io/azure-disk",
+	"azure-file": "kubernetes.io/azure-file",
+	"gce":        "kubernetes.io/gce-pd",
+	"ibm":        "powervs.csi.ibm.com",
+	"vsphere":    "kubernetes.io/vsphere-volume",
+}
+
+// NewPVCDensity holds pvc-density workload
+func NewPVCDensity(wh *WorkloadHelper) *cobra.Command {
+
+	var iterations int
+	var storageProvisioners []string
+	var claimSize string
+	var containerImage string
+	provisioner := "aws"
+
+	cmd := &cobra.Command{
+		Use:          "pvc-density",
+		Short:        "Runs pvc-density workload",
+		SilenceUsage: true,
+		PreRun: func(cmd *cobra.Command, args []string) {
+			wh.Metadata.Benchmark = cmd.Name()
+			os.Setenv("JOB_ITERATIONS", fmt.Sprint(iterations))
+			os.Setenv("CONTAINER_IMAGE", containerImage)
+			os.Setenv("CLAIM_SIZE", fmt.Sprint(claimSize))
+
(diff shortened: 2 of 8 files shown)


examples/workloads/pvc-density/templates/pod.yml
```diff
@@ -0,0 +1,36 @@
+kind: Pod
+apiVersion: v1
+metadata:
+#   labels:
+#     app: pvc-density-{{.Iteration}}
+#   name: {{.JobName}}-{{.Iteration}}
```
afcollins, line 6: I think we need this if we have >1 iteration?

examples/workloads/pvc-density/templates/pvc.yml
```diff
@@ -0,0 +1,10 @@
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: pvc-stress
```
afcollins, line 4: Do we need templated name if we have >1 iteration.

examples/workloads/pvc-density/pvc-density.yml
```diff
@@ -0,0 +1,50 @@
...
+    podWait: false
+    waitWhenFinished: true
+    namespaceLabels:
+      security.openshift.io/scc.podSecurityLabelSync: false
+      pod-security.kubernetes.io/enforce: privileged
+      pod-security.kubernetes.io/audit: privileged
+      pod-security.kubernetes.io/warn: privileged
+    objects:
+
+      - objectTemplate: templates/storageclass.yml
```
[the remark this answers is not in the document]
afcollins: This seems like the right approach. StorageClasses can have their own delays and we want to make sure it is created before we create PVCs which depend on it in the second job.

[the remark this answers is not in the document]
vishnuchalla: Understood. Correct me if I am wrong, the purpose of having both PVCs and Pods in the next job is to let a workload trigger using both and then calculate measurements for both.

examples/workloads/pvc-density/templates/pvc.yml
mfleader: I don't know

examples/workloads/pvc-density/pvc-density.yml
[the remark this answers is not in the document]
mfleader: What do you mean by 'let a workload trigger'?

examples/workloads/pvc-density/templates/pvc.yml
afcollins: Ok. I think if we create an object as part of an iteration, then it's name should be templated with the iteration number.

cmd/kube-burner/ocp-config/pvc-density/pod.yml
```diff
@@ -0,0 +1,35 @@
...
+  - key: os
+    value: Windows
+    effect: NoSchedule
+  volumes:
+    - name: storage-stress-vlm-{{.Iteration}}
+      persistentVolumeClaim:
+        claimName: pvc-{{.JobName}}-{{.Iteration}}
+  containers:
+#  - image: {{.containerImage}}
+  - image: nginx
```
afcollins, line 29: Does this pull from docker.io? Would be good to use images from quay.io cloud-bulldozer. Maybe one used in other workloads, if they'll work here.

pkg/workloads/workloads.go
```diff
@@ -8,4 +8,5 @@ var MetricsProfileMap = map[string]string{
 	"node-density":       "metrics.yml",
 	"node-density-heavy": "metrics.yml",
 	"node-density-cni":   "metrics.yml",
+	"pvc-density":        "metrics-aggregated.yml",
```
afcollins, line 11: I think metrics.yml makes sense here. Aggregate obscures node-level metrics that we might be interested in.

pkg/workloads/pvc-density.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+var dynamicStorageProvisioners = map[string]string{
+	"aws":        "kubernetes.io/aws-ebs",
+	"cinder":     "kuberenetes.io/cinder",
+	"azure-disk": "kubernetes.io/azure-disk",
+	"azure-file": "kubernetes.io/azure-file",
+	"gce":        "kubernetes.io/gce-pd",
+	"ibm":        "powervs.csi.ibm.com",
+	"vsphere":    "kubernetes.io/vsphere-volume",
```
rsevilla87, line 20: In-tree storage are unused in the latest k8s versions and will likely deprecated soon as er https://kubernetes.io/blog/2022/09/26/storage-in-tree-to-csi-migration-status-update-1.25/:

image

We should use csi based drivers instead

mfleader: I was going based off the (available dynamic provisioning plugins in OpenShift from the 4.13)[https://docs.openshift.com/container-platform/4.13/storage/dynamic-provisioning.html].

kube-burner_dynamic-provisioning-rh_docs

mfleader: Since the migration of the plugins allows you to use the deprecated method with Storage Classes, do you want to add this workload since it works, and create an issue to refactor from Storage Classes to Container Storage Interface Drivers?

mfleader: Reading more about the CSI driver migration, it sounds like I don't need to make any changes to the workload b/c the csi driver can still be configured through a storage class resource definition. For the computing environments that have enabled csi driver installation by default, it will allow more customization through a storage class resource.

I think that adapting this workload to be parameterized for every unique computing environment and csi driver is scope creep for this issue.

vishnuchalla (approved):
lgtm

vishnuchalla:
@rsevilla87 Any concerns around this PR?

PR #449, opened 2023-08-28
Functionality to scrape metrics during cleanup phase

vishnuchalla:
## Type of change

- [x] Refactor
- [x] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

* Added functionality to scrape metrics during kube-burner cleanup phase. Now jobsummary has the below fields as well
```
  "timestamp": "2023-08-29T00:17:27.942960538Z",
  "endTimestamp": "2023-08-29T00:18:15.817272025Z",
  "uuid": "83bfcb20-54f1-43f4-b2ad-ad04c2f4fd16",
  "metricName": "jobSummary",
  "elapsedTime": 48,
  "cleanupTimestamp": "2023-08-29T00:18:18.015107794Z",
  "cleanupEndTimestamp": "2023-08-29T00:18:49.014541929Z",
```
Inidicating job start and end timestamps along with start and end timestamps for cleanup phase.
* We can easily filter out metrics by adjusting those timestamps as required and --gc-metrics option is completely optional.
* And correct all out timestamps to be in UTC format to maintain consistency across the application.
* For local indexing cleanup metrics will be available in `*-cleanup` folder.
## Related Tickets & Documents

- Related Issue # https://github.com/cloud-bulldozer/kube-burner/issues/444
- Closes #

## Checklist before requesting a review

- [x] I have performed a self-review of my code.
- [ ] If it is a core feature, I have added thorough tests.

## Testing
Tested on self managed cluster
```
kube-burner ocp node-density-heavy --timeout 1h --metrics-endpoint ~/metrics-endpoints.yaml --pods-per-node=50 --log-level=debug 2>&1 --gc=true --gc-metrics=true | tee -a kube-burner-$(date +"%F_%H-%M-%S")
```

PR #477, opened 2023-09-25
Refactor garbage collection indexing

rsevilla87:
## Type of change

- [x] Refactor
- [ ] New feature
- [x] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

Refactor to simplify the current code. This new approach basically adds a "fake" job to the `prometheusJobList` array, and makes garbage collection blocker in case gcMetrics is enabled.

By adding a new job to `prometheusJobList` there's no need to scrape prometheus multiple times. The metrics from the garbage collection phase should be tagged with their corresponding `jobConfig.name`. i.e:

```shell
$ ./bin/amd64/kube-burner ocp node-density --pods-per-node=150 --local-indexing --gc-metrics=true --qps=5
$ <truncated>
 jq . collected-metrics-475fb3a9-3e1a-44b3-b85b-a081512c49a7/kubeletCPU.json | grep -A2  jobConfig
    "jobConfig": {
      "jobIterations": 282,
      "name": "node-density",
--
    "jobConfig": {
      "jobIterations": 282,
      "name": "node-density",
--
    "jobConfig": {
      "jobIterations": 282,
      "name": "node-density",
--
    "jobConfig": {
      "name": "garbage-collection"
    },
--
    "jobConfig": {
      "jobIterations": 282,
      "name": "node-density",
--
    "jobConfig": {
      "jobIterations": 282,
      "name": "node-density",
--
    "jobConfig": {
      "jobIterations": 282,
      "name": "node-density",
--
    "jobConfig": {
      "name": "garbage-collection"
    },
--
    "jobConfig": {
      "jobIterations": 282,
      "name": "node-density",
--
    "jobConfig": {
      "jobIterations": 282,
      "name": "node-density",
--
    "jobConfig": {
      "jobIterations": 282,
      "name": "node-density",
--
    "jobConfig": {
      "name": "garbage-collection"
    },

```

Also adds the missing fields to be replaced in the OCP wrapper configuration files, and fixes the `findJob` function which wasn't working well, as it always applied the last job name of the list.

Last but not least, it removes the Start/End fields from the executor struct.

- Related Issue #449 
- Closes #


pkg/burner/metadata.go
```diff
@@ -25,41 +25,34 @@ import (
 )
 
 type jobSummary struct {
-	Timestamp           time.Time              `json:"timestamp"`
-	EndTimestamp        time.Time              `json:"endTimestamp"`
-	UUID                string                 `json:"uuid"`
-	MetricName          string                 `json:"metricName"`
-	ElapsedTime         float64                `json:"elapsedTime"`
-	CleanupTimestamp    time.Time              `json:"cleanupTimestamp"`
```
vishnuchalla, line 33: Do we want to consider publishing cleanup start and end times in our `jobSummary` too?

vishnuchalla:
In overall, I like this new approach for scraping end metrics through an extra job, but I think we also need to consider the cleanup functions being called at the end of `burner.end()` method to capture all the cleanup metrics.

pkg/prometheus/prometheus.go
```diff
@@ -115,7 +120,7 @@ func (p *Prometheus) ScrapeJobsMetrics(indexer *indexers.Indexer) error {
 func (p *Prometheus) findJob(timestamp time.Time) config.Job {
 	var jobConfig config.Job
 	for _, prometheusJob := range p.JobList {
-		if timestamp.Before(prometheusJob.End) || timestamp.Equal(prometheusJob.End) {
+		if timestamp.After(prometheusJob.Start) || timestamp.Equal(prometheusJob.Start) {
```
vishnuchalla, line 123: Isn't it the same logic other way round? Just trying to understand on how is it different from current logic.

pkg/burner/job.go
```diff
@@ -192,16 +188,49 @@ func Run(configSpec config.Spec, prometheusClients []*prometheus.Prometheus, ale
...
 		if globalConfig.GC {
-			go CleanupNamespaces(context.TODO(), metav1.ListOptions{LabelSelector: fmt.Sprintf("kube-burner-uuid=%v", uuid)}, false)
+			cleanupStart := time.Now().UTC()
+			// If gcMetrics is enabled, garbage collection must be blocker
+			if globalConfig.GCMetrics {
+				CleanupNamespaces(context.TODO(), metav1.ListOptions{LabelSelector: fmt.Sprintf("kube-burner-uuid=%v", uuid)}, true)
+				// We add an extra dummy job to prometheusJobList to index metrics from this stage
+				prometheusJobList = append(prometheusJobList, prometheus.Job{
+					Start: cleanupStart,
+					End:   time.Now().UTC(),
```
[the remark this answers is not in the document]
rsevilla87: Good call, I've updated the expression you linked to

```golang
if globalConfig.GC && !globalConfig.GCMetrics {
```

Wit this change the behavior should be:
- When gc == true and GCMetrics == true
Run a blocking cleanup operation and add a new prometheus job to the list
When this operation finishes, index metrics, including the cleanup ones.
- When gc == true and GCMetrics == false
Run cleanup in its own goroutine, at the end of the benchmark, run a blocking cleanup operation again to ensure there's not stuff left

pkg/burner/metadata.go
rsevilla87: All the indexed metrics are already tagged with garbage-collection, do we really need this? cc: @afcollins

rsevilla87:
> In overall, I like this new approach for scraping end metrics through an extra job, but I think we also need to consider the cleanup functions being called at the end of `burner.run()` method to capture all the cleanup metrics.

Added an explanation to the in-line message, hope that helps

pkg/prometheus/prometheus.go
rsevilla87: I'm gonna try to explain it:

When you have a list of prometheusJobs [j1, j2, j3], with the following start/end timestamps:  [j1s_j1e, j2s_j2e, j3s_j3e]
With the old logic:
Metrics from j1 and j2 will always be tagged with the name of the latest job of the list (this list is actually sorted from newer to older jobs), in this case `j3`, because the timestamps of its metrics always will be before j3.End.

With this change, we verify that metrics are  => than the jobs start timestamps, so the metrics from job1 will be tagged with j1 and not with j2 or j3 because this metrics won't be after j2 or j3 timestamps

pkg/burner/metadata.go
afcollins: Yes. I collect the cleanupStart and cleanupEnd timestamps from kube-burner logs anyway. It would be helpful to have them to know when each action is triggered since there is a lot of activity that starts at cleanupStart.

pkg/burner/job.go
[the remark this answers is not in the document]
vishnuchalla: Yes sound good. But when gc==true and GCMetrics==true do we also want to cleanup non-namespaced resources as well? Because they will be left out if GCMetrics==true. Is it something expected?
cc: @afcollins

pkg/prometheus/prometheus.go
vishnuchalla: I don't see our job list being sorted in reverse anywhere. All the jobs are getting appended in [j1, j2, j3] etc order. Am I missing on something? still feel like its the same logic other way around, because a metric timestamp let's say a j1 query from prom results will always fall under the specific job's start and end timestamps (i.e j1's).

pkg/burner/job.go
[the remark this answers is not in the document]
afcollins: I agree, looks like the non-namespaced resources are left out in this case.  I think we ought to have a single cleanup method that includes namespaces and non-namespaced so we don't make different distinctions these disparate places.

Also wondering whether it makes more sense for GC to always run as its own delete job, instead of conditional logic to handle it differently.

pkg/prometheus/prometheus.go
rsevilla87: With the previous logic, a metric from j1 query, will fall under j1, j2 and j3 because that metric was **Before** j1.End, j2.End and j3.End, with the new logic, that metric falls only under j1.Start

pkg/burner/job.go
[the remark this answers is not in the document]
rsevilla87: The previous logic:

-  created two indexers, one for the the regular workload and another one for the "garbage-colllection" stage. One of the intentions of this PR is to consolidate that work into a single indexer.
- All the indexed metrics from `garbage-collection`, where tagged with the job's name, that's not strictly correct, garbage-collection is not actually part of a job, but something that runs at the end of a list of jobs, for that reason I think considering it like another job when indexing fits better.
- When using local indexing, garbage-collection metrics are being created in a new directory. In the original implementation all metrics are stored in the same directory, why to create a new one?

wrt CleanupNonNamespacedResourcesUsingGVR, I'll add it back, I missed it

pkg/burner/metadata.go
rsevilla87: You can filter metrics by job name too.

```
uuid.keyword: $uuid AND metricName.keyword: blabla AND jobConfig.name.keyword: garbage-collection
```

afcollins: That's helpful, but it still doesn't give me a measurement at exactly cleanupStart.

The first metric labeled "garbage-collection" is at a time that is the "first step that lands beyond the cleanupStart".  In my local case, that's 7s after GC started.

```
time="2023-09-28 14:28:57" level=info msg="Deleting namespaces with label kube-burner-uuid=f5008e24-01e6-4fc4-8a02-595af5ee57b5" file="namespaces.go:64"

... collected-metrics .. 
    {
    "timestamp": "2023-09-28T18:28:34.235Z",
    "jobConfig": {
      "name": "node-density",
    },
    {
    "timestamp": "2023-09-28T18:29:04.235Z",
    "jobConfig": {
      "name": "garbage-collection"
    },
```

vishnuchalla: > That's helpful, but it still doesn't give me a measurement at exactly cleanupStart.
> 
> The first metric labeled "garbage-collection" is at a time that is the "first step that lands beyond the cleanupStart". In my local case, that's 7s after GC started.
> 
> ```
> time="2023-09-28 14:28:57" level=info msg="Deleting namespaces with label kube-burner-uuid=f5008e24-01e6-4fc4-8a02-595af5ee57b5" file="namespaces.go:64"
> 
> ... collected-metrics .. 
>     {
>     "timestamp": "2023-09-28T18:28:34.235Z",
>     "jobConfig": {
>       "name": "node-density",
>     },
>     {
>     "timestamp": "2023-09-28T18:29:04.235Z",
>     "jobConfig": {
>       "name": "garbage-collection"
>     },
> ```

@rsevilla87 Rest of the changes look good to me except for this concern.

vishnuchalla (approved):

pkg/prometheus/prometheus.go
vishnuchalla: Then in that case, shouldn't we be checking if a metrics from j1 falls between range j1.Start <= metric.timestamp() <= j1.End and so on?

pkg/burner/metadata.go
rsevilla87: Just realized that we're already indexing that :), as  `garbage-collection` is now considered a job, it has its own summary indexed:

```
$ jq . collected-metrics-76f92bf7-29a8-45e2-97f2-c510ed169235/jobSummary-garbage-collection.json 
[
  {
    "timestamp": "2023-10-02T09:52:02.424916277Z",
    "uuid": "76f92bf7-29a8-45e2-97f2-c510ed169235",
    "metricName": "jobSummary",
    "elapsedTime": 63,
    "jobConfig": {
      "name": "garbage-collection"
    },
    "metadata": {
      "k8sVersion": "v1.25.8+37a9a08",
      "ocpMajorVersion": "4.12",
      "ocpVersion": "4.12.18",
      "platform": "AWS",
      "sdnType": "OpenShiftSDN",
      "totalNodes": 3
    },
    "version": "refactor-garbage-collection-indexing@f8378a490c57b71539cfb2650bd21a174811e6a9"
  }
]
```

Anyway, I updated the generated summary to also include and `endDate` (I used this field name to be consistent with the clusterMetadata document) field in case you need it

afcollins: > doesn't give me a measurement at exactly cleanupStart [1](https://github.com/cloud-bulldozer/kube-burner/pull/477#discussion_r1341561963)

This requirement is satisfied by the current implementation, but not the refactor.

As I came to realize myself in talking through this with Raul, if we add functionality to execute instant queries at key timestamps (i.e. jobStart, jobEnd), then I will be able to collect the metrics I am seeking. This is a separate RFE.

And the refactor in this PR may even allow easier collecting of those measurements, by allowing us to use "jobStart" and "jobEnd", without needing to add specific "cleanupStart/End" options.

pkg/burner/job.go
```diff
@@ -192,16 +188,59 @@ func Run(configSpec config.Spec, prometheusClients []*prometheus.Prometheus, ale
...
+				cleanupEnd := time.Now().UTC()
+				prometheusJobList = append(prometheusJobList, prometheus.Job{
+					Start: cleanupStart,
+					End:   cleanupEnd,
+					JobConfig: config.Job{
+						Name: garbageCollectionJob,
+					},
+				})
+			} else {
+				go CleanupNonNamespacedResourcesUsingGVR(context.TODO(), jobList, true)
```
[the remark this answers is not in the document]
vishnuchalla: So these ones are to trigger garbage collection in parallel right after the benchmark is completed when `gc-metrics` option is not  selected. This doesn't change any of existing behaviour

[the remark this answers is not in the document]
rsevilla87: right, no specific reason :)

pkg/burner/metadata.go
vishnuchalla: @afcollins @rsevilla87 This PR is working as expected. It simply calculate the metrics within the range of (jobList[0].Start and JobList[len(jobList)-1].End) and applies step in that range. That is why we were only able to see metrics excluding metrics at start/end of some of the jobs that do not fall under the step interval in between. In order to fix this, we need a code change that captures metrics per job (i.e in their JobStart and JobEnd range) and makes sure runs an instant query for the end if interval is not perfectly divisible by step size. Here is the PR: https://github.com/cloud-bulldozer/kube-burner/pull/484

vishnuchalla:
No we still need this PR changes.

PR #480, opened 2023-09-28
Implementing profile-type

### LINKED ISSUE #479: [RFE] add report metrics with current metric collection
No -- well, sort of. 

Have kube-burner report both the per-second metrics and the aggregated result into ES.

Trying to pull 100000000 metrics from ES is awful and slow, and I just want an aggregated number which kube-burner can query from prom to get.

rsevilla87:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [x] Documentation Update

## Description

Implementing profile-type to allow specifying the type of metrics profile to use in a single run

- Related Issue #
- Closes #479

### DIFF
--- a/cmd/kube-burner/ocp.go
+++ b/cmd/kube-burner/ocp.go
@@ -53,7 +53,7 @@ func openShiftCmd() *cobra.Command {
 	gcMetrics := ocpCmd.PersistentFlags().Bool("gc-metrics", false, "Collect metrics during garbage collection")
 	userMetadata := ocpCmd.PersistentFlags().String("user-metadata", "", "User provided metadata file, in YAML format")
 	extract := ocpCmd.PersistentFlags().Bool("extract", false, "Extract workload in the current directory")
-	profileType := ocpCmd.PersistentFlags().String("profile-type", "regular", "Metrics profile to use, supported options are: regular, reporting or both")
+	profileType := ocpCmd.PersistentFlags().String("profile-type", "both", "Metrics profile to use, supported options are: regular, reporting or both")
 	ocpCmd.MarkFlagsRequiredTogether("es-server", "es-index")
 	ocpCmd.MarkFlagsMutuallyExclusive("es-server", "local-indexing")
 	ocpCmd.PersistentPreRun = func(cmd *cobra.Command, args []string) {
--- a/cmd/kube-burner/ocp.go
+++ b/cmd/kube-burner/ocp.go
@@ -53,7 +53,7 @@ func openShiftCmd() *cobra.Command {
 	gcMetrics := ocpCmd.PersistentFlags().Bool("gc-metrics", false, "Collect metrics during garbage collection")
 	userMetadata := ocpCmd.PersistentFlags().String("user-metadata", "", "User provided metadata file, in YAML format")
 	extract := ocpCmd.PersistentFlags().Bool("extract", false, "Extract workload in the current directory")
-	reporting := ocpCmd.PersistentFlags().Bool("reporting", false, "Enable benchmark report indexing")
+	profileType := ocpCmd.PersistentFlags().String("profile-type", "regular", "Metrics profile to use, supported options are: regular, reporting or both")
 	ocpCmd.MarkFlagsRequiredTogether("es-server", "es-index")
 	ocpCmd.MarkFlagsMutuallyExclusive("es-server", "local-indexing")
 	ocpCmd.PersistentPreRun = func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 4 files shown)


rsevilla87:
I'm not very convinced of the naming of the new flag `--profile-type` but I'm out of ideas today, open to suggestions :)

cmd/kube-burner/ocp.go
```diff
@@ -53,7 +53,7 @@ func openShiftCmd() *cobra.Command {
 	gcMetrics := ocpCmd.PersistentFlags().Bool("gc-metrics", false, "Collect metrics during garbage collection")
 	userMetadata := ocpCmd.PersistentFlags().String("user-metadata", "", "User provided metadata file, in YAML format")
 	extract := ocpCmd.PersistentFlags().Bool("extract", false, "Extract workload in the current directory")
-	reporting := ocpCmd.PersistentFlags().Bool("reporting", false, "Enable benchmark report indexing")
+	profileType := ocpCmd.PersistentFlags().String("profile-type", "regular", "Metrics profile to use, supported options are: regular, reporting or both")
```
jtaleric, line 56: can we default to both?

docs/ocp.md
```diff
@@ -164,11 +164,9 @@ With the help of [networkpolicy](https://kubernetes.io/docs/concepts/services-ne
...
-
-This feature is very useful to avoid sending thousands of documents to the configured indexer, as only a few documents will be indexed per benchmark. The metrics profile used by this feature is defined in [metrics-report.yml](https://github.com/cloud-bulldozer/kube-burner/blob/master/cmd/kube-burner/ocp-config/metrics-report.yml))
+By specifying `--profile-type`, kube-burner can use two different metrics-profile when scraping metrics from prometheus. By default uses the `regular` one which is bound to the workload in question, but it's possible to use [metrics-report.yml](https://github.com/cloud-bulldozer/kube-burner/blob/master/cmd/kube-burner/ocp-config/metrics-report.yml), specified by `reporting`, which is a profile very useful to reduce the number of documents sent to the configured indexer. Thanks to the combination of aggregations and instant queries for prometheus metrics, and 4 summaries for pod latency measurements, only a few documents will be indexed per benchmark. It's also possible to use `both` profiles in the same run.
```
jtaleric, line 169: Lets default to both 🚢

cmd/kube-burner/ocp.go
rsevilla87: Done

rsevilla87:
As I said previously, I'm not a big fan of this change, the initial intention of creating a "report" metrics profile was to reduce the overall number of indexed documents in periodic runs (cpt) and use the "regular" profiles for debugging/troubleshooting purposes **when needed**

Using both by default,  goes against that philosophy 🤷🏽‍♂️

afcollins:
> Using both by default, goes against that philosophy

I am not confident we can rely solely on the metrics report. Running both allows us to build confidence and close any gaps in the metrics-report without losing more data that would require us to re-run.

It's a step towards enabling reporting-only mode. We will get there. :)

vishnuchalla:
Looks like we need to add/modify CI tests for this new use case too.

pkg/workloads/helpers.go
```diff
@@ -200,19 +208,30 @@ func (wh *WorkloadHelper) run(workload, metricsProfile string) {
...
-		if wh.reporting {
-			metricsProfile = reportProfile
-			for i := range configSpec.GlobalConfig.Measurements {
-				configSpec.GlobalConfig.Measurements[i].PodLatencyMetrics = types.Quantiles
-			}
-		}
-		metricsEndpoints = append(metricsEndpoints, prometheus.MetricEndpoint{
+		regularProfile := prometheus.MetricEndpoint{
 			Endpoint:     wh.prometheusURL,
 			AlertProfile: alertsProfile,
```
vishnuchalla, line 213: Here if someone selects only reporting mode, alerts profile is not available. So do we want to make it available for regular and report profiles individually and when selected both we will evaluate alerts only once.

rsevilla87: right, i forgot adding it

jtaleric:
> As I said previously, I'm not a big fan of this change, the initial intention of creating a "report" metrics profile was to reduce the overall number of indexed documents in periodic runs (cpt) and use the "regular" profiles for manual runs and/or debugging/troubleshooting purposes **when needed**

ack - understand the reasoning behind the report.

> Using both by default, goes against that philosophy 🤷🏽‍♂️

True. However, I don't think it makes sense to go cold-turkey (ie flip the bit before we get confidence). 

With Prow, we could eventually just depend on the must-gather to collect the promdump -- but that even isn't a 100% foolproof method.

vishnuchalla (approved):
lgtm

rsevilla87:
> > As I said previously, I'm not a big fan of this change, the initial intention of creating a "report" metrics profile was to reduce the overall number of indexed documents in periodic runs (cpt) and use the "regular" profiles for manual runs and/or debugging/troubleshooting purposes **when needed**
> 
> ack - understand the reasoning behind the report.
> 
> > Using both by default, goes against that philosophy 🤷🏽‍♂️
> 
> True. However, I don't think it makes sense to go cold-turkey (ie flip the bit before we get confidence).
> 
> With Prow, we could eventually just depend on the must-gather to collect the promdump -- but that even isn't a 100% foolproof method.

100%, let's try to make a soft transition :)

PR #483, opened 2023-10-03
Adding instantJobStart feature

rsevilla87:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adding `instantJobStart` so we can fetch the first datapoint of a timeseries during a job.

- Related Issue #
- Closes [#](https://github.com/cloud-bulldozer/kube-burner/issues/482)

### DIFF
--- a/pkg/prometheus/prometheus.go
+++ b/pkg/prometheus/prometheus.go
@@ -57,12 +57,13 @@ func NewPrometheusClient(configSpec config.Spec, url string, auth Auth, step tim
 
 // ScrapeJobsMetrics gets all prometheus metrics required and handles them
 func (p *Prometheus) ScrapeJobsMetrics(indexer *indexers.Indexer) error {
-	start := p.JobList[0].Start
-	end := p.JobList[len(p.JobList)-1].End
-	elapsed := int(end.Sub(start).Seconds())
+	var instantTimestamp time.Time
 	var err error
 	var v model.Value
 	var renderedQuery bytes.Buffer
+	start := p.JobList[0].Start
+	end := p.JobList[len(p.JobList)-1].End
+	elapsed := int(end.Sub(start).Seconds())
 	vars := util.EnvToMap()
 	vars["elapsed"] = fmt.Sprintf("%ds", elapsed)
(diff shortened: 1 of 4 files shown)


vishnuchalla:
This extra configuration is not required as we can achieve the same in the rewamp logic: https://github.com/cloud-bulldozer/kube-burner/pull/484

vishnuchalla (changes requested):
This extra configuration is not required as we can achieve the same in the rewamp logic: https://github.com/cloud-bulldozer/kube-burner/pull/484. Any thoughts? @rsevilla87

rsevilla87:
> This extra configuration is not required as we can achieve the same in the rewamp logic: #484. Any thoughts? @rsevilla87

honestly, I don't understand why are we duplicating some work.

afcollins:
I synced with Vishnu and this PR is still 1 of 3 changes required. #484 serves the purpose of fixing the end timestamp measurement.

This PR allows us to run an instant query at the start of a job boundary and is still required. I have some comments on the code I will leave in a review.

pkg/prometheus/prometheus.go
```diff
@@ -77,7 +78,12 @@ func (p *Prometheus) ScrapeJobsMetrics(indexer *indexers.Indexer) error {
 		renderedQuery.Reset()
 		if md.Instant {
 			log.Debugf("Instant query: %s", query)
-			if v, err = p.Client.Query(query, end); err != nil {
+			if md.InstantJobStart {
+				instantTimestamp = start
+			} else {
+				instantTimestamp = end
+			}
+			if v, err = p.Client.Query(query, instantTimestamp); err != nil {
```
afcollins, line 86: This is good. It allows me to make an instant query at the start of a job, and I can work with it.

Putting some feedback here in case it influences any implementation decisions:

Since I want to query a metric at the start AND end of a job, then I would need two separate metrics definitions: 1) `instant=true` and 2) `instant=true` and `instantJobStart=true`.

For how I am planning to use it, I think this could be a preferable outcome, because then I can query them as separate "columns" (if I am using benchmark-comparison), and then subtract the two columns from one another.  However, it does require I add two metrics definitions for each metric delta I want to capture.

PR #491, opened 2023-10-06
skipIndexing feature

rsevilla87:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [x] Optimization
- [x] Documentation Update

## Description

Ability to skip indexing on certain jobs, useful to prevent capturing datapoints from setup-up jobs

@mohit-sheth I've taken the liberty of adding this flag to the networkPolicy workloads, wdyt?

@josecastillolema I think this might be useful for web-burner as there're some set-up steps in the jobs you probably don't need to capture

- Related Issue #
- Closes #

### DIFF
--- a/cmd/kube-burner/ocp-config/networkpolicy-matchexpressions/networkpolicy-matchexpressions.yml
+++ b/cmd/kube-burner/ocp-config/networkpolicy-matchexpressions/networkpolicy-matchexpressions.yml
@@ -12,6 +12,7 @@ global:
 jobs:
   - name: create-bindings
     namespace: networkpolicy-matchexpressions
+    skipIndexing: true
     jobIterations: 1
     objects:
       - objectTemplate: ../clusterrole.yml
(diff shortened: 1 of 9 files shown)


pkg/prometheus/types.go
```diff
@@ -43,7 +43,6 @@ type Prometheus struct {
 type Job struct {
 	Start     time.Time
 	End       time.Time
-	Name      string
```
vishnuchalla, line 46: Can we take this nits in a different PR?

pkg/util/metrics/metrics.go
```diff
@@ -111,7 +111,6 @@ func ProcessMetricsScraperConfig(metricsScraperConfig ScraperConfig) Scraper {
 			p.JobList = []prometheus.Job{{
 				Start: time.Unix(metricsScraperConfig.StartTime, 0),
 				End:   time.Unix(metricsScraperConfig.EndTime, 0),
-				Name:  metricsScraperConfig.JobName,
```
vishnuchalla, line 114: Looks like this one requires changes at different places as well, from where we are passing this arg.

vishnuchalla:
Do we want to skip [clusterMetadata](https://github.com/cloud-bulldozer/kube-burner/blob/master/pkg/workloads/helpers.go#L261) as well with this flag?

rsevilla87:
> Do we want to skip [clusterMetadata](https://github.com/cloud-bulldozer/kube-burner/blob/master/pkg/workloads/helpers.go#L261) as well with this flag?

If you want to explude cluster's metadata, you should disable indexing :)

vishnuchalla (approved):
lgtm

PR #502, opened 2023-11-05
Add network-policy workload

npinaeva:
## Type of change

- [ ] Refactor
- [X] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Create network-policy scale testing framework and 2 profiles for
openshift and ovn-kubernetes.

kube-burner modification: add ShutdownCondition to jobConfig. This
allows to set up a command that should exit before the workload
can be deleted. It is used by the network policy workload, since network
policy doesn't have a status, and extra means to track when the config
is applied are needed.
Add Binomial, IndexToCombination and GetSubnet24 functions for
templating, since they are needed to generate complex network policy
configs.

- Related Issue #
- Closes #

- [X] I have performed a self-review of my code.
- [?] If it is a core feature, I have added thorough tests.

### DIFF
--- a/pkg/config/types.go
+++ b/pkg/config/types.go
@@ -102,6 +102,8 @@ type Job struct {
 	JobIterationDelay time.Duration `yaml:"jobIterationDelay" json:"jobIterationDelay,omitempty"`
 	// JobPause how much time to pause after finishing the job
 	JobPause time.Duration `yaml:"jobPause" json:"jobPause,omitempty"`
+	// BeforeCleanup allows to run a bash script before the workload is deleted.
+	BeforeCleanup string `yaml:"beforeCleanup" json:"beforeCleanup,omitempty"`
 	// Name job name
 	Name string `yaml:"name" json:"name,omitempty"`
 	// Objects list of objects
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -39,35 +39,36 @@ kube-burner connects k8s clusters using the following methods in this order:
 
 This section contains the list of jobs `kube-burner` will execute. Each job can hold the following parameters.
 
-| Option               | Description                                                                      | Type    | Default |
-|----------------------|----------------------------------------------------------------------------------|---------|----------|
-| `name`                 | Job name                                                                         | String  | ""      |
-| `jobType`              | Type of job to execute. More details at [job types](#job-types)                  | string  | create  |
-| `jobIterations`        | How many times to execute the job                                                | Integer | 0       |
-| `namespace`            | Namespace base name to use                                                       | String  | ""      |
-| `namespacedIterations` | Whether to create a namespace per job iteration                                  | Boolean | true    |
-| `iterationsPerNamespace` | The maximum number of `jobIterations` to create in a single namespace. Important for node-density workloads that create Services.                                  | Integer | 1    |
-| `cleanup`              | Cleanup clean up old namespaces                                                  | Boolean | true    |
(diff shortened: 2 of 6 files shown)


rsevilla87 (commented):
Thanks for this PR @npinaeva , I find the new shutdownCondition parameter very interesting, but I don't think the naming is correct, this parameter allows to execute an arbitrary command in the OS when a job finishes, the user can implement anything on this command and not only shutdown conditions, i.e:

```yaml
jobs:
    - name: myjob
      shutdownCondition: sudo rm -rf  /
```

Can you rename the parameter to something like, for example `endingCmd` and re write the documentation?

rsevilla87 (commented):
Given that you're adding the new workload and its resources to the examples folder, I think you should maintain this workload in a separate repo rather than adding it here

npinaeva:
> Can you rename the parameter to something like, for example `endingCmd` and re write the documentation?
what do you think about `beforeCleanup` maybe?

> Given that you're adding a new framework with all its resources to the examples folder, I think you should keep and maintain it from a separate repo rather than adding it here.

Sure, I can do that. Do you have some workload examples in their own repos?

rsevilla87:
> > Can you rename the parameter to something like, for example `endingCmd` and re write the documentation?
> 
> what do you think about `beforeCleanup` maybe?
> 
> > Given that you're adding a new framework with all its resources to the examples folder, I think you should keep and maintain it from a separate repo rather than adding it here.
> 
> Sure, I can do that. Do you have some workload examples in their own repos?

`beforeCleanup` sounds more reasonable, although job cleanup is optional :/

Wrt the workload I think you can just host the content of `examples/workloads/network-policy/` in a different repository, for example, these projects use kube-burner to run some of their tests https://github.com/stackrox/stackrox/tree/master/tests/performance/scale/tests/kube-burner https://github.com/redhat-performance/web-burner

examples/workloads/network-policy/openshift/openflow-tracker/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM python:slim
```
vishnuchalla, line 1: This folder with python stuff shouldn't go into this repo. We should keep it somewhere else may be `e2e` and then just point out the container image in job configuration files.

cc: @mohit-sheth ^^

examples/workloads/network-policy/ovn-kubernetes/openflow-tracker/openflow-tracker.py
```diff
@@ -0,0 +1,93 @@
+import datetime
```
vishnuchalla, line 1: Same here as well, no python code into this repo. Recommending to keep it at a different and to only used the container image generated using it.

pkg/config/types.go
```diff
@@ -102,6 +102,8 @@ type Job struct {
 	JobIterationDelay time.Duration `yaml:"jobIterationDelay" json:"jobIterationDelay,omitempty"`
 	// JobPause how much time to pause after finishing the job
 	JobPause time.Duration `yaml:"jobPause" json:"jobPause,omitempty"`
+	// BeforeCleanup allows to run a bash script before the workload is deleted.
+	BeforeCleanup string `yaml:"beforeCleanup" json:"shutdownCondition,omitempty"`
```
rsevilla87, line 106:
```suggestion
	BeforeCleanup string `yaml:"beforeCleanup" json:"beforeCleanup,omitempty"`
```

PR #516, opened 2023-11-17
Service latency measurement

### LINKED ISSUE #467: [RFE] New measurement: Service creation latency

When a service is created it takes some time for the CNI plugin to configure it and make it functional.
Adding a logic able to measure the time taken for a service to be ready to receive requests can be very useful in order to establish new SLOs in our benchmarks

TBD

TBD

**Additional context**
TBD

rsevilla87:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [x] Documentation Update

## Description

Service ready latency measurement, please take a look at the updated docs to get more information about this feature

image

- Closes #467 


vishnuchalla:
Can we start adapting the culture of atomic PRs please? This PR along with ServiceLatency Measurement has lot of refactoring and other changes at the implementation level which makes it hard to review and ensure that nothing in the existing code breaks.

pkg/burner/namespaces.go
```diff
@@ -20,65 +20,13 @@ import (
...
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
 )
 
-func createNamespace(namespaceName string, nsLabels map[string]string) error {
```
vishnuchalla, line 32: Why are we keeping some of these functions here at `./namespaces.go` and some of them in the `utils/namespaces.go`? Can't we just move all of them to once place. may be into utils 🤔

pkg/measurements/types/types.go
```diff
@@ -16,6 +16,10 @@ package types
 
 import (
```
vishnuchalla, line 17: Can we just have this `types.go` at root level of this package?

pkg/measurements/metrics/metrics.go
```diff
@@ -74,3 +75,34 @@ func CheckThreshold(thresholds []types.LatencyThreshold, quantiles []interface{}
...
+	val, _ = stats.Max(input)
+	latencyQuantiles.Max = int(val)
+	val, _ = stats.Mean(input)
+	latencyQuantiles.Avg = int(val)
+	return latencyQuantiles
+}
+
+// SetQuantileV2 sets percentiles
+func (plq *LatencyQuantiles) SetQuantileV2(input []float64, percentile float64) {
+	stats.Percentile(input, percentile)
```
vishnuchalla, line 96: Is this statement left over by mistake after testing? I am not sure how this is adding any value in collecting the percentiles when we already have a switch block below to store `plq 50, 95, 90` metrics.

pkg/burner/utils.go
```diff
@@ -17,19 +17,18 @@ package burner
 import (
```
vishnuchalla, line 17: **[Directory structure]:** Same about this `utils.go` too. This new directory structure looks a bit complex to me having files with redundant names and purpose too. Also adds more sub-directories which I believe are not required at this moment.

pkg/measurements/metrics/metrics.go
```diff
@@ -74,3 +75,34 @@ func CheckThreshold(thresholds []types.LatencyThreshold, quantiles []interface{}
 	}
 	return utilerrors.NewAggregate(errs)
 }
+
+func NewLatencySummary(input []float64) LatencyQuantiles {
```
vishnuchalla, line 79: This function and method below it more or less do the same things. I suspect if one of them is redundant.

pkg/measurements/pod_latency.go
```diff
@@ -128,23 +130,40 @@ func (p *podLatency) handleUpdatePod(obj interface{}) {
 }
 
 func (p *podLatency) setConfig(cfg types.Measurement) error {
```
vishnuchalla, line 132: This method by name it says `setConfig` and sets the configuration. But alongside, it also checks if the metric is found on or not. I think the metrics check should be decoupled.

pkg/measurements/service_latency.go
```diff
@@ -0,0 +1,336 @@
...
+				return true, nil
+			}
+		}
+		return false, nil
+	})
+	return err
+}
+
+func (s *serviceLatency) waitForIngress(svc *corev1.Service) error {
+	err := wait.PollUntilContextCancel(context.TODO(), 100*time.Millisecond, true, func(ctx context.Context) (done bool, err error) {
```
vishnuchalla, line 321: `100*time.Millisecond`, regarding these poll time intervals, we do have different values at different places. Do we want to decide on keeping same values across?

vishnuchalla (dismissed):
Also needs a rebase please.

pkg/measurements/metrics/metrics.go
rsevilla87: This new function is a helpers that uses the `stats` package to calculate the percentiles rather than the original calcQuantiles() in pod_latency.go, the idea is to shift the implementation of the pod_latency measurement to use `stats` as well in a future PR, but I didn't want to address that here.

pkg/measurements/util/svc_checker.go
```diff
@@ -0,0 +1,74 @@
...
+		restConfig: restConfig,
+	}, nil
+}
+
+func (lc *SvcLatencyChecker) Ping(address string, port int32, timeout time.Duration) error {
+	var stdout, stderr bytes.Buffer
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	// We use 50ms precision thanks to sleep 0.1
+	cmd := []string{"bash", "-c", fmt.Sprintf("while true; do nc -w 0.1s -z %s %d && break; sleep 0.05; done", address, port)}
```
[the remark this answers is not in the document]
rsevilla87: I kept `wait.PollUnitlContextCancel` to prevent crashing in case the connectivity with the API fails or times out.

pkg/burner/namespaces.go
rsevilla87: The problem is that some functions have arguments defined in the burner package, so moving them to utils /namespaces.go will create a import cycle :( For now I just  moved the functions w/o any import cycle, we can address this in a future PR, this one is already big enough :D

pkg/measurements/types/types.go
rsevilla87: There's a complex to fix import cycle if I move it to measurements/ :(

pkg/measurements/metrics/metrics.go
rsevilla87: done

pkg/measurements/pod_latency.go
rsevilla87: ok, I added a new method `validateConfig() error` to the interface so it's now standard across all measurements

pkg/measurements/service_latency.go
rsevilla87: Consolidating into 100 ms

rsevilla87:
This feature is specially interesting for CNI benchmarking. @dcbw @npinaeva

rsevilla87:
Thanks for the review @npinaeva, I've updated the notes with your suggestions. Feel free to test this feature and provide feedback

mohit-sheth (commented):
tested the PR out, looks good!
thanks @rsevilla87

rsevilla87:
> tested the PR out, looks good! thanks @rsevilla87

I need a LGTM to merge this one :)

mohit-sheth (approved):
lgtm

PR #523, opened 2023-11-22
Capability to disable indexing in OCP wrapper

### LINKED ISSUE #520: [RFE] Vanila k8s support to OCP wrappers for development purposes
While it is clear that the OCP wrappers are designed for OpenShift it would be nice to be able to run them on vanila k8s clusters (i.e.: kind) with indexing disabled for development purposes.
For the workloads that reference OCP specific objects (i.e.: routes, imagestreams, etc.) this will not be possible.
It looks like i.e.: node-density does not reference any specific OCP objects so in principle it should be possible to run it once the prometheus requirement is relaxed.

rsevilla87:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

In the current implementation it's not possible to disable indexing, with the logic proposed here when local-indexing and alerting are disabled,  kube-burner doesn't need to discover the cluster's prometheus information, thanks to these changes kube-burner ocp works on vanilla k8s clusters.

In case of needing indexing and alerting, the can be enabled  through the flag `--metrics-endpoint`

- Related Issue #
- Closes #520


pkg/workloads/types.go
```diff
@@ -41,6 +41,7 @@ type Config struct {
 	Indexer         indexers.IndexerType
 	Alerting        bool
 	Reporting       bool
+	Indexing        bool
```
vishnuchalla, line 44: do we want to name this as `skipIndexing`  to keep it consistent with non-ocp kube-burner code?

rsevilla87: Makes sense, thanks!

rsevilla87: I started to do the modifications, but I didn't like the result, since if I we use `SkipIndexing` rather than `Indexing` complicates the logic since I'd have to check negative conditions rather than positives. 
Also keep in mind that this is an internal variable and it's not exposed to the user in any way.

vishnuchalla: Okay if it is internal, looks good to me.

vishnuchalla (approved):

vishnuchalla:
I am not sure on why this is waiting `linters` check and blocking the merge. May be needs a rebase and re-push? 🤔

josecastillolema:
For future reference, to complete disable indexing and alerting and avoid the no prometheus found error:
```
$ kube-burner-ocp node-density --pods-per-node 10 --alerting=false --local-indexing=false
```

PR #533, opened 2023-12-01
Maintain original object's namespace when specified

rsevilla87:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

Keep original object's namespace when specified.

- Related Issue #
- Closes #


venkataanil (approved):
LGTM

pkg/burner/create.go
```diff
@@ -253,6 +253,10 @@ func createRequest(gvr schema.GroupVersionResource, ns string, obj *unstructured
 	var uns *unstructured.Unstructured
 	var err error
 	RetryWithExponentialBackOff(func() (bool, error) {
+		// When the object has a namespace already specified, we use it
+		if objNs := obj.GetNamespace(); objNs != "" {
```
venkataanil, line 257: I think the intention of this is to allow different namespace for object by specifying in object spec file (i.e this namespace is different from what is provided in job spec).
This will be an issue if the user has specified a namespace in the object spec file and that namespace doesn't exist . For example, I have added namespace "anil" in pod.yml and tried running node-density and I got this error

time="2023-12-01 09:05:33" level=error msg="Error creating object Pod/node-density-45 in namespace anil: namespaces \"anil\" not found" file="create.go:277"
time="2023-12-01 09:05:33" level=error msg="Retrying object creation" file="create.go:282"

venkataanil: Looks like many changes went into the patch related to namespace now. Not sure they came in because of this review comment :)

Checking of namespace exist or not will be very expensive for every object creation. My suggestion was to throw an error if the object creation fails (because of not having namespace) and exit instead of trying multiple times. Currently it is trying multiple times. Error message should say that "intention is to be able to create objects in pre-existing namespaces"

We are giving additional capability to create objects with their own namespaces, so we should document that user should create them before running the tests.

rsevilla87: The main goal of this feature is to be able to create objects in pre-existing namespaces.

rsevilla87: As you mentioned, the object retry logic shouldn't happen when the namespace is not present, I'll fix it.

rsevilla87: Fix is in place, snippet from the new behaviour:

```shell
time="2023-12-05 10:47:02" level=error msg="Error creating object Pod/node-density-2: namespaces \"node-density\" not found" file="create.go:278"
time="2023-12-05 10:47:02" level=error msg="Error creating object Pod/node-density-4: namespaces \"node-density\" not found" file="create.go:278"
time="2023-12-05 10:47:02" level=error msg="Error creating object Pod/node-density-1: namespaces \"node-density\" not found" file="create.go:278"
time="2023-12-05 10:47:02" level=error msg="Error creating object Pod/node-density-7: namespaces \"node-density\" not found" file="create.go:278"
time="2023-12-05 10:47:02" level=error msg="Error creating object Pod/node-density-3: namespaces \"node-density\" not found" file="create.go:278"
time="2023-12-05 10:47:02" level=error msg="Error creating object Pod/node-density-6: namespaces \"node-density\" not found" file="create.go:278"
time="2023-12-05 10:47:02" level=error msg="Error creating object Pod/node-density-5: namespaces \"node-density\" not found" file="create.go:278"
time="2023-12-05 10:47:02" level=error msg="Error creating object Pod/node-density-0: namespaces \"node-density\" not found" file="create.go:278"
time="2023-12-05 10:47:02" level=error msg="Error creating object Pod/node-density-8: namespaces \"node-density\" not found" file="create.go:278"
time="2023-12-05 10:47:02" level=info msg="Waiting up to 4h0m0s for actions to be completed" file="create.go:171"
time="2023-12-05 10:47:02" level=info msg="Actions in namespace node-density completed" file="waiters.go:75"
time="2023-12-05 10:47:02" level=info msg="Verifying created objects" file="utils.go:86"
time="2023-12-05 10:47:02" level=error msg="pods found: 0 Expected: 9" file="utils.go:116"
```

pkg/burner/job.go
```diff
@@ -131,8 +130,14 @@ func Run(configSpec config.Spec, prometheusClients []*prometheus.Prometheus, ale
 				if job.Cleanup {
 					ctx, cancel := context.WithTimeout(context.Background(), globalConfig.GCTimeout)
 					defer cancel()
-					CleanupNamespaces(ctx, metav1.ListOptions{LabelSelector: fmt.Sprintf("kube-burner-job=%s", job.Name)}, true)
+					labelSelector := fmt.Sprintf("kube-burner-job=%s", job.Name)
+					CleanupNamespaces(ctx, labelSelector, true)
 					CleanupNonNamespacedResourcesUsingGVR(ctx, jobList, true)
+					for _, obj := range job.objects {
+						if obj.namespace != "" {
```
vishnuchalla, line 137: Does this indicate that in a workload, each obj can have its own namespace either same or different?

rsevilla87: The namespace property is set only when the object has a defined the `namespace` field in its YAML structure, for that reason it checks for that property before pruning the objects from that namespace using `CleanupNamespaceResourcesUsingGVR`

PR #536, opened 2023-12-10
Fix unsafe access to labels map to avoid a panic

### LINKED ISSUE #534: [BUG] Panic when running `init` (sometimes)
# Bug Description

1.8.0

## **Describe the bug**

Sometimes when running `init` `kube-burner` panics (see attached file). It's unclear what triggers it, but it seems that it's triggered more often when there's nothing to clean up from the previous run (so there's no deletion step prior to creating resources).

This did not occur with the `v1.7.13` as far as I can tell.

[gopanic.txt](https://github.com/cloud-bulldozer/kube-burner/files/13598581/gopanic.txt)

## **To Reproduce**

1. Run/rerun `init` several times, try not cleaning up sometimes.

## **Expected behavior**

No panic.

## **Screenshots or output**

See attached filed

diversario:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Make a per-goroutine copy of the `label` map to avoid concurrent read/writes to it.

- Closes #534 

### DIFF
--- a/pkg/burner/create.go
+++ b/pkg/burner/create.go
@@ -204,7 +204,14 @@ func (ex *Executor) generateNamespace(iteration int) string {
 
 func (ex *Executor) replicaHandler(labels map[string]string, obj object, ns string, iteration int, replicaWg *sync.WaitGroup) {
 	var wg sync.WaitGroup
+
 	for r := 1; r <= obj.Replicas; r++ {
+		// make a copy of the copiedLabels map
+		copiedLabels := make(map[string]string)
+		for k, v := range labels {
+			copiedLabels[k] = v
+		}
+
 		wg.Add(1)
 		go func(r int) {
 			defer wg.Done()
@@ -225,11 +232,13 @@ func (ex *Executor) replicaHandler(labels map[string]string, obj object, ns stri
 			}
 			// Re-decode rendered object
 			yamlToUnstructured(renderedObj, newObject)
+
 			for k, v := range newObject.GetLabels() {
-				labels[k] = v
+				copiedLabels[k] = v
 			}
-			newObject.SetLabels(labels)
-			setMetadataLabels(newObject, labels)
+			newObject.SetLabels(copiedLabels)
+			setMetadataLabels(newObject, copiedLabels)
+
 			json.Marshal(newObject.Object)
 			// replicaWg is necessary because we want to wait for all replicas
 			// to be created before running any other action such as verify objects,
--- a/pkg/burner/create.go
+++ b/pkg/burner/create.go
@@ -17,7 +17,6 @@ package burner
 import (
 	"context"
 	"embed"
-	"encoding/json"
 	"fmt"
 	"io"
 	"math"
@@ -239,7 +238,6 @@ func (ex *Executor) replicaHandler(labels map[string]string, obj object, ns stri
 			newObject.SetLabels(copiedLabels)
 			setMetadataLabels(newObject, copiedLabels)
 
-			json.Marshal(newObject.Object)
 			// replicaWg is necessary because we want to wait for all replicas
 			// to be created before running any other action such as verify objects,
 			// wait for ready, etc. Without this wait group, running for example,
--- a/pkg/burner/create.go
+++ b/pkg/burner/create.go
@@ -206,7 +206,7 @@ func (ex *Executor) replicaHandler(labels map[string]string, obj object, ns stri
 	var wg sync.WaitGroup
 
 	for r := 1; r <= obj.Replicas; r++ {
-		// make a copy of the copiedLabels map
+		// make a copy of the labels map for each goroutine to prevent panic from concurrent read and write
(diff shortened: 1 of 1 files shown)


pkg/burner/create.go
```diff
@@ -204,7 +204,14 @@ func (ex *Executor) generateNamespace(iteration int) string {
 
 func (ex *Executor) replicaHandler(labels map[string]string, obj object, ns string, iteration int, replicaWg *sync.WaitGroup) {
 	var wg sync.WaitGroup
+
 	for r := 1; r <= obj.Replicas; r++ {
+		// make a copy of the copiedLabels map
```
Nick-Triller, line 209:
```suggestion
		// make a copy of the labels map for each goroutine to prevent panic from concurrent read and write
```

pkg/burner/create.go
```diff
@@ -204,7 +204,14 @@ func (ex *Executor) generateNamespace(iteration int) string {
 
 func (ex *Executor) replicaHandler(labels map[string]string, obj object, ns string, iteration int, replicaWg *sync.WaitGroup) {
 	var wg sync.WaitGroup
+
 	for r := 1; r <= obj.Replicas; r++ {
+		// make a copy of the labels map for each goroutine to prevent panic from concurrent read and write
+		copiedLabels := make(map[string]string)
+		for k, v := range labels {
+			copiedLabels[k] = v
+		}
```
rsevilla87, line 213: Rather than using a new variable, `copiedLabels`, why don't you pass the variable `labels` as argument to the goroutine below?

pkg/burner/create.go
```diff
@@ -204,7 +204,14 @@ func (ex *Executor) generateNamespace(iteration int) string {
...
+
 		wg.Add(1)
 		go func(r int) {
```
rsevilla87, line 216:
```suggestion
		go func(r int, labels map[string]string) {
```

pkg/burner/create.go
ntriller-plaid: Just to clarify, are you saying you would prefer this?
```golang
# Shadow label variable with copy
labels := make(map[string]string)
for k, v := range labels {
  copiedLabels[k] = v
}
```
Or are you asking why we create a copy? If the latter, the copy fixes an occasional panic in kube burner that occurs because the go routines below are all using the same `labels` map currently. The goroutines iterate over the map as well as write to the map. If one goroutine happens to iterate over the map at the same time another goroutine writes to the map, kube burner panics.

pkg/burner/create.go
rsevilla87: If you pass the labels variable as argument, a new copy of `labels`  is actually passed to the goroutine, similar of what we're already doing with `r`

https://github.com/cloud-bulldozer/kube-burner/blob/c75cbe529b3aed8278292d863d4b8ba1f69f0201/pkg/burner/create.go#L205-L209

rsevilla87:
Hey, any plans to update this PR?, if not I can file a PR myself to fix this bug

diversario:
> Hey, any plans to update this PR?, if not I can file a PR myself to fix this bug

Will update now. BTW, on line 242 https://github.com/cloud-bulldozer/kube-burner/pull/536/files#diff-76e5989cd670c9a6f994b653f5f41d8eacf7d309dc9074b48f9754bbac133018R242 there's `json.Marshal(newObject.Object)` that's unused - should I remove that?

pkg/burner/create.go
ntriller-plaid: Based on this quick test, I do not thing the suggestion works as expected:
```go
package main

import (
	"fmt"
	"time"
)

func main() {
	theMap := make(map[string]int)
	go func(m map[string]int) {
		for {
			// Write to map
			m["pi"] = 3
		}
	}(theMap)
	go func(m map[string]int) {
		// Iterate over map
		for {
			for k, v := range m {
				// do nothing
				if k == "" || v == 0 {
					fmt.Println("should never happen")
				}
			}
		}
	}(theMap)
	time.Sleep(10)
}
```
Above code snippet errors with `fatal error: concurrent map iteration and map write`.

rsevilla87:
> > Hey, any plans to update this PR?, if not I can file a PR myself to fix this bug
> 
> Will update now. BTW, on line 242 https://github.com/cloud-bulldozer/kube-burner/pull/536/files#diff-76e5989cd670c9a6f994b653f5f41d8eacf7d309dc9074b48f9754bbac133018R242 there's `json.Marshal(newObject.Object)` that's unused - should I remove that?

yes please, that's probably a typo

pkg/burner/create.go
rsevilla87: You're right!, just realized that maps are passed by reference and not by value...

rsevilla87: using a mutex can fix the issue, but I think your initial approach is easier to understand, let's keep it :)

PR #537, opened 2023-12-11
Use stats package to calculate quantiles!

rsevilla87:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

Use the stats package helpeers to calculate pod/vm/vmi latency percentiles rather than using our own logic.

- Related Issue #
- Closes #


pkg/measurements/metrics/metrics.go
```diff
@@ -36,23 +37,10 @@ type LatencyQuantiles struct {
 	Avg          int         `json:"avg"`
 	Timestamp    time.Time   `json:"timestamp"`
 	MetricName   string      `json:"metricName"`
-	JobName      string      `json:"jobName"`
```
rsevilla87, line 39: JobName is already defined at jobConfig

pkg/measurements/metrics/metrics.go
```diff
@@ -74,3 +62,21 @@ func CheckThreshold(thresholds []types.LatencyThreshold, quantiles []interface{}
...
 	return utilerrors.NewAggregate(errs)
 }
+
+func NewLatencySummary(input []float64, name string) LatencyQuantiles {
+	latencyQuantiles := LatencyQuantiles{
+		QuantileName: name,
+		Timestamp:    time.Now().UTC(),
+	}
+	val, _ := stats.Percentile(input, 50)
+	latencyQuantiles.P50 = int(val)
```
afcollins, line 72: nit: inline 'val's here?

rsevilla87: I don't follow you, what do you mean?

It's not possible to combine both lines as the function stats.Percentile() returns two values.

afcollins: Ah! Ok. thank you.

PR #540, opened 2023-12-18
config: bump RequestTimeout to 60s by default

dcbw:
In very large clusters or with large scale tests (100,000 plus pods) requests can take more than 15 seconds to list all objects. Bump the timeout to accommodate lots of objects.


rsevilla87:
Thanks for the PR @dcbw , mind you update the docs as well?

dcbw:
@rsevilla87 done!

rsevilla87 (approved):
lgtm

PR #557, opened 2024-01-03
option to create an object only once in multi-iteration jobs

mukrishn:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

In certain cases on creating cluster wide resources like clusterroles, operator spec CRDs, network spec, they would not have to run for every iteration although it is idempotent, this additional option let us attempt just once during the  iteration within a job. 
We have many of such scenarios with ICNI implementaion like serviceaccount, SRIOV specs, ICNI CRDs are required to run only once within a job.

example object, when jobIteration is > 1
```
      - objectTemplate: serviceaccount.yml
        replicas: 1
        runOnce: true

      - objectTemplate: pod.yml
        replicas: 1
```

- Related Issue #
- Closes #

### DIFF
--- a/pkg/burner/create.go
+++ b/pkg/burner/create.go
@@ -152,7 +152,7 @@ func (ex *Executor) RunCreateJob(iterationStart, iterationEnd int, waitListNames
 			if obj.RunOnce {
 				if i == 0 {
 					// this executes only once during the first iteration of an object
-					log.Infof("RunOnce set to %s, so creating object once", obj.ObjectTemplate)
+					log.Debugf("RunOnce set to %s, so creating object once", obj.ObjectTemplate)
 					ex.replicaHandler(labels, obj, ns, i, &wg)
 				}
 			} else {
--- a/pkg/burner/create.go
+++ b/pkg/burner/create.go
@@ -149,7 +149,15 @@ func (ex *Executor) RunCreateJob(iterationStart, iterationEnd int, waitListNames
 				"kube-burner-runid": ex.runid,
 			}
 			ex.objects[objectIndex].labelSelector = labels
-			ex.replicaHandler(labels, obj, ns, i, &wg)
+			if obj.RunOnce {
+				if i == 0 {
+					// this executes only once during the first iteration of an object
+					log.Infof("RunOnce set to %s, so creating object once", obj.ObjectTemplate)
+					ex.replicaHandler(labels, obj, ns, i, &wg)
+				}
+			} else {
+				ex.replicaHandler(labels, obj, ns, i, &wg)
+			}
 		}
 		if !ex.WaitWhenFinished && ex.PodWait {
 			if !ex.NamespacedIterations || !namespacesWaited[ns] {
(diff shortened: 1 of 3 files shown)


venkataanil (approved):
Agree, needed feature.

pkg/burner/create.go
```diff
@@ -149,7 +149,15 @@ func (ex *Executor) RunCreateJob(iterationStart, iterationEnd int, waitListNames
 				"kube-burner-runid": ex.runid,
 			}
 			ex.objects[objectIndex].labelSelector = labels
-			ex.replicaHandler(labels, obj, ns, i, &wg)
+			if obj.RunOnce {
+				if i == 0 {
```
rsevilla87, line 153:
```suggestion
			if obj.RunOnce && i == 0 {
				if i == 0 {
```

pkg/config/types.go
```diff
@@ -92,6 +92,8 @@ type Object struct {
 	Wait bool `yaml:"wait" json:"wait"`
 	// WaitOptions define custom behaviors when waiting for objects creation
 	WaitOptions WaitOptions `yaml:"waitOptions" json:"waitOptions,omitempty"`
+	// Run Once to create the object only once incase of multiple iterative jobs
+	RunOnce bool `yaml:"runonce" json:"runonce,omitempty"`
```
rsevilla87, line 96: Can you add this param to the docs?

pkg/burner/create.go
```diff
@@ -149,7 +149,15 @@ func (ex *Executor) RunCreateJob(iterationStart, iterationEnd int, waitListNames
...
+					// this executes only once during the first iteration of an object
+					log.Infof("RunOnce set to %s, so creating object once", obj.ObjectTemplate)
```
rsevilla87, line 155: I'd remove this log message, is very verbose, it prints the entire template

rsevilla87:
I normally recommend to use a different job with just one iteration to create this kind of objects, but this feature can be useful for other scenarios as well

pkg/burner/create.go
mukrishn: I had this first but it didn't work as expected, it goes in to `if` condition during first iteration and goes to `else` for all other iteration irrespective of `RunOnce`

pkg/burner/create.go
mukrishn: @rsevilla87 I can make it to debug, but this prints just this only once per object, 
```
time="2024-01-03 17:53:46" level=info msg="Creating pods.yml object only once" file="create.go:155"
```
I changed the message later but was using the same variable `obj.ObjectTemplate`

pkg/config/types.go
rsevilla87, line 96: Just one last nit, can you rename this new param to `runOnce` in order to follow camelcase semanthics as the other params?, thanks!

rsevilla87 (approved):
lgtm

PR #570, opened 2024-01-12
Report max CPU usage

venkataanil:
Add max CPU usage of nodes and control plane components in metrics-report.yml


rsevilla87:
> Add max CPU usage of nodes and control plane components in metrics-report.yml

Can you add max CPU usage for the remaining components too?, that way the dashboard won't fail, since it has a repeat for rule for each component

venkataanil:
Even one or two spikes, which can be ignorable, can result in showing high max CPU usage. Should we instead capture histogram_quantile(0.90)?

PR #573, opened 2024-01-15
Switch to Google's uuid module

mikelolasagasti:
`satori/uuid` repo is dead and has different problems as shown in https://github.com/satori/go.uuid/issues/84

Although a fork exists, IMO Google's module is simpler, even if the `NewString` function can be bit confusing as doesn't reference UUID V4 in the name as satori's and other modules do.

## Type of change

- [X] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

- [X] I have performed a self-review of my code.


rsevilla87 (approved):
lgtm, thanks!

Issue #575, opened 2024-01-16
[RFE] Read/GET job type 


riuvshyn:
**Is your feature request related to a problem? Please describe.**
I would want to create load scenario not only for mutation requests but also simulation intensive GET load on API server.

**Describe the solution you'd like**
Add `jobType: read` to job definition.

**Describe alternatives you've considered**
Currently I have to run other tools to generate GET load on API server.

**Additional context**
could look like:
```
  - name: get-crds
    qps: 20
    burst: 20
    jobType: read
    objects:
      - kind: CustomResourceDefinition
        labelSelector: {kube-burner-job: api-intensive}
        apiVersion: apiextensions.k8s.io/v1
```

PR #577, opened 2024-01-16
[WIP] Implement read jobType to simulate load with GET requests

riuvshyn:
## Type of change

Attempt to address https://github.com/kube-burner/kube-burner/issues/575

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [x] Documentation Update

## Description

Adds new `jobType: read` which can be used to simulate load of GET requests 

- Related Issue #
- Closes #

### DIFF
--- a/pkg/burner/read.go
+++ b/pkg/burner/read.go
@@ -1,0 +1,110 @@
+// Copyright 2020 The Kube-burner Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package burner
+
+import (
+	"context"
+	"time"
+
+	"github.com/kube-burner/kube-burner/pkg/config"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func setupReadJob(jobConfig config.Job) Executor {
+	var ex Executor
+	log.Debugf("Preparing read job: %s", jobConfig.Name)
+	mapper := newRESTMapper()
+	for _, o := range jobConfig.Objects {
+		if o.APIVersion == "" {
+			o.APIVersion = "v1"
+		}
+		gvk := schema.FromAPIVersionAndKind(o.APIVersion, o.Kind)
+		mapping, err := mapper.RESTMapping(gvk.GroupKind())
+		if err != nil {
+			log.Fatal(err)
+		}
+		if len(o.LabelSelector) == 0 {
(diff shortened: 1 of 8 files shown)


pkg/burner/read.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+		if len(o.LabelSelector) == 0 {
+			log.Fatalf("Empty labelSelectors not allowed with: %s", o.Kind)
+		}
+		obj := object{
+			gvr:           mapping.Resource,
+			labelSelector: o.LabelSelector,
+		}
+		obj.Namespaced = mapping.Scope.Name() == meta.RESTScopeNameNamespace
+		log.Infof("Job %s: %d iterations", jobConfig.Name, jobConfig.JobIterations)
```
rsevilla87, line 52: I think is better to take this message outside the loop

pkg/burner/read.go
```diff
@@ -0,0 +1,114 @@
...
+		log.Debugf("Job %s: Read %s with selector %s", jobConfig.Name, gvk.Kind, labels.Set(obj.labelSelector))
+		ex.objects = append(ex.objects, obj)
+	}
+	return ex
+}
+
+// RunReadJob executes a reading job
+func (ex *Executor) RunReadJob(iterationStart, iterationEnd int) {
+	var wg sync.WaitGroup
```
rsevilla87, line 61: A workgroup is not required here, the rate limiter already takes care of limiting the number of requests in parallel

rsevilla87 (changes requested):
This feature is quite cool, simple but effective. Now I understand your desire of running jobs in parallel...

However I have some doubts, when the number of objects is high, that will span as many goroutines as the number of objects listed.

I wonder if we can limit the number of concurrent goroutines to QPS to prevent potential issues. (just realized that same issue can happen in delete jobs)

Also, can you a update tests to include this new feature?

pkg/burner/read.go
riuvshyn: absolutely, missed that bit.

pkg/burner/read.go
riuvshyn: oh, very nice!
done.

riuvshyn:
@rsevilla87 I was thinking about problem with high amount of objects spawning many goroutines, limiting amount of concurrent goroutines to QPS probably is not very good idea because based on my use cases I have not that many objects defined in my job while in some jobs I am tempting to set quite high QPS to produce significant load which I think is not unique use case for kube-burner.
I am not very concerned with creating goroutines for every object as I can't imagine sane use-case with hundreds of objects being involved, while setting up QPS to hundreds or even thousands makes alot of sense to me.

pkg/burner/read.go
riuvshyn: done

pkg/burner/read.go
rsevilla87: My bad, we need a waitGroup to wait for goroutines to finish before moving forward to the next iteration. sorry about the confusion, I think the initial code was missing the statement wg.Wait()

rsevilla87:
> @rsevilla87 I was thinking about problem with high amount of objects spawning many goroutines, limiting amount of concurrent goroutines to QPS probably is not very good idea because based on my use cases I have not that many objects defined in my job while in some jobs I am tempting to set quite high QPS to produce significant load which I think is not unique use case for kube-burner. I am not very concerned with creating goroutines for every object as I can't imagine sane use-case with hundreds of objects being involved, while setting up QPS to hundreds or even thousands makes alot of sense to me.

ok, makes sense, let's give it a try, we can always update and improve it in the future. Can you add a simple test case to https://github.com/kube-burner/kube-burner/blob/main/test/test-k8s.bats

Thanks!

riuvshyn:
@riuvshyn I've added simple read test case seems it is passing on CI, I was unable to run it locally because I don't have required local infra.

rsevilla87:
> @rsevilla87 I've added simple read test case seems it is passing on CI, I was unable to run it locally because I don't have required local infra.

Great!, thank you very much!

rsevilla87 (approved):
lgtm!

PR #586, opened 2024-02-02
Remove indexer from workload config

rsevilla87:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

The Indexer field from the workloads.Config  struct, this field is not necessary, we should use the value of the configuration file rather than relying on a field set on this structure.

This PR is part of a bug fix that happens when using a configuration file in the ocp-wrapper with an indexer in the configuration file and a workload is executed w/o passing es-server/es-index/local-indexing

- Related Issue https://github.com/kube-burner/kube-burner-ocp/issues/7


vishnuchalla (approved):
lgtm

PR #593, opened 2024-02-14
Make measurement indexing not a blocking operation

### LINKED ISSUE #589: [RFE] Don't restart measurements on each job
With the current implementation, kube-burner restarts all measurements on a job basis, this behavior doesn't have much sense, ideally we'd like to take these observations during the whole duration of the benchmark.

**Alternatives**

Restarting the measurement's controllers does not consume much time, however indexing the metrics do, rather than prevent restarting the measurements maybe we should index the metrics on a non blocking way (in background) so that the benchmark can continue without this interruption

rsevilla87:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

Indexing measurement metrics is not a blocking operation anymore, subsequent jobs don't have to wait for the metrics to be indexed, some changes were required to prevent race conditions when we restart measurements and the metrics hasn't been indexed yet.

- Related Issue #
- Closes #589

### DIFF
--- a/pkg/burner/job.go
+++ b/pkg/burner/job.go
@@ -85,7 +85,7 @@ func Run(configSpec config.Spec, prometheusClients []*prometheus.Prometheus, ale
 	var rc int
 	var executedJobs []prometheus.Job
 	var jobList []Executor
-	var gcWg sync.WaitGroup
+	var msWg, gcWg sync.WaitGroup
 	embedFS = configSpec.EmbedFS
 	embedFSDir = configSpec.EmbedFSDir
 	errs := []error{}
@@ -97,8 +97,15 @@ func Run(configSpec config.Spec, prometheusClients []*prometheus.Prometheus, ale
 	log.Infof("🔥 Starting kube-burner (%s@%s) with UUID %s", version.Version, version.GitCommit, uuid)
 	go func() {
 		var innerRC int
-		measurements.NewMeasurementFactory(configSpec, indexer, metadata)
+		measurements.NewMeasurementFactory(configSpec, metadata)
 		jobList = newExecutorList(configSpec, uuid, timeout)
+		for _, job := range jobList {
+			if job.PreLoadImages && job.JobType == config.CreationJob {
+				if err = preLoadImages(job); err != nil {
+					log.Fatal(err.Error())
+				}
+			}
+		}
 		// Iterate job list
 		for jobPosition, job := range jobList {
 			var waitListNamespaces []string
@@ -116,11 +123,6 @@ func Run(configSpec config.Spec, prometheusClients []*prometheus.Prometheus, ale
 			}
 			discoveryClient = discovery.NewDiscoveryClientForConfigOrDie(restConfig)
 			DynamicClient = dynamic.NewForConfigOrDie(restConfig)
-			if job.PreLoadImages && job.JobType == config.CreationJob {
-				if err = preLoadImages(job); err != nil {
-					log.Fatal(err.Error())
-				}
-			}
 			currentJob := prometheus.Job{
 				Start:     time.Now().UTC(),
 				JobConfig: job.Job,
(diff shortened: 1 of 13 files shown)


pkg/measurements/types/types.go
```diff
@@ -58,8 +57,6 @@ type Measurement struct {
 	PProfInterval time.Duration `yaml:"pprofInterval"`
 	// PProfDirectory output directory
 	PProfDirectory string `yaml:"pprofDirectory"`
-	// Pod latency metrics to index
-	PodLatencyMetrics latencyMetric `yaml:"podLatencyMetrics"`
```
rsevilla87, line 62: This parameter is useless, it was created for the ocp wrapper and is not using it anymore

pkg/prometheus/prometheus.go
```diff
@@ -53,10 +53,10 @@ func NewPrometheusClient(configSpec config.Spec, url string, auth Auth, step tim
 }
 
 // ScrapeJobsMetrics gets all prometheus metrics required and handles them
-func (p *Prometheus) ScrapeJobsMetrics() error {
+func (p *Prometheus) ScrapeJobsMetrics(jobList ...Job) error {
```
rsevilla87, line 56: Small code optimization

pkg/util/namespaces.go
```diff
@@ -78,7 +78,6 @@ func CleanupNamespaces(ctx context.Context, clientSet *kubernetes.Clientset, lab
 }
 
 func waitForDeleteNamespaces(ctx context.Context, clientSet *kubernetes.Clientset, labelSelector string) {
-	log.Info("Waiting for namespaces to be definitely deleted")
```
rsevilla87, line 81: This log message is redundant with the log message of the caller fuction

pkg/measurements/pod_latency.go
```diff
@@ -150,14 +150,10 @@ func (p *podLatency) validateConfig() error {
 // start podLatency measurement
 func (p *podLatency) start(measurementWg *sync.WaitGroup) error {
 	defer measurementWg.Done()
-	if factory.jobConfig.JobType == config.DeletionJob {
```
rsevilla87, line 153: It's actually compatible, a user could want to delete pods from a deployment/ds and measure how long they take to be ready

pkg/burner/job.go
```diff
@@ -97,8 +97,19 @@ func Run(configSpec config.Spec, prometheusClients []*prometheus.Prometheus, ale
 	log.Infof("🔥 Starting kube-burner (%s@%s) with UUID %s", version.Version, version.GitCommit, uuid)
 	go func() {
 		var innerRC int
-		measurements.NewMeasurementFactory(configSpec, indexer, metadata)
+		mf := measurements.NewMeasurementFactory(configSpec, metadata)
```
vishnuchalla, line 100: why do we want to clone it into another variable? why not use the previous style of doing `measurements.`?

pkg/burner/job.go
```diff
@@ -97,8 +97,19 @@ func Run(configSpec config.Spec, prometheusClients []*prometheus.Prometheus, ale
...
 		jobList = newExecutorList(configSpec, uuid, timeout)
+		ClientSet, restConfig, err = config.GetClientSet(10, 20)
```
vishnuchalla, line 102: These `qps` and `burst` numbers seem a bit random to me. Can't we go ahead with default values or inputted `qps/burst` for such uncertain cases to be consistent across?

pkg/measurements/factory.go
```diff
@@ -43,32 +42,33 @@ type measurement interface {
...
+var Factory MeasurementFactory
 var measurementMap = make(map[string]measurement)
 var globalCfg config.GlobalConfig
 
 // NewMeasurementFactory initializes the measurement facture
-func NewMeasurementFactory(configSpec config.Spec, indexer *indexers.Indexer, metadata map[string]interface{}) {
+func NewMeasurementFactory(configSpec config.Spec, metadata map[string]interface{}) *MeasurementFactory {
 	globalCfg = configSpec.GlobalConfig
-	factory = measurementFactory{
+	Factory = MeasurementFactory{
```
vishnuchalla, line 55: Did we decide on changing the naming convention?

pkg/measurements/vmi_latency.go
```diff
@@ -391,29 +391,23 @@ func (p *vmiLatency) stop() error {
...
 }
 
 // index sends metrics to the configured indexer
-func (p *vmiLatency) index() {
+func (p *vmiLatency) index(indexer indexers.Indexer, jobName string) {
 	metricMap := map[string][]interface{}{
 		podLatencyMeasurement:          p.normLatencies,
 		podLatencyQuantilesMeasurement: p.latencyQuantiles,
 	}
-	if p.config.PodLatencyMetrics == types.Quantiles {
```
vishnuchalla, line 408: Does this mean, we would by default publish all the pod latency metrics along with quantiles and want to get rid of quantiles option from measurement schema?

pkg/measurements/pod_latency.go
vishnuchalla: This one sounds like the story of an update job. How do the pods that are deleted as a part of delete job come alive again?

rsevilla87: As I said in my previous comment, a user can delete the pods from a deployment created by the previous "create" job, then those pods will be recreated by k8s as they belong to a Deployment

pkg/burner/job.go
rsevilla87: I probably left this code by mistake, as I was having some issues with nil pointers, just restored the previous logic including the previous naming convention

pkg/burner/job.go
rsevilla87: Let's use the default ones, we can't use the inputted ones since they are defined within each job

pkg/measurements/vmi_latency.go
rsevilla87: correct

pkg/measurements/pod_latency.go
vishnuchalla: okay understood. 👍🏼

vishnuchalla (approved):
looks good to me.

Issue #592, opened 2024-02-13
[RFE] Churning cycles


rsevilla87:
**Is your feature request related to a problem? Please describe.**
At the moment the namespace churning feature provides the possibility of limiting its duration through `churnDuration`. Adding the possibility of limiting the number of churning cycles not only by time can be useful to get a more deterministic outcome from the benchmarks using it.

**Describe the solution you'd like**
Add a new parameter `churningCycles` with a logic to limit the number of churning  cycles. That's to say, if `churningCycles: 5`, kube-burner will run the churning stage exactly 5 times,

We should offer this feature the possibility to be combined with `churnDuration`, so kube-burner is also able to quit the churning stage when the timeout is reached.

rsevilla87:
fyi @dry923, as developer of churning implementation :)

PR #597, opened 2024-02-17
Adding churnCycles as an option for churning

vishnuchalla:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Adding churnCycles as an option for churning.

- Related Issue: https://github.com/kube-burner/kube-burner/issues/592
- Closes: https://issues.redhat.com/browse/PERFSCALE-2842

## Testing
Tested integrated with https://github.com/kube-burner/kube-burner-ocp/pull/16

### DIFF
--- a/pkg/burner/create.go
+++ b/pkg/burner/create.go
@@ -325,6 +325,7 @@ func (ex *Executor) RunCreateJobWithChurn() {
 	// Determine the number of job iterations to churn (min 1)
 	numToChurn := int(math.Max(float64(ex.ChurnPercent*ex.JobIterations/100), 1))
 	now := time.Now().UTC()
+	cyclesCount := 0
 	rand.NewSource(now.UnixNano())
 	// Create timer for the churn duration
 	timer := time.After(ex.ChurnDuration)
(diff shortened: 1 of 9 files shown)


pkg/burner/create.go
```diff
@@ -338,6 +339,11 @@ func (ex *Executor) RunCreateJobWithChurn() {
 		default:
 			log.Debugf("Next churn loop, workload churning started %v ago", time.Since(now))
 		}
+		// Exit if churn cycles are completed
+		if cyclesCount >= ex.ChurnCycles {
```
afcollins, line 343: So, just to check my understanding... churn duration is not being removed, so duration will act like a timeout if it is set too low? Do we want to have both flags?

rsevilla87: I think we should keep it for backwards compatibility, we can remove/deprecate it the future

vishnuchalla: Yes thinking about backward compatibility, I have kept both. So whichever condition meets first, it will bring churn to the end.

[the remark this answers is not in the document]
vishnuchalla: I have kept its default value to 3. So we need not worry about it not being set. And added a check for >0. Thanks

rsevilla87 (approved):
lgtm

Issue #594, opened 2024-02-14
[RFE] Point out churning start date in indexed documents


rsevilla87:
**Is your feature request related to a problem? Please describe.**

To improve benchmark's result analysis, we need to be able to rule out the collected metrics during churning stage.

**Describe the solution you'd like**

Several changes will be needed:

- As part of the `jobSummary` metric we currently generate

https://github.com/kube-burner/kube-burner/blob/69cbf3c694879f140d1a6acbf38f130a73b6f0fd/pkg/burner/metadata.go#L33-L40

Add a new field to this document to indicate the churning stage start date.

- In the Prometheus/measurement/alerts based indexed results, add an extra (`churning: false/true`?) field to indicate they were collected during the churning stage or not

Once we've these fields we should be able to rule out metrics of churning stage of a benchmark.

**Additional context**

This feature popped up in today's meeting as analyzing metrics for benchmarks with churning can be challenging since their results can be misleading, ie the duration of namespace removal has lot of variation.

cc: @kube-burner/ocp-perfscale @kube-burner/ocp-perfscale-telco

vishnuchalla:
Just to check if my understanding is correct from an implementation perspective, so we want to keep track of start and end times of all the churn cycles and mark all the Prometheus data-points that fall in between those intervals as churn metrics. Does that sound good?
Also along with the start time of churn, do we also want to publish the end time of churn in the `jobSummary`? Because benchmark end time is not same as churn end time as we have some sleep delay [here](https://github.com/kube-burner/kube-burner/blob/main/pkg/burner/create.go#L375).

rsevilla87:
> Just to check if my understanding is correct from an implementation perspective, so we want to keep track of start and end times of all the churn cycles and mark all the Prometheus data-points that fall in between those intervals as churn metrics. Does that sound good? Also along with the start time of churn, do we also want to publish the end time of churn in the `jobSummary`? Because benchmark end time is not same as churn end time as we have some sleep delay [here](https://github.com/kube-burner/kube-burner/blob/main/pkg/burner/create.go#L375).

correct, adding an extra field `churning: true/false` to the indexed documents falling between churning/not-churning intervals should be the correct way.

wrt the second question, `churningDelay` is something that happens during the churning stage, so I think we should include it. 
Churning start/end dates should be taken before/after calling this function:

[...]

PR #599, opened 2024-02-24
Adding churn metrics

vishnuchalla:
## Type of change

- [x] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Adding a flag to filter out churn metrics among the metrics collected from our metrics and alert profiles.

- Related Issue # https://github.com/kube-burner/kube-burner/issues/594
- Closes # https://issues.redhat.com/browse/PERFSCALE-2840

## Testing
Tested in local using the below command and observed its behavior before and after churn.
```
kube-burner init -c kube-burner.yml --uuid="${UUID}" --log-level=debug -m metrics-profile.yaml -u $PROM_URL -t $PROM_TOKEN -a alert-profile.yaml
```
If churn is enabled, we will see a `churnMetric: true` flag in relevant metrics or else we see it. Keeping it this way to keep our docs content optimal.
fields in jobSummary
```
    "churnStartTimestamp": "2024-02-24T23:42:12.259757696Z",
    "churnEndTimestamp": "2024-02-24T23:44:04.688532369Z",
```
fields in metrics
```
  "query": "histogram_quantile(0.99, sum(rate(ovnkube_master_network_programming_duration_seconds_bucket{kind=\"service\"}[2m])) by (le))",
  "churnMetric": true,
  "metricName": "serviceSyncLatency",
```

### DIFF
--- a/pkg/burner/job.go
+++ b/pkg/burner/job.go
@@ -104,9 +104,6 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider, 
 		measurements.NewMeasurementFactory(configSpec, metricsScraper.Metadata)
 		jobList = newExecutorList(configSpec, kubeClientProvider, uuid, timeout)
 		ClientSet, restConfig = kubeClientProvider.DefaultClientSet()
-		if err != nil {
-			log.Fatalf("Error creating clientSet: %s", err)
-		}
 		for _, job := range jobList {
 			if job.PreLoadImages && job.JobType == config.CreationJob {
 				if err = preLoadImages(job); err != nil {
@@ -252,7 +249,7 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider, 
 			}
 			if !job.JobConfig.SkipIndexing {
 				for _, indexer := range metricsScraper.IndexerList {
-					if !job.ChurnStart.IsZero() {
+					if job.JobConfig.Churn {
 						jobTimings.ChurnStartTimestamp = &job.ChurnStart
 						jobTimings.ChurnEndTimestamp = &job.ChurnEnd
 						if churnStart == nil {
--- a/pkg/burner/job.go
+++ b/pkg/burner/job.go
@@ -97,6 +97,8 @@ func Run(configSpec config.Spec, prometheusClients []*prometheus.Prometheus, ale
 	log.Infof("🔥 Starting kube-burner (%s@%s) with UUID %s", version.Version, version.GitCommit, uuid)
 	go func() {
 		var innerRC int
+		var churnStart *time.Time
+		var churnEnd *time.Time
 		measurements.NewMeasurementFactory(configSpec, metadata)
 		jobList = newExecutorList(configSpec, uuid, timeout)
 		ClientSet, restConfig, err = config.GetClientSet(rest.DefaultQPS, rest.DefaultBurst)
@@ -160,7 +162,9 @@ func Run(configSpec config.Spec, prometheusClients []*prometheus.Prometheus, ale
 					log.Error(err.Error())
 				}
 				if job.Churn {
+					currentJob.ChurnStart = time.Now().UTC()
 					job.RunCreateJobWithChurn()
+					currentJob.ChurnEnd = time.Now().UTC()
 				}
 				globalWaitMap[strconv.Itoa(jobPosition)+job.Name] = waitListNamespaces
 				executorMap[strconv.Itoa(jobPosition)+job.Name] = job
@@ -243,11 +247,19 @@ func Run(configSpec config.Spec, prometheusClients []*prometheus.Prometheus, ale
 					EndTimestamp: job.End,
 					ElapsedTime:  job.End.Sub(job.Start).Round(time.Second).Seconds(),
 				}
+				if !job.ChurnStart.IsZero() {
+					jobTimings.ChurnStartTimestamp = &job.ChurnStart
+					jobTimings.ChurnEndTimestamp = &job.ChurnEnd
+					if churnStart == nil {
+						churnStart = &job.ChurnStart
+					}
+					churnEnd = &job.ChurnEnd
+				}
 				indexjobSummaryInfo(indexer, uuid, jobTimings, job.JobConfig, metadata)
 			}
 		}
 		for _, alertM := range alertMs {
-			if err := alertM.Evaluate(executedJobs[0].Start, executedJobs[len(jobList)-1].End); err != nil {
+			if err := alertM.Evaluate(executedJobs[0].Start, executedJobs[len(jobList)-1].End, churnStart, churnEnd); err != nil {
 				errs = append(errs, err)
 				innerRC = 1
 			}
--- a/pkg/burner/job.go
+++ b/pkg/burner/job.go
@@ -27,7 +27,6 @@ import (
 
 	"github.com/cloud-bulldozer/go-commons/indexers"
 	"github.com/cloud-bulldozer/go-commons/version"
-	"github.com/kube-burner/kube-burner/pkg/alerting"
 	"github.com/kube-burner/kube-burner/pkg/config"
 	"github.com/kube-burner/kube-burner/pkg/measurements"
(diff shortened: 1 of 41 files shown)


vishnuchalla:
Adding flag in measurements and alerts. Hope these should suffice, feel free to let me know if we want to add them in more places.

rsevilla87 (commented):
Haven't taken a deep look at the code yet... but as far as I've seen seems ok.
I just realized that there's a loose end ... what do we do with instant metrics??, all of them are going to be tagged with `churnMetric: true` (except the -start ones), shall we consider indexing an extra document per instant query? the timestamp of this document should be churnStartTimestamp  


What do you think @vishnuchalla  @jtaleric

vishnuchalla:
> Haven't taken a deep look at the code yet... but as far as I've seen seems ok.
> 
> I just realized that there's a loose end ... what do we do with instant metrics??, all of them are going to be tagged with `churnMetric: true` (except the -start ones), shall we consider indexing an extra document per instant metric? the timestamp of this document should be churnStartTimestamp
> 
> What do you think @vishnuchalla @jtaleric

I think in case of instant metrics, we shouldn't mark them as churn at all. Because they are executed at two timestamps i.e `jobStart` and `jobEnd` which doesn't fall under churn phase.

rsevilla87:
> > Haven't taken a deep look at the code yet... but as far as I've seen seems ok.
> > I just realized that there's a loose end ... what do we do with instant metrics??, all of them are going to be tagged with `churnMetric: true` (except the -start ones), shall we consider indexing an extra document per instant metric? the timestamp of this document should be churnStartTimestamp
> > What do you think @vishnuchalla @jtaleric
> 
> I think in case of instant metrics, we shouldn't mark them as churn at all. Because they are executed at two timestamps i.e `jobStart` and `jobEnd` which doesn't fall under churn phase.

I don't think so


job End is set with the current date  at
https://github.com/kube-burner/kube-burner/blob/49ae8170f19b0e2ef5104ed655e91acd36a98145/pkg/burner/job.go#L194

And RunCreateJobWithChurn is called before
https://github.com/kube-burner/kube-burner/blob/49ae8170f19b0e2ef5104ed655e91acd36a98145/pkg/burner/job.go#L165-L167

vishnuchalla:
> > > Haven't taken a deep look at the code yet... but as far as I've seen seems ok.
> > > I just realized that there's a loose end ... what do we do with instant metrics??, all of them are going to be tagged with `churnMetric: true` (except the -start ones), shall we consider indexing an extra document per instant metric? the timestamp of this document should be churnStartTimestamp
> > > What do you think @vishnuchalla @jtaleric
> > 
> > 
> > I think in case of instant metrics, we shouldn't mark them as churn at all. Because they are executed at two timestamps i.e `jobStart` and `jobEnd` which doesn't fall under churn phase.
> 
> I don't think so
> 
> job End is set with the current date at
> 
> https://github.com/kube-burner/kube-burner/blob/49ae8170f19b0e2ef5104ed655e91acd36a98145/pkg/burner/job.go#L194
> 
> And RunCreateJobWithChurn is called before

> My bad, yes we are on same page, just phrased it in a complex way. Churn interval is within the job Start/End times but not viceversa is what I meant.

> https://github.com/kube-burner/kube-burner/blob/49ae8170f19b0e2ef5104ed655e91acd36a98145/pkg/burner/job.go#L165-L167
> 
> Anyway I like the idea of not taking into account churning stage for instant metrics at all.

pkg/burner/job.go
```diff
@@ -243,11 +247,19 @@ func Run(configSpec config.Spec, prometheusClients []*prometheus.Prometheus, ale
 					EndTimestamp: job.End,
 					ElapsedTime:  job.End.Sub(job.Start).Round(time.Second).Seconds(),
 				}
+				if !job.ChurnStart.IsZero() {
+					jobTimings.ChurnStartTimestamp = &job.ChurnStart
+					jobTimings.ChurnEndTimestamp = &job.ChurnEnd
+					if churnStart == nil {
```
rsevilla87, line 253: Can you use `churnStart.IsZero()` rather than a pointer?

pkg/prometheus/prometheus.go
```diff
@@ -84,14 +84,14 @@ func (p *Prometheus) ScrapeJobsMetrics(jobList ...Job) error {
 			query := renderedQuery.String()
 			renderedQuery.Reset()
 			if md.Instant {
-				docsToIndex[md.MetricName+"-start"] = append(docsToIndex[md.MetricName+"-start"], p.runInstantQuery(query, md.MetricName+"-start", jobStart, eachJob.JobConfig)...)
-				docsToIndex[md.MetricName] = append(docsToIndex[md.MetricName], p.runInstantQuery(query, md.MetricName, jobEnd, eachJob.JobConfig)...)
+				docsToIndex[md.MetricName+"-start"] = append(docsToIndex[md.MetricName+"-start"], p.runInstantQuery(query, md.MetricName+"-start", jobStart, eachJob)...)
+				docsToIndex[md.MetricName] = append(docsToIndex[md.MetricName], p.runInstantQuery(query, md.MetricName, jobEnd, eachJob)...)
```
rsevilla87, line 88: With this, instant metrics will be scraped using the job ending timestamp as I noted previously. Can we index an extra document in the middle of the job? i.e:

```suggestion
				docsToIndex[md.MetricName+"-start"] = append(docsToIndex[md.MetricName+"-start"], p.runInstantQuery(query, md.MetricName+"-start", jobStart, eachJob)...)
				if !eachJob.ChurnStart.IsZero() {
				    docsToIndex[md.MetricName] = append(docsToIndex[md.MetricName], p.runInstantQuery(query, md.MetricName, jobEnd, eachJob.job.ChurnEnd)...)
				} else {
			    	docsToIndex[md.MetricName] = append(docsToIndex[md.MetricName], p.runInstantQuery(query, md.MetricName, jobEnd, eachJob)...)
			 }
```

pkg/burner/job.go
vishnuchalla: We are using pointer here because we don't want to publish `chrunStart` attribute at all in case of a non-churn job. And `nil` would be an appropriate check for it.  I also verified its working in grafana.

Regarding those churn start and end times, The intention was to get them published as a part of job summary. Did we decide not to do that?

pkg/prometheus/prometheus.go
vishnuchalla: Instant queries are already not being taken into account according to the check [here](https://github.com/kube-burner/kube-burner/blob/49ae8170f19b0e2ef5104ed655e91acd36a98145/pkg/prometheus/prometheus.go#L181). I am not sure about the intention behind this suggestion.

vishnuchalla:
@rsevilla87 I have updated the PR as per our discussion offline. Please take a look. Thanks

pkg/burner/job.go
```diff
@@ -99,9 +99,14 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
...
 	go func() {
 		var innerRC int
+		var churnStart *time.Time
+		var churnEnd *time.Time
 		measurements.NewMeasurementFactory(configSpec, metricsScraper.Metadata)
 		jobList = newExecutorList(configSpec, kubeClientProvider, uuid, timeout)
 		ClientSet, restConfig = kubeClientProvider.DefaultClientSet()
+		if err != nil {
+			log.Fatalf("Error creating clientSet: %s", err)
+		}
```
rsevilla87, line 109: This is not necessary

vishnuchalla: Removed it.

pkg/burner/job.go
```diff
@@ -245,12 +252,20 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
 			}
 			if !job.JobConfig.SkipIndexing {
 				for _, indexer := range metricsScraper.IndexerList {
+					if !job.ChurnStart.IsZero() {
```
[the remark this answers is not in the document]
vishnuchalla: Changed as suggested.

rsevilla87 (approved):
lgtm

PR #604, opened 2024-03-04
Add flags to specify the kubeconfig path and context

### LINKED ISSUE #535: [RFE] Flag to pass kube context to kube burner explicitly
Setting current k8s context so that kube-burner uses the correct cluster is unergonomic and potentially dangerous.

I would like a flag that allows setting the k8s context that should be used by kube burner explicitly. The flag can be optional, with the current behaviour if the flag isn't set.

hhiroshell:
## Type of change

- [x] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [x] Documentation Update

## Description
This PR adds flags to specify the kubeconfig path and context. The target subcommands are as follows:
- init
- health-check
- destroy
- measure

Additionally, this PR includes some small refactorings below:
- Commonize the initialization logic of client config wrote in multiple locations, by adding a new struct KubeClientProvider. This struct provides kubernetes client objects based on single initialization logic.
- Replace `*kubernetes.clientSet` with `kubernetes.Interface` adhere to the concept of utilizing objects through an interface.

- Related Issue #535 
- Closes #535

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -24,18 +24,15 @@ import (
 	"time"
 
 	"github.com/cloud-bulldozer/go-commons/indexers"
+	uid "github.com/google/uuid"
 	mtypes "github.com/kube-burner/kube-burner/pkg/measurements/types"
 	"github.com/kube-burner/kube-burner/pkg/util"
 	log "github.com/sirupsen/logrus"
-
-	uid "github.com/google/uuid"
 	"gopkg.in/yaml.v3"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
-	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
-
 	"k8s.io/client-go/tools/clientcmd"
 )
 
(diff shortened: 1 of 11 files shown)


pkg/config/config.go
```diff
@@ -153,20 +150,61 @@ func Parse(uuid string, f io.Reader) (Spec, error) {
...
-	clientSet := kubernetes.NewForConfigOrDie(restConfig)
+	return &KubeClientProvider{restConfig: restConfig}
+}
+
+func (p *KubeClientProvider) DefaultRestConfig() *rest.Config {
+	restConfig := *p.restConfig
+	return &restConfig
+}
+
+func (p *KubeClientProvider) DefaultClientSet() kubernetes.Interface {
```
rsevilla87, line 189: These two methods are normally executed together, can you join them into a single one returning *rest.Config and kubernetes.Interface?

pkg/config/config.go
```diff
@@ -153,20 +150,61 @@ func Parse(uuid string, f io.Reader) (Spec, error) {
...
+	return kubernetes.NewForConfigOrDie(p.DefaultRestConfig())
+}
+
+func (p *KubeClientProvider) RestConfig(QPS float32, burst int) *rest.Config {
```
rsevilla87, line 193: Same as above

pkg/config/config.go
hhiroshell: Thank you for your review. Fixed in https://github.com/kube-burner/kube-burner/pull/604/commits/bcdd28f6a78c258951ec679658c9b8f919221a96

pkg/config/config.go
hhiroshell: I've done it as well.
https://github.com/kube-burner/kube-burner/pull/604#discussion_r1511165175

pkg/config/config.go
```diff
@@ -153,20 +150,53 @@ func Parse(uuid string, f io.Reader) (Spec, error) {
...
 
-// FetchConfigMap Fetchs the specified configmap and looks for config.yml, metrics.yml and alerts.yml files
-func FetchConfigMap(configMap, namespace string) (string, string, error) {
-	log.Infof("Fetching configmap %s", configMap)
-	var kubeconfig, metricProfile, alertProfile string
-	if os.Getenv("KUBECONFIG") != "" {
-		kubeconfig = os.Getenv("KUBECONFIG")
-	} else if _, err := os.Stat(filepath.Join(os.Getenv("HOME"), ".kube", "config")); kubeconfig == "" && !os.IsNotExist(err) {
-		kubeconfig = filepath.Join(os.Getenv("HOME"), ".kube", "config")
+type KubeClientProvider struct {
```
vishnuchalla, line 153: I think this one should go into types.go of this package.

pkg/config/config.go
```diff
@@ -153,20 +150,53 @@ func Parse(uuid string, f io.Reader) (Spec, error) {
...
+	var restConfig *rest.Config
+	var err error
+	if kubeConfigPath == "" {
+		if restConfig, err = rest.InClusterConfig(); err != nil {
+			log.Fatalf("error preparing kubernetes client: %s", err)
+		}
+	} else {
+		kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+			&clientcmd.ClientConfigLoadingRules{ExplicitPath: kubeConfigPath},
+			&clientcmd.ConfigOverrides{CurrentContext: context},
```
vishnuchalla, line 175: What if the user manages to specify the `kubeconfig` and forgets about `context`. Do want to have a default implementation set for that case? For example
```
return clientcmd.BuildConfigFromFlags("", kubeConfigPath)
```
Because it might be a problem when a kubeconfig doesn't have a default context set.

vishnuchalla:
In overall, I like this idea. Since this is a new feature, can we get some CI tests added/modified for these cases?
1. With kubeconfig and context provided.
2. Another one with just a kubeconfig input.

pkg/config/config.go
hhiroshell: When `--kubeconfig` is specified, returning the default kubeconfig and context may not align with the user's original intent.
Moreover, this could inadvertently result in executing kube-burner to a cluster that was not intended, which I think is not desirable.

pkg/config/config.go
hhiroshell: Thank you for your review. Fixed in https://github.com/kube-burner/kube-burner/pull/604/commits/40f952ed81c273a10be07e9a82580b8fde78bb59 .

hhiroshell:
@vishnuchalla 
I've added the test cases you mentioned; please check them.

@vishnuchalla @rsevilla87 
Additionally, I changed the name of the `--kube-config` flag to `--kubeconfig` to align with kubectl and Helm. Note that Helm uses `--kubeconfig`, even though it has `--kube-context` (separated with a hyphen).

vishnuchalla (approved):
lgtm

hhiroshell:
@rsevilla87
I have resolved the conflicts. If there are no issues, please merge it into the main.

rsevilla87 (approved):
lgtm!, thanks for the PR!

PR #611, opened 2024-03-18
Link profiles to indexers

rsevilla87:
## Type of change

- [x] Refactor
- [x] New feature
- [ ] Bug fix
- [x] Optimization
- [x] Documentation Update

## Description

Multiple changes with the goal to adopt the updated version the `metricsEndpoints` API in the configuration file and link metrics, alerts (supporting multiple files now) and measurements to indexers:

With this change the configuration file now looks like:

```yaml
metricsEndpoints:
- prometheusURL: http://localhost:9090
  token: blablabla
  indexer: 
    type: opensearch
    esServers: ["https://es-instance.com:9200"]
    defaultIndex: kube-burner
  metrics: [metrics-profile.yaml]
  alerts: [alert-profile.yaml]
- prometheusURL: http://localhost:9090
  token: blablabla
  indexer: 
    type: local
  metrics: [metrics-profile.yaml]
  alerts: [alert-profile.yaml]
```

Another important feature is that now It's possible to "split" the measurement metrics and send them to different indexers thanks to the new fields _timeseriesIndexer_ and _quantilesIndexer_. i.e:

```yaml
metricsEndpoints:
- indexer:
    type: local
    alias: local-indexer
- indexer:
    type: opensearch
    defaultIndex: kube-burner
    esServers: ["https://opensearch.domain:9200"]
    alias: opensearch-indexer
global:
  measurements:
  - name: podLatency
    timeseriesIndexer: local-indexer
    quantilesIndexer: opensearch-indexer
```

With the snippet above, the measurement `podLatency` would use the local indexer for timeseries metrics and opensearch for the quantile metrics. cc: @afcollins 

- Related Issue #
- Closes #

### DIFF
--- a/cmd/kube-burner/kube-burner.go
+++ b/cmd/kube-burner/kube-burner.go
@@ -431,7 +431,7 @@ func alertCmd() *cobra.Command {
 				Token:         token,
 				SkipTLSVerify: skipTLSVerify,
 			}
-			p, err := prometheus.NewPrometheusClient(configSpec, url, auth, prometheusStep, nil, false, *indexer)
+			p, err := prometheus.NewPrometheusClient(configSpec, url, auth, prometheusStep, nil, false, indexer)
 			if err != nil {
 				log.Fatal(err)
 			}
--- a/cmd/kube-burner/kube-burner.go
+++ b/cmd/kube-burner/kube-burner.go
@@ -97,7 +97,7 @@ func initCmd() *cobra.Command {
 				log.Fatalf("Config error: %s", err.Error())
 			}
 			metricsScraper := metrics.ProcessMetricsScraperConfig(metrics.ScraperConfig{
-				ConfigSpec:      configSpec,
+				ConfigSpec:      &configSpec,
 				MetricsEndpoint: metricsEndpoint,
 				UserMetaData:    userMetadata,
 			})
@@ -306,7 +306,7 @@ func indexCmd() *cobra.Command {
 			}
 			configSpec.MetricsEndpoints = append(configSpec.MetricsEndpoints, indexer)
 			metricsScraper := metrics.ProcessMetricsScraperConfig(metrics.ScraperConfig{
-				ConfigSpec:      configSpec,
+				ConfigSpec:      &configSpec,
 				MetricsEndpoint: metricsEndpoint,
 				UserMetaData:    userMetadata,
 			})
--- a/cmd/kube-burner/kube-burner.go
+++ b/cmd/kube-burner/kube-burner.go
@@ -187,8 +187,8 @@ func measureCmd() *cobra.Command {
 	var configFile string
 	var jobName string
 	var userMetadata string
-	var indexerList []indexers.Indexer
 	var kubeConfig, kubeContext string
+	indexerList := make(map[string]indexers.Indexer)
 	metadata := make(map[string]interface{})
 	cmd := &cobra.Command{
 		Use:   "measure",
@@ -215,7 +215,7 @@ func measureCmd() *cobra.Command {
 				if err != nil {
 					log.Fatalf("Error creating indexer %d: %v", pos, err.Error())
 				}
-				indexerList = append(indexerList, *idx)
+				indexerList[string(indexer.Alias)] = *idx
 			}
 			if userMetadata != "" {
 				metadata, err = util.ReadUserMetadata(userMetadata)
--- a/cmd/kube-burner/kube-burner.go
+++ b/cmd/kube-burner/kube-burner.go
@@ -122,6 +122,7 @@ func initCmd() *cobra.Command {
 	cmd.Flags().StringVar(&kubeConfig, "kubeconfig", "", "Path to the kubeconfig file")
 	cmd.Flags().StringVar(&kubeContext, "kube-context", "", "The name of the kubeconfig context to use")
 	cmd.Flags().SortFlags = false
+	cmd.MarkFlagRequired("config")
 	return cmd
 }
 
--- a/cmd/kube-burner/kube-burner.go
+++ b/cmd/kube-burner/kube-burner.go
@@ -70,15 +70,11 @@ To configure your bash shell to load completions for each session execute:
 }
 
 func initCmd() *cobra.Command {
-	var err error
 	var clientSet kubernetes.Interface
 	var kubeConfig, kubeContext string
-	var url, metricsEndpoint, metricsProfile, alertProfile, configFile string
-	var metricsProfiles []string
-	var alertsProfiles []string
-	var username, password, uuid, token, configMap, namespace, userMetadata string
+	var metricsEndpoint, configFile string
+	var uuid, userMetadata string
 	var skipTLSVerify bool
-	var prometheusStep time.Duration
 	var timeout time.Duration
 	var rc int
 	cmd := &cobra.Command{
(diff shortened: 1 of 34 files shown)


cmd/kube-burner/kube-burner.go
```diff
@@ -92,20 +88,6 @@ func initCmd() *cobra.Command {
 		Run: func(cmd *cobra.Command, args []string) {
 			kubeClientProvider := config.NewKubeClientProvider(kubeConfig, kubeContext)
 			clientSet, _ = kubeClientProvider.DefaultClientSet()
-			if configMap != "" {
```
rsevilla87, line 95: We're not using this functionality anymore and it's useless for most (if not all) users

cmd/kube-burner/kube-burner.go
```diff
@@ -140,20 +114,10 @@ func initCmd() *cobra.Command {
 		},
 	}
 	cmd.Flags().StringVar(&uuid, "uuid", uid.NewString(), "Benchmark UUID")
-	cmd.Flags().StringVarP(&url, "prometheus-url", "u", "", "Prometheus URL")
-	cmd.Flags().StringVarP(&token, "token", "t", "", "Prometheus Bearer token")
-	cmd.Flags().StringVar(&username, "username", "", "Prometheus username for authentication")
-	cmd.Flags().StringVarP(&password, "password", "p", "", "Prometheus password for basic authentication")
-	cmd.Flags().StringVarP(&metricsProfile, "metrics-profile", "m", "", "Metrics profile file or URL")
```
rsevilla87, line 147: We can get rid of a bunch of flags since they're now specified in the configuration file.

afcollins:
So `timeseriesIndexer: 1` is the index ordinal of the indexers defined under endpoints starting with 1 (not 0) ?

I like being able to specify different metrics profiles for each indexer, so that then we can push report mode to elastic and capture the timeseries locally, for metrics profile metrics.

rsevilla87:
> So `timeseriesIndexer: 1` is the index ordinal of the indexers defined under endpoints starting with 1 (not 0) ?
> 
> I like being able to specify different metrics profiles for each indexer, so that then we can push report mode to elastic and capture the timeseries locally, for metrics profile metrics.

Yeah, that number points to the indexer to be used. The reason not to use 0 was that 0 is the "null" value of an integer variable, so in case of not specifying `timeseriesIndexer` 0 would be assumed. Open to other ideas though; Im not super-fan of this either.

vishnuchalla:
Except for the build and test failures, overall idea looks good to me.

vishnuchalla:
> > So `timeseriesIndexer: 1` is the index ordinal of the indexers defined under endpoints starting with 1 (not 0) ?
> > I like being able to specify different metrics profiles for each indexer, so that then we can push report mode to elastic and capture the timeseries locally, for metrics profile metrics.
> 
> Yeah, that number points to the indexer to be used. The reason not to use 0 was that 0 is the "null" value of an integer variable, so in case of not specifying `timeseriesIndexer` 0 would be assumed. Open to other ideas though; Im not super-fan of this either.

Instead of numbering here, can we have a name filed in each `metricsEndpoint` that can be used to reference as an alias for indexer? that way I feel it will be more intuitive for an end user.

vishnuchalla:
> > > So `timeseriesIndexer: 1` is the index ordinal of the indexers defined under endpoints starting with 1 (not 0) ?
> > > I like being able to specify different metrics profiles for each indexer, so that then we can push report mode to elastic and capture the timeseries locally, for metrics profile metrics.
> > 
> > 
> > Yeah, that number points to the indexer to be used. The reason not to use 0 was that 0 is the "null" value of an integer variable, so in case of not specifying `timeseriesIndexer` 0 would be assumed. Open to other ideas though; Im not super-fan of this either.
> 
> Instead of numbering here, can we have a name filed in each `metricsEndpoint` that can be used to reference as an alias for indexer? that way I feel it will be more intuitive for an end user.

Few more thoughts to add on top of it.
* Currently Prometheus metrics are being indexed to all the indexers. Can we have an option to restrict it to one specific indexer?
* As an extension of above thought I think it is good to have 3 fields for indexing, `timeSeriesIndexer` (for prom metrics), `latenciesIndexer` (for pod latencies etc) and `quantilesIndexer` (for pod latency quantiles) at global (i.e outside measurements config) and at each job level. Meaning indexers at job level will override the ones at global level if a user wants to.
* Default values of global indexers should be all until unless user explicitly specifies index aliases.

rsevilla87:
> > > > So `timeseriesIndexer: 1` is the index ordinal of the indexers defined under endpoints starting with 1 (not 0) ?
> > > > I like being able to specify different metrics profiles for each indexer, so that then we can push report mode to elastic and capture the timeseries locally, for metrics profile metrics.
> > > 
> > > 
> > > Yeah, that number points to the indexer to be used. The reason not to use 0 was that 0 is the "null" value of an integer variable, so in case of not specifying `timeseriesIndexer` 0 would be assumed. Open to other ideas though; Im not super-fan of this either.
> > 
> > 
> > Instead of numbering here, can we have a name filed in each `metricsEndpoint` that can be used to reference as an alias for indexer? that way I feel it will be more intuitive for an end user.
> 
> Few more thoughts to add on top of it.
> 
>     * Currently Prometheus metrics are being indexed to all the indexers. Can we have an option to restrict it to one specific indexer?

They are not, the idea of this PR is to link metricsProfiles to indexers so we can restrict what we send to each indexer, if not working like this there must be a bug

rsevilla87:
> > So `timeseriesIndexer: 1` is the index ordinal of the indexers defined under endpoints starting with 1 (not 0) ?
> > I like being able to specify different metrics profiles for each indexer, so that then we can push report mode to elastic and capture the timeseries locally, for metrics profile metrics.
> 
> Yeah, that number points to the indexer to be used. The reason not to use 0 was that 0 is the "null" value of an integer variable, so in case of not specifying `timeseriesIndexer` 0 would be assumed. Open to other ideas though; Im not super-fan of this either.

Update, I've updated the code to use indexer aliases rather than integers pointing to an indexer

pkg/util/metrics/metrics.go
```diff
@@ -54,56 +47,48 @@ func ProcessMetricsScraperConfig(scraperConfig ScraperConfig) Scraper {
...
-					Profile:     metricsProfile,
-					Username:    scraperConfig.Username,
-					Password:    scraperConfig.Password,
-					EmbedConfig: scraperConfig.EmbedConfig,
-				})
+	// MetricsEndpoint has preference over the configuration file
+	if scraperConfig.MetricsEndpoint != "" {
+		scraperConfig.ConfigSpec.MetricsEndpoints = DecodeMetricsEndpoint(scraperConfig.MetricsEndpoint)
+	}
+	for pos, metricsEndpoint := range scraperConfig.ConfigSpec.MetricsEndpoints {
```
vishnuchalla, line 54: Is this loop indexing `metrics.yaml` in all the given list of indexers?

pkg/measurements/factory.go
```diff
@@ -50,12 +50,37 @@ var globalCfg config.GlobalConfig
...
 func NewMeasurementFactory(configSpec config.Spec, metadata map[string]interface{}) {
+	var indexerFound bool
 	globalCfg = configSpec.GlobalConfig
 	factory = measurementFactory{
 		createFuncs: make(map[string]measurement),
 		metadata:    metadata,
 	}
 	for _, measurement := range globalCfg.Measurements {
+		indexerFound = false
+		if measurement.QuantilesIndexer != "" {
```
vishnuchalla, line 61: Can you refactor this `if-block`? Looks like common logic being used for both the indexers.

pkg/util/metrics/metrics.go
rsevilla87: I don't follow you... what do you mean?

pkg/measurements/factory.go
rsevilla87: updated and optimized this logic, PTAL

vishnuchalla (approved):
lgtm

PR #613, opened 2024-03-21
Updating contributing guidelines and roadmap

vishnuchalla:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [x] Documentation Update

## Description

With recent repo split, it has been a bit of hectic process to make code changes in these repos. To make it simple and well aware updating contributing guidelines and roadmap.


docs/contributing/index.md
```diff
@@ -21,3 +29,57 @@ GOPATH=/home/rsevilla/go/
 GOARCH=amd64 CGO_ENABLED=0 go build -v -ldflags "-X github.com/cloud-bulldozer/go-commons/version.GitCommit=4c9c3f43db83adb053efc58220ddd696d1d19a35 -X github.com/cloud-bulldozer/go-commons/version.BuildDate=2024-01-10-21:24:20 -X github.com/cloud-bulldozer/go-commons/version.Version=main" -o bin/amd64/kube-burner ./cmd/kube-burner
 github.com/kube-burner/kube-burner/cmd/kube-burner
 ```
+
+## Install
+
+To install kube-burner in local, just execute `sudo make install`, once finished the kube-burner binary should be available across the system.
```
krishvoor, line 35: Do we need `sudo`?

docs/contributing/index.md
```diff
@@ -21,3 +29,57 @@ GOPATH=/home/rsevilla/go/
...
+```
+# This will replace all the imports in GO files with your fork path
+find . -type f -name "*.go" -exec sed -i 's/github.com\/kube-burner\/kube-burner/github.com\/<your-fork>\/kube-burner/g' {} +
+```
+3. Finally update the `go.mod` to use your fork. For example
+```
+Replace - module github.com/kube-burner/kube-burner
+With    -> module github.com/<your-fork>/kube-burner
+```
+4. Verify all your changes through `make build; sudo make install`, run some tests locally to verify functionality if required. After that please make a second `git commit` to your changes and push them to your fork.
```
krishvoor, line 54: Do we need `sudo` ?

docs/contributing/index.md
rsevilla87: sudo or root are required for installing the binary in /usr/bin/

krishvoor: > Either sudo or root are required to install the binary in /usr/bin/

Ack, we should consider moving this to `/usr/local/bin` instead of `/usr/bin` 
https://github.com/kube-burner/kube-burner/blob/main/Makefile#L55-L56

vishnuchalla: but why do we want to move it to `/usr/local/bin` in specific?

vishnuchalla: Updated the description.

docs/contributing/index.md
vishnuchalla: Updated the description

docs/contributing/index.md
krishvoor: Package Manager like `dnf install` typically unpacks binaries into `/usr/bin/`.
so I do not want to touch that directory, instead I prefer them to be situated under `/usr/local/bin/` for a system-wide access.

Makefile
```diff
@@ -8,6 +8,7 @@ BIN_DIR = bin
 BIN_PATH = $(BIN_DIR)/$(ARCH)/$(BIN_NAME)
 CGO = 0
 
+GITHUB_USERNAME := $(shell git remote get-url origin | sed -n 's#.*/\([^.]*\)/[^/]*\.git#\1#p' 2>/dev/null)
```
rsevilla87, line 11: This assumes the fork is tagged as origin, in my case in tagged as fork, are we sure about this?

docs/contributing/index.md
vishnuchalla: Sounds good to me. Updated as suggested.

Makefile
vishnuchalla: All it does is, try to extract the username from the remote URL. Every repo will definitely have an origin. Are there any specific edge cases that you are thinking about?

rsevilla87: In my case my own fork is pointed by the remote `fork`, 

```shell
$ git remote -v
fork    git@github.com:rsevilla87/kube-burner.git (fetch)
fork    git@github.com:rsevilla87/kube-burner.git (push)
origin  git@github.com:kube-burner/kube-burner.git (fetch)
origin  git@github.com:kube-burner/kube-burner.git (push)
```

vishnuchalla: Since the name is arbitarary, parameterized it through make target.

vishnuchalla: Made the same change here as well: https://github.com/kube-burner/kube-burner-ocp/pull/37

docs/contributing/index.md
```diff
@@ -21,3 +29,46 @@ GOPATH=/home/rsevilla/go/
...
+    If the above command fails either use `sudo` or try running it as `root`.
+
+## Making changes relevant to OCP wrapper (.i.e. [kube-burner-ocp](https://github.com/kube-burner/kube-burner-ocp))
+
+As of today, we have an OCP wrapper (.i.e. `kube-burner-ocp`) as one of our dependant repo. There are few steps that needs to be followed while making changes in this current repo that effect OCP wrapper's functionality as well.
+
+### Changes at kube-burner/kube-burner repo level
+
+1. Clone this repo and make changes in your fork.
+2. Once done with the changes, please execute `make replace-imports GITHUB_USERNAME=<your-fork-username>`. This will replace all your import and module statement names with your fork's github username.
```
rsevilla87, line 47: I don't understand why do we need to replace imports, when compiling a go module, the compiler will pick the local code rather than reusing the code from the repository.

vishnuchalla: Yes correct, regarding the local build. The problem will arise when using this module as a dependency in `kube-burner-ocp` or any different go package. The module name and imports have to be same while using this in a different GO repo for testing purposes.

rsevilla87: ok, that makes sense, then I think it's not necessary to include these instructions here but in kube-burner-ocp

Also, I personally use a different strategy to use my own kube-burner fork

What I usually do is to run the following command in the kube-burner-ocp directory to tell the golang compiler to fetch dependencies from a local directory:

´´´shell
go mod edit  -replace  github.com/kube-burner/kube-burner=/home/rsevilla/labs/kube-burner
```

Maybe that's easier?, what do you think?

vishnuchalla: Even better. I feel like having some guidelines here to make contributors aware that they also need to raise a PR in `ocp-wrapper` if there are breaking changes would be helpful.  But I think this is a common software engineering dependency problem and guidelines at `ocp-wrapper` should be helpful for people to fix these issues on their own. Closing this PR.

PR #615, opened 2024-04-11
Improve pod waiter

rsevilla87:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [x] Documentation Update

## Description

The `waitForPod` implementation was incorrect and follows a different logic compared to the `waitForDeployments` function.

In the original implementation, this logic waits for all pods in the specified namespace to have status.phase as Running. Which according to the docs means
>- `"Running"` means the pod has been bound to a node and all of the containers have been started. At least one container is still running or is in the process of being restarted.

Meaning that the status phase of the pod can be `Running`, w/o passing any probe.

This behaviour is different in Pods handled by a controller such as Deployment or DaemonSet, in which kube-burner uses the ReadyReplicas/NumberReady status fields, and these fields depend on the Ready condition of the child pods, this condition is marked as true when all containers of the pod are also Ready (incuding probes)

This fix shouldn't affect any the result of any `kube-burner-ocp` workload, the only one potentially affected can be node-density, which uses bare pods w/o any probes.

### DIFF
--- a/docs/measurements.md
+++ b/docs/measurements.md
@@ -31,7 +31,6 @@ One document, such as the following, is indexed per each pod created by the work
   "namespace": "kubelet-density",
   "podName": "kubelet-density-13",
   "nodeName": "worker-001",
-  "jobConfig": {"config": "params"}
 }
 ```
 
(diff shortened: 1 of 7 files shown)


vishnuchalla (approved):
Looks good to me.

PR #632, opened 2024-04-24
Remove unsupported FieldSelector in waitForPVC

ZhangShuaiyi:
## Type of change

- [x] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
`waitForPVC` does not work, kube-burner will not wait pvc to be `Bound` before deleting pvc.
`PersistentVolumeClaim` does not support field-selector `status.phase`
```
[root@master-1 ~]# kubectl get pvc --field-selector status.phase!=Bound
Error from server (BadRequest): Unable to find "/v1, Resource=persistentvolumeclaims" that match label selector "", field selector "status.phase!=Bound": "status.phase" is not a known field selector: only "metadata.name", "metadata.namespace"
``` 
Only `Pod` and `Namespace` support `status.phase` Field.
https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/#list-of-supported-fields

- Related Issue #
- Closes #


vishnuchalla (approved):
looks good to me. Thanks for the thorough analysis

PR #637, opened 2024-04-30
Capturing start is optional in instant metrics

### LINKED ISSUE #603: [RFE] Optional -start instant queries

With the current logic, when kube-burner scrapes an instant query, it generates at least two documents, one uses the start timestamp and the latter the end timestamp of the job. It is not always necessary to capture the `-start` metric, as this metric normally makes sense when applied to metrics that normally get increased over time.

Add an extra field to the metricsProfile, ie: `captureStart: true/false` (false by default), to specify which metrics we want to capture capture from the start of the job.

cc: @afcollins

rsevilla87:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

We normally don't need to capture the start of the metric

- Related Issue #
- Closes #603

### DIFF
--- a/docs/observability/metrics.md
+++ b/docs/observability/metrics.md
@@ -27,7 +27,7 @@ In addition to the default range queries, kube-burner has the ability execute [i
 ```
 
 !!! info
-    When using instant queries, the generated documents are resulting from scraping the last timestamp of each job, it's possible to generate an extra document resulting from scraping the first timestamp of the jobs by adding `captureStart: true` to the metric definition, the resulting document's `metricName` are appended the `-start` suffix.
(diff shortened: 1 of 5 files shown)


docs/observability/metrics.md
```diff
@@ -27,7 +27,7 @@ In addition to the default range queries, kube-burner has the ability execute [i
 ```
 
 !!! info
-    When using instant queries, at least two documents are generated, one resulting from scraping the last timestamp of the job, which would have the configued `metricName` field and an another one resulting from scraping the first timestamp of the job, the `metricName` of document is appended the `-start` suffix.
+    When using instant queries, the generated documents are resulting from scraping the last timestamp of each job, it's possible to generate an extra document resulting from scraping the first timestamp of the jobs by adding `captureStart: true` to the metric definition, the resulting document's `metricName` are appended the `-start` suffix.
```
afcollins, line 30: nit: "... timestamp of each jobs. It is possible ..."

comma makes a run-on sentence.

vishnuchalla (approved):
lgtm

Issue #641, opened 2024-05-08
[RFE] Write output to log file


afcollins:
**Is your feature request related to a problem? Please describe.**
In my experience helping others looking into failed runs, I ask for the output of the kube-burner-ocp run. Others rarely save this output and lose it in a terminal screen, which can also be limited by the screen buffer size, at best.

**Describe the solution you'd like**
I would like kube-burner-ocp to log to console out, but also write to a local file. Preferably with some unique identifier in the file name so they don't clobber each other.

**Describe alternatives you've considered**
When I do runs locally, I use `tee -a` to write to a local file. It usually looks like this:
```
kube-burner init ... 2>&1 | tee -a kube-burner.$(date +"%F-%H%M%S").log
```

**Additional context**
Would prefer this to be enabled by default, so newer users do not need to explicitly set anything and re-run in order to ask for help and they can simply share the output from the failure. In my mind, this reduces a full round trip between colleagues.

PR #655, opened 2024-06-26
Log output

shashank-boyapally:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Adds log output as a file

- Related Issue #641 
- Closes #

### DIFF
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -15,6 +15,7 @@
 package util
 
 import (
+	"fmt"
 	"io"
 	"math"
 	"os"
@@ -37,7 +38,7 @@ func RetryWithExponentialBackOff(fn wait.ConditionFunc, duration time.Duration, 
 }
 
 func SetupLogging(uuid string) {
-	logFileName := "logfile_" + uuid + ".log"
+	logFileName := fmt.Sprintf("logfile_%s.log", uuid)
 	file, err := os.OpenFile(logFileName, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
 	if err != nil {
 		log.Fatalf("Failed to open log file: %v", err)
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -38,8 +38,8 @@ func RetryWithExponentialBackOff(fn wait.ConditionFunc, duration time.Duration, 
 }
 
 func SetupLogging(uuid string) {
-	logFileName := fmt.Sprintf("logfile_%s.log", uuid)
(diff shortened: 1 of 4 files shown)


pkg/util/utils.go
```diff
@@ -32,3 +36,13 @@ func RetryWithExponentialBackOff(fn wait.ConditionFunc, duration time.Duration,
 	}
 	return wait.ExponentialBackoff(backoff, fn)
 }
+
+func SetupLogging(uuid string) {
+	logFileName := fmt.Sprintf("logfile_%s.log", uuid)
```
rsevilla87, line 41: Let's call prefix it with kube-burner better? sounds more intuitive

```suggestion
	logFileName := fmt.Sprintf("kube-burner-%s.log", uuid)
```

pkg/util/utils.go
```diff
@@ -32,3 +36,13 @@ func RetryWithExponentialBackOff(fn wait.ConditionFunc, duration time.Duration,
...
+	file, err := os.OpenFile(logFileName, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
```
rsevilla87, line 42: I think is better to use os.Create (it doesn't require to specify flags or filemode), although it opens the file without the O_APPEND flag, do we need this flag @afcollins ?, uuids are supposed to be unique.

pkg/util/utils.go
shashank-boyapally: Yep suits more to our tool.

pkg/util/utils.go
shashank-boyapally: Moving this to os.Create, primarily when uuid was not included I wanted to append it to the file, did not change it. os.Create makes more sense now. let me know if we want to revert back.

rsevilla87:
Can you add some kind of check in tests to verify the log file is being successfully created and populated?

pkg/util/utils.go
```diff
@@ -32,3 +36,13 @@ func RetryWithExponentialBackOff(fn wait.ConditionFunc, duration time.Duration,
 	}
 	return wait.ExponentialBackoff(backoff, fn)
 }
+
+func SetupLogging(uuid string) {
+	logFileName := fmt.Sprintf("kube-burner-%s.log", uuid)
+	file, err := os.Create(logFileName)
+	if err != nil {
+		log.Fatalf("Failed to open log file: %v", err)
```
vishnuchalla, line 44: s/open/create/

rsevilla87 (approved):
lgtm!

PR #663, opened 2024-07-07
Handling empty metadata cases to be handled in metrics.

vishnuchalla:
## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

Removing metadata when it is empty to have more cleaner metrics.

## Testing
Tested in local and verified
```
  {
    "timestamp": "2024-07-07T19:27:14.434Z",
    "labels": {
      "container": "lightspeed-service-api",
      "namespace": "openshift-lightspeed",
      "node": "ip-10-0-78-217.us-west-2.compute.internal",
      "pod": "lightspeed-app-server-6c695f7995-ztp4q"
    },
    "value": 0.11032656977364322,
    "uuid": "158de0a0-7fd0-4879-8798-ae1ddcb8b5c2",
    "query": "(sum(irate(container_cpu_usage_seconds_total{name!=\"\",container!~\"POD|\",namespace=~\"openshift-lightspeed\"}[2m]) * 100) by (container, pod, namespace, node)) > 0",
    "metricName": "containerCPU"
  }
```

### DIFF
--- a/pkg/prometheus/prometheus.go
+++ b/pkg/prometheus/prometheus.go
@@ -175,7 +175,9 @@ func (p *Prometheus) createMetric(query, metricName string, job Job, labels mode
 		Query:      query,
 		MetricName: metricName,
 		Timestamp:  timestamp,
-		Metadata:   p.metadata,
+	}
+	if len(p.metadata) != 0 {
+		m.Metadata = p.metadata 
 	}
 	for k, v := range labels {
 		if k != "__name__" {
--- a/pkg/util/metrics/metrics.go
+++ b/pkg/util/metrics/metrics.go
@@ -26,8 +26,11 @@ import (
 
 // Processes common config and executes according to the caller
 func ProcessMetricsScraperConfig(scraperConfig ScraperConfig) Scraper {
+	metadata := make(map[string]interface{})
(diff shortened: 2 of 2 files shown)


pkg/util/utils.go
```diff
@@ -32,3 +33,16 @@ func RetryWithExponentialBackOff(fn wait.ConditionFunc, duration time.Duration,
 	}
 	return wait.ExponentialBackoff(backoff, fn)
 }
+
+// Helper function to check if the interface{} is an empty map
+func IsEmptyMap(v interface{}) bool {
```
rsevilla87, line 38: You can check if the map is empty with just `len(p.metadata) == 0`, have you tried it?

vishnuchalla: Correct. But my idea behind this function is to have a generic implementation that works for any type of map also if a map is nil that shouldn't be considered as empty. Because most of types are written as 
xyz type             `json:"xyz,omitempty"` instead of
xyz *type           `json:"xyz,omitempty"`
where in pointer type handles both nil and empty value case to be omitted in the json output.

rsevilla87: I didn't say to compare with nil, but to compare the number of keys with 0, which means is an empty map

rsevilla87 (approved):
lgtm

PR #664, opened 2024-07-11
Making version optional

vishnuchalla:
## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Making version optional for `jobSummary` so that it will not be present in tools that don't require it.


rsevilla87 (approved):
lgtm

PR #665, opened 2024-07-16
Removing metadata from documents except jobSummary

rsevilla87:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

With this PR, cluster's metadata is no longer added to all documents except `jobSummary`. This helps to reduce the size of the generated documents.

- Related Issue #
- Closes #

### DIFF
--- a/pkg/burner/metadata.go
+++ b/pkg/burner/metadata.go
@@ -15,6 +15,7 @@
 package burner
 
 import (
+	"encoding/json"
 	"time"
 
 	"github.com/cloud-bulldozer/go-commons/indexers"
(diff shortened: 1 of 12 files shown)


pkg/burner/metadata.go
```diff
@@ -31,24 +32,31 @@ type JobSummary struct {
 	UUID                string                 `json:"uuid"`
 	MetricName          string                 `json:"metricName"`
 	JobConfig           config.Job             `json:"jobConfig"`
-	Metadata            map[string]interface{} `json:"metadata,omitempty"`
 	Version             string                 `json:"version,omitempty"`
 	Passed              bool                   `json:"passed"`
 	ExecutionErrors     string                 `json:"executionErrors,omitempty"`
+	Metadata            map[string]interface{} `json:"-"`
```
vishnuchalla, line 38: Now that we have flattened the metadata here, can we use the same flattened version in grafana dashboards as well please? I still see some traces like below while querying `jobSummary` in the dashboard JSON model.
```
              "metadata.clusterType": true,
              "metadata.infraNodesCount": false,
```

vishnuchalla (changes requested):
I think this PR can safely go in once conflicts are resolved.

Issue #650, opened 2024-06-14
[BUG] fatal error: concurrent map iteration and map write


jicki:
# Bug Description

## **Output of `kube-burner` version**

```
Version: 1.9.8
Git Commit: 204e1eea0402b3cb0f42e1cc2de89b3d1bae38dd
Build Date: 2024-06-03T16:15:16Z
Go Version: go1.21.10
OS/Arch: linux amd64
```

## **Describe the bug**

```
time="2024-06-14 03:25:29" level=info msg="Stopping measurement: podLatency" file="factory.go:125"
fatal error: concurrent map iteration and map write

goroutine 37 [running]:
github.com/kube-burner/kube-burner/pkg/measurements.(*podLatency).normalizeMetrics(0xc0001aa8c0)
        /home/runner/work/kube-burner/kube-burner/pkg/measurements/pod_latency.go:262 +0x85
github.com/kube-burner/kube-burner/pkg/measurements.(*podLatency).stop(0xc0001aa8c0)
        /home/runner/work/kube-burner/kube-burner/pkg/measurements/pod_latency.go:231 +0x74
github.com/kube-burner/kube-burner/pkg/measurements.Stop()
        /home/runner/work/kube-burner/kube-burner/pkg/measurements/factory.go:126 +0x148
github.com/kube-burner/kube-burner/pkg/burner.Run.func1()
        /home/runner/work/kube-burner/kube-burner/pkg/burner/job.go:200 +0x1833
created by github.com/kube-burner/kube-burner/pkg/burner.Run in goroutine 1
        /home/runner/work/kube-burner/kube-burner/pkg/burner/job.go:100 +0x5e5

goroutine 1 [select, 4 minutes]:
github.com/kube-burner/kube-burner/pkg/burner.Run({{0xc0004d8100, 0x1, 0x1}, {{0xc0000440f0, 0x24}, {0xc00044a390, 0x24}, {0xc00009bc00, 0x1, 0x1}, ...}, ...}, ...)
        /home/runner/work/kube-burner/kube-burner/pkg/burner/job.go:279 +0x64d
main.initCmd.func2(0xc0001aed00?, {0xc000471b30?, 0x4?, 0x210e339?})
        /home/runner/work/kube-burner/kube-burner/cmd/kube-burner/kube-burner.go:110 +0x4ed
github.com/spf13/cobra.(*Command).execute(0xc000004600, {0xc000471ad0, 0x3, 0x3})
[...]

PR #667, opened 2024-07-17
Using sync.Map for thread safety to calculate measurements

vishnuchalla:
## Type of change

- [x] Refactor
- [ ] New feature
- [x] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description
Using sycn.Map in all our measurement calculations so that we don't have to worry about thread safety as it is offered in-built. With this change we get rid of all the mutex locks that we have. Also this will help us stay away from concurrent read/write issues in the metrics maps.

- Related Issue # https://github.com/kube-burner/kube-burner/issues/650

## Testing
Tested in local integrating with ocp wrapper.
```
time="2024-07-17 15:48:04" level=info msg="Stopping measurement: podLatency" file="factory.go:125"
time="2024-07-17 15:48:04" level=info msg="Evaluating latency thresholds" file="metrics.go:46"
time="2024-07-17 15:48:04" level=info msg="node-density-heavy: PodScheduled 50th: 0 99th: 1000 max: 1000 avg: 18" file="pod_latency.go:236"
time="2024-07-17 15:48:04" level=info msg="node-density-heavy: ContainersReady 50th: 6000 99th: 41000 max: 42000 avg: 21462" file="pod_latency.go:236"
time="2024-07-17 15:48:04" level=info msg="node-density-heavy: Initialized 50th: 0 99th: 1000 max: 1000 avg: 75" file="pod_latency.go:236"
time="2024-07-17 15:48:04" level=info msg="node-density-heavy: Ready 50th: 6000 99th: 41000 max: 42000 avg: 21462" file="pod_latency.go:236"
time="2024-07-17 15:48:04" level=info msg="Garbage collection metrics on, waiting for GC" file="job.go:225"
time="2024-07-17 15:48:04" level=info msg="Indexing collected data from measurement: podLatency" file="factory.go:137"
time="2024-07-17 15:48:04" level=info msg="Indexing metric podLatencyMeasurement" file="common.go:27"
time="2024-07-17 15:48:04" level=info msg="File collected-metrics-abcd/podLatencyMeasurement-node-density-heavy.json created with 160 documents" file="common.go:36"
time="2024-07-17 15:48:04" level=info msg="Indexing metric podLatencyQuantilesMeasurement" file="common.go:27"
time="2024-07-17 15:48:04" level=info msg="File collected-metrics-abcd/podLatencyQuantilesMeasurement-node-density-heavy.json created with 4 documents" file="common.go:36"
```

Collected metrics also seem to be fine
```
    "quantileName": "Ready",
    "uuid": "abcd",
    "P99": 41000,
    "P95": 41000,
    "P50": 6000,
    "max": 42000,
    "avg": 21462,
    "timestamp": "2024-07-17T19:48:04.444971388Z",
    "metricName": "podLatencyQuantilesMeasurement",
```

### DIFF
--- a/pkg/measurements/pod_latency.go
+++ b/pkg/measurements/pod_latency.go
@@ -58,64 +58,60 @@ type podMetric struct {
 type podLatency struct {
 	config           types.Measurement
 	watcher          *metrics.Watcher
-	metrics          map[string]podMetric
-	metricLock       sync.RWMutex
+	metrics          sync.Map
 	latencyQuantiles []interface{}
 	normLatencies    []interface{}
 }
 
 func init() {
 	measurementMap["podLatency"] = &podLatency{
-		metrics: make(map[string]podMetric),
+		metrics: sync.Map{},
 	}
 }
 
 func (p *podLatency) handleCreatePod(obj interface{}) {
 	pod := obj.(*corev1.Pod)
-	p.metricLock.Lock()
-	defer p.metricLock.Unlock()
-	if _, exists := p.metrics[string(pod.UID)]; !exists {
-		p.metrics[string(pod.UID)] = podMetric{
-			Timestamp:  pod.CreationTimestamp.Time.UTC(),
-			Namespace:  pod.Namespace,
-			Name:       pod.Name,
-			MetricName: podLatencyMeasurement,
-			UUID:       globalCfg.UUID,
-			Metadata:   factory.metadata,
-		}
-	}
+	p.metrics.LoadOrStore(string(pod.UID), podMetric{
+		Timestamp:  pod.CreationTimestamp.Time.UTC(),
+		Namespace:  pod.Namespace,
+		Name:       pod.Name,
+		MetricName: podLatencyMeasurement,
+		UUID:       globalCfg.UUID,
+		Metadata:   factory.metadata,
+	})
 }
 
 func (p *podLatency) handleUpdatePod(obj interface{}) {
 	pod := obj.(*corev1.Pod)
-	p.metricLock.Lock()
-	defer p.metricLock.Unlock()
-	if pm, exists := p.metrics[string(pod.UID)]; exists && pm.podReady.IsZero() {
-		for _, c := range pod.Status.Conditions {
-			if c.Status == corev1.ConditionTrue {
-				switch c.Type {
-				case corev1.PodScheduled:
-					if pm.scheduled.IsZero() {
-						pm.scheduled = c.LastTransitionTime.Time.UTC()
-						pm.NodeName = pod.Spec.NodeName
-					}
-				case corev1.PodInitialized:
-					if pm.initialized.IsZero() {
-						pm.initialized = c.LastTransitionTime.Time.UTC()
-					}
-				case corev1.ContainersReady:
-					if pm.containersReady.IsZero() {
-						pm.containersReady = c.LastTransitionTime.Time.UTC()
-					}
-				case corev1.PodReady:
-					if pm.podReady.IsZero() {
-						log.Debugf("Pod %s is ready", pod.Name)
-						pm.podReady = c.LastTransitionTime.Time.UTC()
+	if value, exists := p.metrics.Load(string(pod.UID)); exists {
+		pm := value.(podMetric)
+		if pm.podReady.IsZero() {
+			for _, c := range pod.Status.Conditions {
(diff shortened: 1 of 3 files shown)


pkg/measurements/service_latency.go
```diff
@@ -248,13 +246,17 @@ func (s *serviceLatency) stop() error {
 func (s *serviceLatency) normalizeMetrics() {
 	var latencies []float64
 	var ipAssignedLatencies []float64
-	for _, metric := range s.metrics {
+	sLen := 0
+	s.metrics.Range(func(key, value interface{}) bool {
+		sLen++
```
afcollins, line 251: Is this thread safe?

In Java, I'd use an AtomicInteger in Java for a thread-safe increment. But also: What's wrong with running `len(...)` on 268?

rsevilla87 (commented):
Not sure about the actual benefits of this PR, it makes logic more complex and doesn't solve any issue

pkg/measurements/service_latency.go
vishnuchalla: Yes you are right about atomic integers. But we do not need one here. Because `sLen` is being incremented inside the `s.metrics.Range` loop which is already thread safe. Even though an another thread tries to access `sLen` which is specific to this function, it has to wait until `.Range` releases it again for use.

And regarding `len(...)`, sync.Map type doesn't have `len(...)` compatibility inbuilt, hence we have to write our own counter.

vishnuchalla:
> Not sure about the actual benefits of this PR, it makes logic more complex and doesn't solve any issue. Under the hood, sync.Map uses mutex as well

I agree with its usage of `mutex` under the hood. But the main purpose of this change is to not having us to worry about placing mutex locks everywhere needed. If we miss out on something we it might lead in to errors like these
```
goroutine 37 [running]:
github.com/kube-burner/kube-burner/pkg/measurements.(*podLatency).normalizeMetrics(0xc0001aa8c0)
        /home/runner/work/kube-burner/kube-burner/pkg/measurements/pod_latency.go:262 +0x85
github.com/kube-burner/kube-burner/pkg/measurements.(*podLatency).stop(0xc0001aa8c0)
```
Which is because of no lock being present [here](https://github.com/kube-burner/kube-burner/blob/main/pkg/measurements/pod_latency.go#L262). But if we have thread safety handled inbuilt, then we don't have to worry about such cases, which is recommended for robustness.

pkg/measurements/service_latency.go
```diff
@@ -170,15 +168,15 @@ func (s *serviceLatency) handleCreateSvc(obj interface{}) {
 			UUID:              globalCfg.UUID,
 			Metadata:          factory.metadata,
 			IPAssignedLatency: ipAssignedLatency,
-			JobName:           factory.jobConfig.Name,
-		}
-		s.metricLock.Unlock()
+      JobName:           factory.jobConfig.Name,
```
rsevilla87, line 171: lint?

pkg/measurements/vmi_latency.go
```diff
@@ -98,37 +97,36 @@ func init() {
...
+			p.metrics.Store(vmID, &vmiMetric{
 				Timestamp:  time.Now().UTC(),
 				Namespace:  vm.Namespace,
 				Name:       vm.Name,
 				MetricName: vmiLatencyMeasurement,
 				UUID:       globalCfg.UUID,
 				Metadata:   factory.metadata,
-				JobName:    factory.jobConfig.Name,
-			}
+        JobName:    factory.jobConfig.Name,
```
rsevilla87, line 109: linting

pkg/measurements/pod_latency.go
```diff
@@ -59,65 +59,61 @@ type podMetric struct {
...
-		}
-	}
+	p.metrics.LoadOrStore(string(pod.UID), podMetric{
+		Timestamp:  pod.CreationTimestamp.Time.UTC(),
+		Namespace:  pod.Namespace,
+		Name:       pod.Name,
+		MetricName: podLatencyMeasurement,
+		UUID:       globalCfg.UUID,
+		Metadata:   factory.metadata,
+    JobName:    factory.jobConfig.Name,
```
rsevilla87, line 82: linting :)

rsevilla87 (approved):
lgtm, this pr helps on simplifying the structs and making concurrency more golang native

PR #679, opened 2024-08-12
[WIP] network policy latency at scale

venkataanil:
New approach to network policy latency testing.

## Type of change

- [ ] Refactor

## Description

This workload uses 3 jobs.
1) 1st Job, network-policy-proxy, creates a pod and a route. kube-burner uses this pod to communicate with workload pods.
 2) 2nd job, network-policy-pods, creates namespaces and pods on which we will be applying network policies
3) 3rd job, network-policy, creates network polcies between pre existing pods created in previous jobs

Network policy allows connection between different namespace pods on the specified ports. So the testing involves communication between these pods.


pkg/measurements/metrics/watcher.go
```diff
@@ -46,6 +48,15 @@ func NewWatcher(restClient *rest.RESTClient, name string, resource string, names
 	}
 }
 
+func NewNetworkWatcher(restClient kubernetes.Interface, name string, namespace string, optionsModifier func(options *metav1.ListOptions), indexers cache.Indexers) *Watcher {
```
rsevilla87, line 51: This function is not actually necessay, you can get a watcher with a Network policy informer using the generic NewWatcher function

pkg/measurements/factory.go
```diff
@@ -47,10 +48,14 @@ type measurement interface {
 var factory measurementFactory
 var measurementMap = make(map[string]measurement)
 var globalCfg config.GlobalConfig
+var embedFS embed.FS
+var embedFSDir string
 
 // NewMeasurementFactory initializes the measurement facture
 func NewMeasurementFactory(configSpec config.Spec, metadata map[string]interface{}) {
 	var indexerFound bool
+	embedFS = configSpec.EmbedFS
```
rsevilla87, line 57: Why are you instantiating embedFSand embedFSDir?

pkg/measurements/metrics/watcher.go
rsevilla87: i.e:
```golang
metrics.NewWatcher(
		factory.clientSet.CoreV1().RESTClient().(*rest.RESTClient),
		"netpolWatcher",
		"networkpolicies",
		corev1.NamespaceAll,
		func(options *metav1.ListOptions) {
			options.LabelSelector = fmt.Sprintf("kube-burner-uuid=%v", globalCfg.UUID)
		},
		cache.Indexers{},
	)
```

pkg/measurements/netpol_latency.go
```diff
@@ -0,0 +1,613 @@
...
+	return uns.GetKind()
+}
+
+// Start network policy measurement
+func (n *netpolLatency) start(measurementWg *sync.WaitGroup) error {
+	defer measurementWg.Done()
+	// Use proxyRoute to communicate with proxy pod
+	proxyRoute = n.config.Config["proxyRoute"]
+	// Skip latency measurement for 1st job which creates only pods
+	if value, ok := factory.jobConfig.NamespaceLabels["kube-burner.io/skip-networkpolicy-latency"]; ok {
```
rsevilla87, line 500: Using annotations is better for this kind of actions

pkg/measurements/netpol_latency.go
```diff
@@ -0,0 +1,613 @@
...
+	yamlToUnstructured(o.ObjectTemplate, cleanTemplate, uns)
+	return uns.GetKind()
+}
+
+// Start network policy measurement
+func (n *netpolLatency) start(measurementWg *sync.WaitGroup) error {
+	defer measurementWg.Done()
+	// Use proxyRoute to communicate with proxy pod
+	proxyRoute = n.config.Config["proxyRoute"]
+	// Skip latency measurement for 1st job which creates only pods
```
rsevilla87, line 499: Let's not assume that there's going to be a first job

```suggestion
	// Skip latency measurement in namespaces with this label
```

pkg/measurements/netpol_latency.go
```diff
@@ -0,0 +1,613 @@
...
+	if value, ok := factory.jobConfig.NamespaceLabels["kube-burner.io/skip-networkpolicy-latency"]; ok {
+		if value == "true" {
+			log.Debugf("Discarding network policy latency measurement for the job %v", factory.jobConfig.Name)
+			return nil
+		}
+	}
```
rsevilla87, line 505: Are we sure we want to do this?, this check can be done in runtime, i.e:

check the label/annotation when the namespace is created

pkg/measurements/netpol_latency.go
```diff
@@ -0,0 +1,613 @@
...
+
+// It parses the network policy object for each iteration and identify the remote pods from which the namespace has to allow ingress traffic.
+// To test ingress, remote pods will be the client pods which send requests to current namespace local pods on which ingress policy is defined.
+// For example, local pods in namespace-1 will be accepting http requests from remote namespace-2 pods when ingress-1 in namespace-1 specify namespace-2 as remote.
+// connections is map of each client pod with list of addresses it has to send connection requests. Network policy name is also stored along with the addresses.
+// Note: this measurement package only parses template to get IP addresses of pods but never creates network policies. kube-burner's "burner" package creates network policies
+func prepareConnections() {
+	// Reset latency slices, required in multi-job benchmarks
+	for _, obj := range factory.jobConfig.Objects {
+		cleanTemplate, err := readTemplate(obj)
```
rsevilla87, line 249: rather than reading and render all templates again, why don't you pass the job (Executor object) to the factory object?

rsevilla87: example:
The current implementation passes a config.Job instance

https://github.com/kube-burner/kube-burner/blob/0bf128f8f34d8d3d499beeac7c25111ee5d15d48/pkg/burner/job.go#L141-L143


If we pass the job executor object we can take advantage of its objects array from here:

```golang
measurements.SetJobConfig(&job, kubeClientProvider)
```

This approach will require some small changes in other files, but I think the change is worthy to avoid duplicated actions.

venkataanil: That's a great suggestion. I will do it

pkg/measurements/netpol_latency.go
venkataanil: As you know, we have 2 jobs here both using the **same namespace**. First job creates pods and we skip measurments in this first job. Second job creates network policies and we want to trigger this latency measurment. So we can't use namespace annotation which impacts second job as well.

Also we use the same skip check in stop() function.

pkg/measurements/factory.go
venkataanil: I might be using this for reading template file from the directory. I will check and remove if it's not needed.

pkg/measurements/netpol_latency.go
venkataanil: I tried your suggestion, however this is not possible now as I need to define varaibles of "type Executor struct" inside "type measurementFactory struct {}". However "type Executor struct" belongs to "burner" package and I can't import burner package in factory.go

This is not possible for now
[root@vkommadi kube-burner]# git diff pkg/measurements/factory.go
diff --git a/pkg/measurements/factory.go b/pkg/measurements/factory.go
index 9ceae61..814908b 100644
--- a/pkg/measurements/factory.go
+++ b/pkg/measurements/factory.go
@@ -20,6 +20,7 @@ import (
        "sync"
 
        "github.com/cloud-bulldozer/go-commons/indexers"
+       "github.com/kube-burner/kube-burner/pkg/burner"
        "github.com/kube-burner/kube-burner/pkg/config"
        "github.com/kube-burner/kube-burner/pkg/measurements/types"
        log "github.com/sirupsen/logrus"
@@ -29,7 +30,7 @@ import (
 )
 
 type measurementFactory struct {
-       jobConfig   *config.Job
+       job         *buner.Executor
        clientSet   kubernetes.Interface
        restConfig  *rest.Config
        createFuncs map[string]measurement
@@ -97,7 +98,7 @@ func (mf *measurementFactory) register(measurement types.Measurement, measuremen
        return nil
 }

As "burner" package itself is importing "measurements" package

pkg/measurements/metrics/watcher.go
rsevilla87: My bad, try this instead

```golang
metrics.NewWatcher(
		factory.clientSet.NetworkingV1().RESTClient().(*rest.RESTClient),    # Note where rest client is coming from
		"netpolWatcher",
		"networkpolicies",
		corev1.NamespaceAll,
		func(options *metav1.ListOptions) {
			options.LabelSelector = fmt.Sprintf("kube-burner-uuid=%v", globalCfg.UUID)
		},
		cache.Indexers{},
	)
``

Comment (venkataanil) [reply pkg/measurements/metrics/watcher.go:51]: Thanks Raul, I will take this change.

File: pkg/measurements/netpol_latency.go (hunk 6 above)
Comment (rsevilla87) [reply pkg/measurements/netpol_latency.go:249]: ok, makes sense, we can fix that in a follow-up pr

File: pkg/measurements/netpol_latency.go [hunk 7]
```diff
@@ -0,0 +1,614 @@
...
+	yamlToUnstructured(o.ObjectTemplate, cleanTemplate, uns)
+	return uns.GetKind()
+}
+
+// Start network policy measurement
+func (n *netpolLatency) start(measurementWg *sync.WaitGroup) error {
+	defer measurementWg.Done()
+	// Use proxyRoute to communicate with proxy pod
+	proxyRoute = n.config.Config["proxyRoute"]
+	// Skip latency measurement for 1st job which creates only pods
```
rsevilla87, line 499: kube-burner needs to be agnostic, we need remove any reference to the kube-burner-ocp network policy workload, like this one

pkg/measurements/netpol_latency.go
rsevilla87: This is not going to work well, once the namespace is created, kube-burner won't patch it with this label

pkg/measurements/netpol_latency.go
```diff
@@ -0,0 +1,614 @@
...
+				} else {
+					log.Errorf("Netpol workload supports only IN operator for NamespaceSelector.MatchExpressions")
+				}
+			}
+		}
+	}
+	return namespaces
+}
+
+func addPodsByLabel(ns string, ps *metav1.LabelSelector) []string {
```
rsevilla87, line 164: This function is complex, please add some docs

pkg/measurements/netpol_latency.go
```diff
@@ -0,0 +1,614 @@
...
+}
+
+// Send connections information to proxy pod
+func sendConnections() {
+	data, err := json.Marshal(connections)
+	if err != nil {
+		log.Fatalf("Failed to marshal payload: %v", err)
+	}
+
+	url := fmt.Sprintf("http://%s/initiate", proxyRoute)
```
rsevilla87, line 334: This pod should be created here, rather than in `kube-burner-ocp`, the network policy measurement needs to be executable from `kube-burner`, not only by `kube-burner-ocp`

rsevilla87: Also, the concept of route, is openshift specific, replace it by something more agnostic, like `proxyEndpoint`

pkg/measurements/netpol_latency.go
```diff
@@ -0,0 +1,614 @@
...
+	waitForCondition(fmt.Sprintf("http://%s/checkConnectionsStatus", proxyRoute))
+}
+
+// Periodically check if proxy pod finished communicataion with all client pods,
+// during 1) sending connection information 2) retrieving results
+func waitForCondition(url string) {
+	timeout := time.After(30 * time.Minute)
+	ticker := time.NewTicker(5 * time.Second)
+	defer ticker.Stop()
+
```
rsevilla87, line 353: Let's try to add some kind of log message here

pkg/measurements/netpol_latency.go
```diff
@@ -0,0 +1,614 @@
...
+	for {
+		select {
+		case <-timeout:
+			log.Fatalf("Timeout reached. Stopping further checks.")
+			return
+		case <-ticker.C:
+			resp, err := http.Get(url)
+			if err != nil {
+				log.Fatalf("Failed to check status: %v", err)
```
rsevilla87, line 362:
```suggestion
				log.Fatalf("Failed to check status from proxy: %v", err)
```

pkg/measurements/netpol_latency.go
venkataanil: If the pod is to be created by kube-burner code, how shoould it create the route (or the endpoint) for the pod?

pkg/measurements/netpol_latency.go
venkataanil: Any idea how to achieve that in this case?

rsevilla87 (commented):
The main issue I see with this PR is that we're trying to make this project distro agnostic, and it is doing lot of assumptions.
Also, the current implementation of the measurement is useless w/o the job described in kube-burner-ocp

pkg/measurements/netpol_latency.go
rsevilla87: Thinking on this again, there's no need to have a route, ingress or NodePort, there're other alternatives to access a remote API:

- port forward, you can find a good example directly in kubectl: https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/portforward/portforward.go#L420-L454
- remote command: [example](https://github.com/kube-burner/kube-burner/blob/main/pkg/measurements/util/svc_checker.go#L37-L74)

venkataanil:
I think the concerns here are 
1) using the labels in the job factory.jobConfig.NamespaceLabels["kube-burner.io/skip-networkpolicy-latency"]
2) pod and route created in kube-burner-ocp and using the route here
@rsevilla87   To fix them
1) I will try to avoid label and apply measurements to all the defined jobs.
2) I will let measurements package only create the proxy pod and use port forward (as you suggested in the comments) to communicate with this proxy pod
Do you have any other non-complicance concerns?

rsevilla87:
> @rsevilla87 I think the concerns here are
> 
>     1. using the labels in the job factory.jobConfig.NamespaceLabels["kube-burner.io/skip-networkpolicy-latency"]
> 
>     2. pod and route created in kube-burner-ocp and using the route here
> 
> 
> To fix them
> 
>     1. I will try to avoid label and apply measurements to all the defined jobs.
> 
>     2. I will let measurements package only create the proxy pod and use port forward (as you suggested in the comments) to communicate with this proxy pod
>        Do you have any other non-complicance concerns?

This measurement has as requirement to deploy several components and assets, they shouldn't be part of the workload itself, they should be deployed when the measurement is initialized, IIUC these assets are defined the first two jobs of the workload:

- network-policy-perf-pods
- convergence-tracker

Im concerned about convergence tracker, since this pod expects a series of environment variables.

venkataanil:
> > @rsevilla87 I think the concerns here are
> > ```
> > 1. using the labels in the job factory.jobConfig.NamespaceLabels["kube-burner.io/skip-networkpolicy-latency"]
> > 
> > 2. pod and route created in kube-burner-ocp and using the route here
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > To fix them
> > ```
> > 1. I will try to avoid label and apply measurements to all the defined jobs.
> > 
> > 2. I will let measurements package only create the proxy pod and use port forward (as you suggested in the comments) to communicate with this proxy pod
> >    Do you have any other non-complicance concerns?
> > ```
> 
> This measurement has as requirement to deploy several components and assets, they shouldn't be part of the workload itself, they should be deployed when the measurement is initialized, IIUC these assets are defined the first two jobs of the workload:
> 
> * network-policy-perf-pods
2 main reasons for this - 
1)  network policy is not applied only between the local pods. Instead the network policy create rules between the local pods and remote pods which are hosted on remote namespaces. For example, this https://github.com/venkataanil/files/blob/master/log_netpol.txt  is the output of all the network policies in a namespace network-policy-perf-0 (we got this from our test runs). This namespace has 50 ingress and 50 egress network policies.
Network policy ingress-0-29 ( https://github.com/venkataanil/files/blob/master/log_netpol.txt#L19372 ) is  allowing traffic from pods in namespaces network-policy-perf-41, network-policy-perf-42, network-policy-perf-43, network-policy-perf-44 and network-policy-perf-41 to pods in network-policy-perf-0.
In the current implementation, pods which are already created in "network-policy-perf-41, network-policy-perf-42, network-policy-perf-43, network-policy-perf-44 and network-policy-perf-41" try to send http requests to pods in "network-policy-perf-0" even before the nettwork policy "ingress-0-29" created. But these http requests will be failing as the network policy is not yet created. Once the "ingress-0-29" network policies created and ovs flows get programmed, the http requests will be succesful.
So this approach is helpful in testing connections between pods of different namespaces. Here all other resources (namespaces, pods, client app with peer addresses to send http requests) are ready and the dependency is only on the network policy creation to make the connection succesful.
2) Also as pods are already created, this pod latency doesn't impact the network policy latency. 

> * convergence-tracker
convergence-tracker is not stable in current implementation as well. So I will remove this from the PR for now.
> 
> Im concerned about convergence tracker, since this pod expects a series of environment variables.

docs/measurements/index.md
```diff
@@ -289,6 +289,90 @@ And the quantiles document has the structure:
 
 When there're `LoadBalancer` services, an extra document with `quantileName` as `LoadBalancer` is also generated as shown above.
 
+##Network Policy Latency
```
rsevilla87, line 292:
```suggestion
## Network Policy Latency
```

pkg/measurements/common.go
```diff
@@ -78,3 +86,53 @@ func calculateQuantiles(normLatencies []interface{}, getLatency func(interface{}
 
 	return latencyQuantiles
 }
+
+func DeployPod(namespace, podName, image string, command []string) error {
```
rsevilla87, line 90:
```suggestion
func deployPodInNamespace(namespace, podName, image string, command []string) error {
```

pkg/measurements/service_latency.go
```diff
@@ -186,7 +160,10 @@ func (s *serviceLatency) start(measurementWg *sync.WaitGroup) error {
 	// Reset latency slices, required in multi-job benchmarks
 	s.latencyQuantiles, s.normLatencies = nil, nil
 	defer measurementWg.Done()
-	if err := deployAssets(); err != nil {
+	svcLatencyCheckerImage := "quay.io/cloud-bulldozer/fedora-nc:latest"
+	svcLatencyCheckerCommand := []string{"sleep", "inf"}
+	err := DeployPod(types.SvcLatencyNs, types.SvcLatencyCheckerName, svcLatencyCheckerImage, svcLatencyCheckerCommand)
```
rsevilla87, line 165: We don't need variables if they're not used more than once.

```suggestion
	err := DeployPod(types.SvcLatencyNs, types.SvcLatencyCheckerName, "quay.io/cloud-bulldozer/fedora-nc:latest", []string{"sleep", "inf"})
```

pkg/measurements/netpol_latency.go
```diff
@@ -0,0 +1,645 @@
...
+			log.Debugf("Discarding network policy latency measurement for the job %v", factory.jobConfig.Name)
+			return nil
+		}
+	}
+	_, err = factory.clientSet.CoreV1().Pods(networkPolicyProxy).Get(context.TODO(), networkPolicyProxy, metav1.GetOptions{})
+	if err != nil {
+		netpolProxyImage := "quay.io/cloud-bulldozer/netpolproxy:latest"
+		err = DeployPod(networkPolicyProxy, networkPolicyProxy, netpolProxyImage, nil)
+		if err != nil {
+			log.Fatal(err)
```
rsevilla87, line 523: Not needed

pkg/measurements/util/port_forward.go
```diff
@@ -0,0 +1,97 @@
...
+		return "", fmt.Errorf("unable to parse local port: %w", err)
+	}
+	return localPort, nil
+}
+
+func NewPodPortForwarder(clientset kubernetes.Interface, restConfig rest.Config, remotePort, namespace, podName string) (PodPortForwarder, error) {
+	var localPort string
+	roundTripper, upgrader, err := spdy.RoundTripperFor(&restConfig)
+	if err != nil {
+		log.Fatalf("Error setting up transport: %v", err)
```
rsevilla87, line 41: log.Fatal is not needed before return as it aborts the execution of the binary

pkg/measurements/netpol_latency.go
```diff
@@ -0,0 +1,645 @@
...
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/kubectl/pkg/scheme"
+)
+
+var networkPolicyProxyPort string = "9002"
+var networkPolicyProxy = "network-policy-proxy"
+var proxyPortForwarder util.PodPortForwarder
+var nsPodAddresses = make(map[string]map[string][]string)
+var serverPingTestPort int32 = 8080
+var proxyEndpoint string
```
rsevilla87, line 59: A good practice is to use constants when possible to define inmutable variables

pkg/measurements/netpol_latency.go
```diff
@@ -0,0 +1,645 @@
...
+	if isEmpty(original) {
+		return nil, fmt.Errorf("template is empty")
+	}
+	r, err := regexp.Compile(`\{\{.*\}\}`)
+	if err != nil {
+		return nil, fmt.Errorf("regexp creation error: %v", err)
+	}
+	original = r.ReplaceAll(original, []byte{})
+	return original, nil
+}
```
rsevilla87, line 139: These functions are duped

pkg/measurements/netpol_latency.go
```diff
@@ -0,0 +1,645 @@
...
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusOK {
+		log.Debugf("Connection information sent successfully")
+	}
+	waitForCondition(fmt.Sprintf("http://%s/checkConnectionsStatus", proxyEndpoint))
+}
+
+// Periodically check if proxy pod finished communicataion with all client pods,
```
rsevilla87, line 356:
```suggestion
// Periodically check if proxy pod finished communication with all client pods,
```

pkg/measurements/netpol_latency.go
```diff
@@ -0,0 +1,645 @@
...
+		val, _ = stats.Percentile(npl, 99)
+		p99 := int(val)
+		val, _ = stats.Max(npl)
+		maxVal := int(val)
+		val, _ = stats.Mean(npl)
+		avgVal := int(val)
+		val, _ = stats.Min(npl)
+		minVal := int(val)
+		log.Tracef("netpol %s latency slice %v\n", name, npl)
+		log.Tracef("%s: 50th: %d 95th: %d 99th: %d min: %d max: %d avg: %d\n", name, p50, p95, p99, minVal, maxVal, avgVal)
```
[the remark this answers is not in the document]
venkataanil: I need min Values as well. Unlike other measurements, each network policy has set of values (each value represents a single connection test between the client & server pod defined by that network policy). We are capturing min and max from these values.
So lets retain this for timebeing.

examples/workloads/network-policy/templates/pod.yml
```diff
@@ -0,0 +1,32 @@
...
+        memory: "10Mi"
+        cpu: "10m"
+    imagePullPolicy: Always
+    ports:
+     - containerPort: 8080
+       protocol: TCP
+    securityContext:
+       privileged: false
+  - name: curlapp 
+    image: quay.io/vkommadi/netpolvalidator:latest
```
vishnuchalla, line 22: Can we get this added to `cloud-bulldozer` too?

examples/workloads/network-policy/templates/ingress-np.yml
```diff
@@ -0,0 +1,73 @@
...
+      {{- $next_namespace := print "network-policy-perf-" $nextNs }}
+      {{- $nsList = append $nsList $next_namespace }}
+    {{- end }}
+    {{- $nsNames := toJson $nsList }}
+  - from:
+    - namespaceSelector:
+        matchExpressions:
+          - key: kubernetes.io/metadata.name
+            operator: In
+            values: {{$nsNames}}
```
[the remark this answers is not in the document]
venkataanil: Can we take up that as an enhancement?

examples/workloads/network-policy/templates/pod.yml
venkataanil: My bad. Its already there. I will fix this typo.

examples/workloads/network-policy/templates/ingress-np.yml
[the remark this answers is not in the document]
vishnuchalla: sounds good 👍🏼

vishnuchalla (dismissed):
lgtm, except for Raul's comments. We can take the rest of the improvements in follow-up PRs.

pkg/measurements/netpol_latency.go
```diff
@@ -0,0 +1,623 @@
...
+		"podWatcher",
+		"pods",
+		corev1.NamespaceAll,
+		func(options *metav1.ListOptions) {
+			options.LabelSelector = fmt.Sprintf("kube-burner-runid=%v", globalCfg.RUNID)
+		},
+		cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc},
+	)
+	n.netpolLister = lnetv1.NewNetworkPolicyLister(n.netpolWatcher.Informer.GetIndexer())
+	n.podLister = lcorev1.NewPodLister(n.podWatcher.Informer.GetIndexer())
```
rsevilla87, line 546: Are these two uses somewhere?

docs/measurements/index.md
```diff
@@ -292,6 +292,90 @@ And the quantiles document has the structure:
...
 When there're `LoadBalancer` services, an extra document with `quantileName` as `LoadBalancer` is also generated as shown above.
 
+## Network Policy Latency
+
+Calculates the time taken to apply the network policy rules by the SDN through connection testing between pods specified in the network policy.
+At a high level, Kube-burner utilizes a proxy pod, `network-policy-proxy`, to distribute connection information (such as remote IP addresses) to client pods. These client pods then send requests to the specified addresses and record a timestamp once a connection is successfully established. Kube-burner retrieves these timestamps from the client pods via the proxy pod and calculates the network policy latency by comparing the recorded timestamp with the timestamp when the network policy was created.
+
+During this testing, kube-burner creates two separate jobs:
+1. **Job 1**: Creates all necessary namespaces and pods.
+2. **Job 2**: Applies network policies to the pods defined on above namespaces and tests connections by sending HTTP requests.
```
rsevilla87, line 302: I think the docs need some udpate with latest refactors

pkg/measurements/netpol_latency.go
```diff
@@ -0,0 +1,623 @@
...
+		func(options *metav1.ListOptions) {
+			options.LabelSelector = fmt.Sprintf("kube-burner-uuid=%v", globalCfg.UUID)
+		},
+		cache.Indexers{},
+	)
+	n.netpolWatcher.Informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: n.handleCreateNetpol,
+	})
+	// Pod watcher is needed to check if the pods are in active state or not
+	n.podWatcher = metrics.NewWatcher(
```
rsevilla87, line 535: Where is podWatcher used?

docs/measurements/index.md
rsevilla87, line 302: Do we still to define a workload as this one to make this measurement to work?

rsevilla87: I think this doc needs some update

venkataanil: Yes, I prefered this approach as per my earlier comment https://github.com/kube-burner/kube-burner/pull/679#issuecomment-2476159957
I will update the doc with the reason why we chose this approach.

pkg/measurements/netpol_latency.go
venkataanil: Earlier I was checking if the pods created in job1 are active before retrieving their ip addresses for preparing connection list. But this was not necessary so I removed those checks. So watcher is not needed, I will remove it.

pkg/measurements/netpol_latency.go
venkataanil: No, I will remove them.

venkataanil:
I see some code optimizations happening in kube-burner becuase of this PR. I am not having cycles to follow them and integrate into this network policy measurement.  Can we merge this PR and allow optimizations go in later?

rsevilla87 (dismissed):
lgtm

rsevilla87:
There's a conflict that needs to be resolved first @venkataanil

rsevilla87 (approved):

PR #682, opened 2024-08-15
remove unsupported init cli options

venkataanil:
## Type of change
- [ ] Documentation Update

## Description
Cli options which are not part of init command are removed from docs.


vishnuchalla (approved):
Thanks for adding this change.

Issue #684, opened 2024-08-22
[RFE] Map error codes to specific error types


dry923:
**Is your feature request related to a problem? Please describe.**
There is no clear way to tell why a job failed programatically based on its error code. Think running this in a wrapper or something that is going to look at the exit status and report on it.

**Describe the solution you'd like**
Map basic error types to specific error codes. Example:
| Exit Code | Code Meaning |
| ------------- | ---------------- |
| 0               | Success      |
| 1               | CLI Argument error/parsing error |
| 2               | Job timeout |
| 3               | Alerting metric not passing |
| 4               | Indexing failure |
| 5               | Create object failure |

**Describe alternatives you've considered**
Parsing the cli output can obtain some of these errors but its very variable and not clean.

**Additional context**
Ultimately the goal is to make it easy and programatically practical to tell why a job failed without having to parse its extensive output.

PR #688, opened 2024-08-29
Implement and explan extra return codes

rsevilla87:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Implementing new alert and measurement return codes and add a section to explain them.

@dry923, regarding the `Create object failure` return code you suggested in the #684, this is an error we're not currently tracking, at the moment kube-burner retries the object creation when it fails and the failure is different from:

- Object already exists: In this the execution goes on and the error is printed but ignored
- API authorization error: When this error happens kube-burner stops the benchmark executionand exits with rc 1

If you're curious, see the logic here https://github.com/kube-burner/kube-burner/blob/e1880a31d261626ebc210ec24e187a6ac3d03dce/pkg/burner/create.go#L274-L318

- Related Issue #684
- Closes #


dry923:
@rsevilla87 re the `Create object failure`, The current functionality is fine. I was just thinking of some examples. Thanks for picking this up!

rsevilla87:
Hey @dry923 , are yo missing anything on this PR?

vishnuchalla (approved):
lgtm

PR #689, opened 2024-08-30
Fix golangci-lint precommit hook

rsevilla87:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The previous hook wasn't working as expected as it was relying on this configuration https://github.com/golangci/golangci-lint/blob/master/.pre-commit-hooks.yaml#L1-L8 `golangci-lint run --new-from-rev HEAD --fix`, the hook's command basically fixes linting errors when the current code is different from HEAD (so no action is taken when changes are committed), this is not what we actually need, we want to report and hang on linting errors caused by a PR.

Also fixing some pieces of code to meet linting.

- Related Issue #
- Closes #

### DIFF
--- a/pkg/burner/create.go
+++ b/pkg/burner/create.go
@@ -298,12 +298,11 @@ func createRequest(gvr schema.GroupVersionResource, ns string, obj *unstructured
 			} else if kerrors.IsNotFound(err) {
 				log.Errorf("Error creating object %s/%s: %v", obj.GetKind(), obj.GetName(), err.Error())
 				return true, nil
+			}
+			if ns != "" {
+				log.Errorf("Error creating object %s/%s in namespace %s: %s", obj.GetKind(), obj.GetName(), ns, err)
 			} else {
-				if ns != "" {
-					log.Errorf("Error creating object %s/%s in namespace %s: %s", obj.GetKind(), obj.GetName(), ns, err)
-				} else {
-					log.Errorf("Error creating object %s/%s: %s", obj.GetKind(), obj.GetName(), err)
-				}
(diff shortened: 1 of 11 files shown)


pkg/burner/create.go
```diff
@@ -45,7 +45,6 @@ func setupCreateJob(jobConfig config.Job) Executor {
 	mapper := newRESTMapper()
 	log.Debugf("Preparing create job: %s", jobConfig.Name)
 	ex := Executor{}
-	ex.DefaultMissingKeysWithZero = jobConfig.DefaultMissingKeysWithZero
```
vishnuchalla, line 48: This change was intended to provide users flexibility to choose between replacing missing keys with default zero  or code logging errors when missing keys occur. Removing this will not let user to make a choice here if they wanted errors to be logged out when they have missing keys. It will simply replace with its default zero as per [line#237](https://github.com/kube-burner/kube-burner/pull/689/files#diff-76e5989cd670c9a6f994b653f5f41d8eacf7d309dc9074b48f9754bbac133018R237).

rsevilla87: Fixed, I removed that line mistakenly

pkg/burner/create.go
```diff
@@ -221,11 +221,6 @@ func (ex *Executor) replicaHandler(labels map[string]string, obj object, ns stri
 			copiedLabels[k] = v
 		}
 
-		templateOption := util.MissingKeyError
```
vishnuchalla, line 224: We need this as well as per my previous comment.

vishnuchalla (approved):
lgtm

PR #704, opened 2024-10-01
Remove os.exit in Run

venkataanil:
kube-burner-ocp workloads are using workload helpers for Run command. However this Run() has os.exit(), which is blocking kube-burner-ocp to run any PostRun() commands.

As we removed os.exit() here, kube-burner-ocp workloads explicitly need to define PostRun() with os.exit().

kube-burner-ocp PR https://github.com/kube-burner/kube-burner-ocp/pull/115 which adds PostRun() in all workloads

### DIFF
--- a/pkg/workloads/helpers.go
+++ b/pkg/workloads/helpers.go
@@ -44,7 +44,7 @@ func NewWorkloadHelper(config Config, embedConfig embed.FS, kubeClientProvider *
 	return wh
 }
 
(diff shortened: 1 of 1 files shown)


venkataanil:
This change is needed to add any cleanup in PostRun of kube-burner-workload.
I am adding network policy workload in kube-burner-ocp which creates a new namespace with pod, svc and route. I want to delete the namespace in PostRun of this kube-burner-workload.

PR #708, opened 2024-10-03
Handle absolute file system paths

nocturnalastro:
## Type of change

- [ ] Refactor
- [ ] New feature
- [X] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Running kube-burner with a config file with an absolute path was causing the it to be handled as a URL. I fixed it by treating urls with an empty host as a local file.


nocturnalastro:
Looks like the template files are handled relative to the execution not the config file.

PR #709, opened 2024-10-10
Better support StatefulSet

nocturnalastro:
## Type of change

- [ ] Refactor
- [ ] New feature
- [X] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

- allows the processing of StatefulSets by adding labels to both the template and the selector
- fetch the images for pre-loading

- Related Issue #
- Closes #

### DIFF
--- a/pkg/burner/utils.go
+++ b/pkg/burner/utils.go
@@ -71,7 +71,7 @@ func setMetadataLabels(obj *unstructured.Unstructured, labels map[string]string)
 	// us achieve that without breaking any of our labeling functionality.
 	templatePath := []string{"spec", "template", "metadata", "labels"}
 	setLabels(obj, labels, templatePath)
-	if obj.GetKind() == "StatefulSet" {
+	if obj.GetKind() == StatefulSet {
 		templatePath = []string{"spec", "selector", "matchLabels"}
 		setLabels(obj, labels, templatePath)
 	}
--- a/pkg/burner/utils.go
+++ b/pkg/burner/utils.go
@@ -52,6 +52,17 @@ func prepareTemplate(original []byte) ([]byte, error) {
 	return original, nil
 }
 
+func setLabels(obj *unstructured.Unstructured, labels map[string]string, templatePath []string) {
+	labelMap, found, _ := unstructured.NestedMap(obj.Object, templatePath...)
+	if !found {
+		return
+	}
+	for k, v := range labels {
+		labelMap[k] = v
+	}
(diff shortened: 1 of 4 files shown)


pkg/burner/utils.go
```diff
@@ -52,21 +53,30 @@ func prepareTemplate(original []byte) ([]byte, error) {
...
 	// object.SetLabels(labels) doesn't actually set labels for the underlying
 	// objects (i.e Pods under deployment/replicastes). So this function should help
 	// us achieve that without breaking any of our labeling functionality.
 	templatePath := []string{"spec", "template", "metadata", "labels"}
-	metadata, found, _ := unstructured.NestedMap(obj.Object, templatePath...)
-	if !found {
-		return
+	setLabels(obj, labels, templatePath)
+	if obj.GetKind() == StatefulSet {
+		templatePath = []string{"spec", "selector", "matchLabels"}
```
rsevilla87, line 76: please, refresh my mind... why are you using a different templatePath for stateful set?

rsevilla87: labels are already added at .spec.template.metadata.labels to statefulsets with the original implementation

nocturnalastro: See the comment on the template label here 
https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#components

rsevilla87: oh, sounds good to me, I don't use STS very regularly and I wasn't aware of this nuance

pkg/burner/waiters.go
```diff
@@ -60,7 +60,7 @@ func (ex *Executor) waitForObjects(ns string, limiter *rate.Limiter) {
 				waitForRS(waitNs, ex.MaxWaitTimeout, obj, limiter)
 			case "ReplicationController":
 				waitForRC(waitNs, ex.MaxWaitTimeout, obj, limiter)
-			case "StatefulSet":
+			case StatefulSet:
```
rsevilla87, line 63: We're not using constants for other kinds, in order to have a healthy code I think we should consolidate on use or not  touse them

nocturnalastro: The linter complained and failed the integration tests, because there was more than 3 instances of the string. 

If you want me to I'll make constants for the other types. I didn't want to do too much in this PR without otherwise being instructed too as I though it would be more confusing.

rsevilla87 (approved):
lgtm

PR #714, opened 2024-10-21
Add support for sequential handling on Patch jobs and objects

### LINKED ISSUE #713: [RFE] Add support for sequential patch job processing
Currently, patch jobs are handled in parallel - in both the objects and job iterations level. As a result, patching back and forth a certain value requires separate jobs which cannot be iterated. Also, there is no way to wait between objects or iterations.
For example, I would like to change the `nodeSelector` value, wait for a while and change it again. In the current implementation all requests are sent pretty much together and as a result, only some of the requests are actually handled.

Implement a way to run the objects sequentially with a way to wait between objects and also between job iterations

I can duplicate the jobs and use `jobPause`. But, this will produce an ugly test yaml with a lot of duplications.

**Additional context**
The test case I described is used to move pods with a PVC between nodes to test the ability of the CSI to move volumes between nodes

ygalblum:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Sequential execution runs all objects in a job iteration before reiterating
It allows delaying between objects and/or job iterations
Add PatchExecutionType for backward compatibility

- Related Issue #
- Closes #713

### DIFF
--- a/pkg/burner/patch.go
+++ b/pkg/burner/patch.go
@@ -34,12 +34,30 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+var (
+	defaultPatchExecutionMode   = config.PatchExecutionModeParallel
+	supportedPatchExecutionMode = map[config.PatchExecutionMode]struct{}{
+		config.PatchExecutionModeParallel:   {},
+		config.PatchExecutionModeSequential: {},
+	}
+)
+
 func setupPatchJob(jobConfig config.Job) Executor {
 	var f io.Reader
 	var err error
 	log.Debugf("Preparing patch job: %s", jobConfig.Name)
-	var ex Executor
-	ex.DefaultMissingKeysWithZero = jobConfig.DefaultMissingKeysWithZero
+
+	ex := Executor{
+		Job: jobConfig,
+	}
+
+	if len(ex.PatchExecutionMode) == 0 {
+		ex.PatchExecutionMode = defaultPatchExecutionMode
+	}
+	if _, ok := supportedPatchExecutionMode[ex.PatchExecutionMode]; !ok {
+		log.Fatalf("Unsupported Patch Execution Mode: %s", ex.PatchExecutionMode)
+	}
+
 	mapper := newRESTMapper()
 	for _, o := range jobConfig.Objects {
 		if o.APIVersion == "" {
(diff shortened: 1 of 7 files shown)


pkg/burner/patch.go
```diff
@@ -97,22 +114,66 @@ func (ex *Executor) RunPatchJob() {
...
 				wg.Add(1)
 				go ex.patchHandler(obj, item, i, &wg)
 			}
+			// Wait for all items in the object
+			wg.Wait()
+			waitRateLimiter := rate.NewLimiter(rate.Limit(restConfig.QPS), restConfig.Burst)
+			ex.waitForObjects("", waitRateLimiter)
+
+			// Wait between object
+			if ex.ObjectDelay > 0 {
```
rsevilla87, line 154: Thanks, this wasn't implemented previously

pkg/burner/patch.go
```diff
@@ -97,22 +114,66 @@ func (ex *Executor) RunPatchJob() {
...
 		// Try to find the list of resources by GroupVersionResource.
 		err := util.RetryWithExponentialBackOff(func() (done bool, err error) {
-			itemList, err = DynamicClient.Resource(obj.gvr).List(context.TODO(), listOptions)
+			itemList, err := DynamicClient.Resource(obj.gvr).List(context.TODO(), listOptions)
 			if err != nil {
 				log.Errorf("Error found listing %s labeled with %s: %s", obj.gvr.Resource, labelSelector, err)
 				return false, nil
 			}
+			log.Infof("Found %d %s with selector %s; patching them", len(itemList.Items), obj.gvr.Resource, labelSelector)
+			objItemLists = append(objItemLists, itemList)
```
rsevilla87, line 123: Just to clarify, now we list all objects meeting the labelSelectors and gvr and then we patch them, correct?

ygalblum: Yes, the idea was to avoid having to list the items on every job iteration. But, now I'm thinking about something. 
Let's say there are two objects and then the first object changes something that affects the list of the second one. Then we'll miss that. So, maybe I should revert this behaviour and list on every object in every iteration?

pkg/burner/patch.go
ygalblum: Without this, I had to use different jobs and then I can't use iterations

pkg/burner/patch.go
rsevilla87: In the original implementation, objects were not listed on a per iteration basis, but listed sequentially. Maybe the old implementation is still valid., the number of list API calls shouldn't be excessive

ygalblum: I'm looking at the original code again and the answer is yes and no. Yes, because the code did list the items of each object separately. But, no because first it ran all the iterations of each object before moving to the next object and it didn't wait for the commands of the object to complete before moving to the next object.
Having said that, I do think it's probably best to list per object which will also be per job

rsevilla87: correct, I don't have any issue with your implementation, (we're not actively using patch jobs in our workloads and I'm open to add changes like these w/o further discussion) just wanted to note that.

Is the PR ready to merge then?, I wonder if you can add a note to the docs (patch jobs section) explaining how this logic works

ygalblum: Yeah, you're right about the docs. Don't merge it just yet. I want it to go in together.

ygalblum:
@rsevilla87 I've made the change we discussed regarding running list on every run and added a test that verifies that it in fact works.
As for documentation, I already had a section on the different execution mode. I guess that because I wrote the code, I don't see what's missing. But, please LMK what information you think is missing

rsevilla87 (approved):
lgtm

PR #719, opened 2024-10-31
Add test output files to git ignore list

ygalblum:
## Type of change

- [x] Optimization

## Description

Running tests locally creates many log files that git should ignore


rsevilla87 (approved):
lgtm

PR #722, opened 2024-10-31
Use sleep without specifying the unit

ygalblum:
## Type of change

- [x] Optimization

## Description

Some implementations do not support setting the unit. Since seconds is already the default, no harm in just removing the s


rsevilla87 (approved):
lgtm

PR #724, opened 2024-10-31
CRD Test - use an independent dummy CRD

ygalblum:
## Type of change

- [x] Refactor

## Description
The test verifies creation of CRDs. Using real CRD might cause issues with the cluster. It's safer to use a dummy one/


rsevilla87 (approved):
lgtm

PR #726, opened 2024-11-02
[GH#683] fix: absolute file paths interpreted as urls

### LINKED ISSUE #683: [BUG] Absolute path breaks metrics endpoint parsing
# Bug Description

kube-burner v1.10.5

## **Describe the bug**
When we specify an absolute path for the metrics endpoint in the config. We are seeing the below error
```
time="2024-08-14 18:24:16" level=fatal msg="Error reading metricsEndpoint /root/work/e2e-benchmarking/workloads/kube-burner-ocp-wrapper/metrics-profiles/metrics-endpoint.yaml: Get "/root/work/e2e-benchmarking/workloads/kube-burner-ocp-wrapper/metrics-profiles/metrics-endpoint.yaml": unsupported protocol scheme """ file="utils.go:32"
```

## **To Reproduce**

1. Download kube-burner binary with v1.10.5 version.
2. Use it in your init sub command specifying an absolute metrics endpoint path as below
```
kube-burner init --metrics-endpoint "/home/xyz/metrics-endpoint.yaml"
```
3. This should reproduce the bug with the error above.

## **Expected behavior**

The above error should be gracefully handled by [this part](https://github.com/kube-burner/kube-burner/blob/main/pkg/util/url_reader.go#L33) of code block that we have for parsing the config file.

pfeifferj:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The current `ReadConfig` function attempts to read the `configFile` string as a URL. It falls back to treating it as a file path only if parsing the URL fails. This logic is causing the issue where absolute file paths pass the initial check. 

My suggested change uses `url.Parse` to check for a valid URL format. Instead of assuming any valid URI is a URL, it explicitly checks for the `http` or `https` schemes. Only if those schemes are present does it treat the `configFile` string as a URL.

- Closes #683

### DIFF
--- a/test/test-k8s.bats
+++ b/test/test-k8s.bats
@@ -40,9 +40,10 @@ teardown_file() {
   $OCI_BIN rm -f prometheus
 }
 
-@test "kube-burner init: churn=true" {
+@test "kube-burner init: churn=true; absolute-path=true" {
   export CHURN=true
   export CHURN_CYCLES=2
+  cp kube-burner.yml /tmp/kube-burner.yml
   run_cmd kube-burner init -c kube-burner.yml --uuid="${UUID}" --log-level=debug
   check_destroyed_ns kube-burner-job=not-namespaced,kube-burner-uuid="${UUID}"
   check_destroyed_pods default kube-burner-job=not-namespaced,kube-burner-uuid="${UUID}"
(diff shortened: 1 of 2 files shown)


test/test-k8s.bats
```diff
@@ -40,6 +40,13 @@ teardown_file() {
   $OCI_BIN rm -f prometheus
 }
 
+@test "kube-burner init: absolute-path=true" {
```
rsevilla87, line 43: Rather than adding an extra test case, can you update the next one to use absolute path?, that way we can save some time on testing.

rsevilla87 (approved):
lgtm, thanks!

PR #729, opened 2024-11-05
Reuse runJob for Delete and Read

ygalblum:
## Type of change

- [x] Refactor

## Description

Rename executionMode to unlink from patch
Move runJob and its internals to utiles  and use in Read and Delete jobs
Add ObjectFinalizer in addition to ItemHandler
Set Job for Executor and make sure it is not accesses once it is passed to executor
Remove redundant labelSelector and patchType fields from object

### DIFF
--- a/pkg/burner/delete.go
+++ b/pkg/burner/delete.go
@@ -20,7 +20,6 @@ import (
 	"time"
 
 	"github.com/kube-burner/kube-burner/pkg/config"
-	"github.com/kube-burner/kube-burner/pkg/util"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -31,8 +30,20 @@ import (
 )
 
 func setupDeleteJob(jobConfig config.Job) Executor {
-	var ex Executor
 	log.Debugf("Preparing delete job: %s", jobConfig.Name)
+	ex := Executor{
+		Job:         jobConfig,
+		itemHandler: deleteHandler,
+	}
+	if ex.WaitForDeletion {
+		ex.objectFinalizer = verifyDelete
+	}
+
+	// Only a single iteration is supported for Delete job
+	ex.JobIterations = 1
+	// Use the sequencial mode
+	ex.ExecutionMode = config.ExecutionModeSequential
+
 	mapper := newRESTMapper()
 	for _, o := range jobConfig.Objects {
 		if o.APIVersion == "" {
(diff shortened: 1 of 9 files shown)


ygalblum:
@rsevilla87 @vishnuchalla This is a first stage in some code consolidation work that I did in preparation of adding new functionality

pkg/burner/job.go
```diff
@@ -343,19 +349,18 @@ func newExecutorList(configSpec config.Spec, kubeClientProvider *config.KubeClie
 			log.Fatalf("Unknown jobType: %s", job.JobType)
 		}
 		for _, j := range executorList {
-			if job.Name == j.Job.Name {
-				log.Fatalf("Job names must be unique: %s", job.Name)
+			if ex.Name == j.Name {
+				log.Fatalf("Job names must be unique: %s", ex.Name)
 			}
```
rsevilla87, line 354: We can remove this conditional, just realized that this is checked by the jobIsDuped() function in config/config.go

pkg/burner/job.go
```diff
@@ -43,23 +44,26 @@ import (
...
+	gvr        schema.GroupVersionResource
+	objectSpec []byte
+	kind       string
+	namespace  string
 	config.Object
 }
 
 // Executor contains the information required to execute a job
+type ItemHandler func(ex *Executor, obj object, originalItem unstructured.Unstructured, iteration int, wg *sync.WaitGroup)
+type ObjectFinalizer func(obj object)
```
rsevilla87, line 56: To we actually need these two new types?

pkg/burner/delete.go
```diff
@@ -47,72 +58,47 @@ func setupDeleteJob(jobConfig config.Job) Executor {
 			log.Fatalf("Empty labelSelectors not allowed with: %s", o.Kind)
 		}
 		obj := object{
-			gvr:           mapping.Resource,
-			labelSelector: o.LabelSelector,
```
rsevilla87, line 51: Interesting... seems we can use the LabelSelector attribute directly from the config/object, rather than re-defining it

ygalblum: Yes, since executor extends the Job it can access all fields directly.

pkg/burner/job.go
ygalblum: These types are used in the `Executor` struct below. I think it's cleaner and easier to read than just adding the method signature in the struct's definition.

pkg/burner/job.go
```diff
@@ -141,8 +152,7 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
 			measurements.SetJobConfig(&job.Job, kubeClientProvider)
 			log.Infof("Triggering job: %s", job.Name)
 			measurements.Start()
-			switch job.JobType {
-			case config.CreationJob:
+			if job.JobType == config.CreationJob {
```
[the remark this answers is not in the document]
ygalblum: Yes, the create operation is different than the rest and that's why I did not address it yet. We can address it later

pkg/burner/job.go
ygalblum: Done

rsevilla87 (approved):
lgtm

PR #730, opened 2024-11-06
Plugable measurements

rsevilla87:
## Type of change

- [x] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [x] Documentation Update

## Description

Add new mechanism to add custom measurements from wrappers. It can be used by populating `CustomMeasurements` in `workloads/types.go`

As new measurements can use an arbitrary configuration, it was needed to create a new field `config`  in each measurement, which holds the OOTB measurement configurations and arbitrary fields.

This is a **breaking** change and will require to update configuration files accordingly.

- Related Issue #
- Closes #

### DIFF
--- a/docs/measurements/index.md
+++ b/docs/measurements/index.md
@@ -97,7 +97,7 @@ And the metrics are:
 
 ### Pod latency thresholds
 
-It is possible to establish pod latency thresholds to the different pod conditions and metrics by defining the option `thresholds` within in the the config of this measurement:
+It is possible to establish pod latency thresholds to the different pod conditions and metrics by defining the option `thresholds` within the config of this measurement:
 
 Establishing a threshold of 2000ms in the P99 metric of the `Ready` condition.
 
--- a/docs/measurements/index.md
+++ b/docs/measurements/index.md
@@ -97,17 +97,18 @@ And the metrics are:
 
 ### Pod latency thresholds
 
-It is possible to establish pod latency thresholds to the different pod conditions and metrics by defining the option `thresholds` within this measurement:
+It is possible to establish pod latency thresholds to the different pod conditions and metrics by defining the option `thresholds` within in the the config of this measurement:
 
 Establishing a threshold of 2000ms in the P99 metric of the `Ready` condition.
 
 ```yaml
   measurements:
   - name: podLatency
-    thresholds:
-    - conditionType: Ready
-      metric: P99
-      threshold: 2000ms
+    config:
+      thresholds:
+      - conditionType: Ready
+        metric: P99
+        threshold: 2000ms
 ```
 
 Latency thresholds are evaluated at the end of each job, showing an informative message like the following:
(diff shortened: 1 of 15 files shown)


venkataanil (dismissed):
LGTM
Thanks for bringing up this which is needed in https://github.com/kube-burner/kube-burner/pull/679

test/test-k8s.bats
```diff
@@ -156,11 +159,6 @@ teardown_file() {
   check_file_list alerts/alert.json
 }
 
-@test "kube-burner log file output" {
```
vishnuchalla, line 159: Guessing the reason behind removing this is because of not having enough weight to be a functional test. CMIIW.

docs/measurements/index.md
```diff
@@ -91,17 +91,18 @@ And the metrics are:
 
 ### Pod latency thresholds
 
-It is possible to establish pod latency thresholds to the different pod conditions and metrics by defining the option `thresholds` within this measurement:
+It is possible to establish pod latency thresholds to the different pod conditions and metrics by defining the option `thresholds` within in the the config of this measurement:
```
vishnuchalla, line 94: `the` has been duplicated in the text

pkg/workloads/types.go
```diff
@@ -20,16 +20,23 @@ import (
 
 	ocpmetadata "github.com/cloud-bulldozer/go-commons/ocp-metadata"
 	"github.com/kube-burner/kube-burner/pkg/config"
+	"github.com/kube-burner/kube-burner/pkg/measurements"
 )
 
+type CustomMeasurement struct {
```
vishnuchalla, line 26: Code changes are good. Can we add a small snippet example here as well? Similar to the documentation on how to pass a function as parameter in template rendering. Might be useful to someone not much familiar with golang

pkg/measurements/types/types.go
```diff
@@ -27,23 +27,22 @@ const (
 )
 
 // UnmarshalYAML implements Unmarshaller to customize object defaults
-func (m *Measurement) UnmarshalMeasurement(unmarshal func(interface{}) error) error {
-	type rawMeasurement Measurement
-	measurement := rawMeasurement{
+func (m *MeasurementConfig) UnmarshalMeasurementConfig(unmarshal func(interface{}) error) error {
+	type rawMeasurementConfig MeasurementConfig
+	measurementCfg := rawMeasurementConfig{
```
vishnuchalla, line 32:
```suggestion
	measurementCfg := MeasurementConfig{
```

vishnuchalla:
@rsevilla87 I am not sure if we have a use case for it right away. If we have one, can we get these changes verified from ocp-wrapper as well to uncover any unknowns earlier?

test/test-k8s.bats
rsevilla87: correct

rsevilla87:
Closing in favor of https://github.com/kube-burner/kube-burner/pull/841

PR #732, opened 2024-11-08
Tests - allow running against an existing cluster

ygalblum:
## Type of change

- [x] Optimization

## Description

Allow running the tests with an existing cluster instead of running a dedicated one using Kind. This resolves cases in which users do not have root access in order to use rootless podman or if KVM is not supported (e.g. MacOS)

### DIFF
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -34,8 +34,15 @@ setup-kind() {
 
 create_test_kubeconfig() {
   echo "Creating another kubeconfig"
-  "${KIND_FOLDER}/kind-linux-${ARCH}" export kubeconfig --kubeconfig "${TEST_KUBECONFIG}"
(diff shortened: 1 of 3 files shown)


test/helpers.bash
```diff
@@ -34,8 +34,15 @@ setup-kind() {
 
 create_test_kubeconfig() {
   echo "Creating another kubeconfig"
-  "${KIND_FOLDER}/kind-linux-${ARCH}" export kubeconfig --kubeconfig "${TEST_KUBECONFIG}"
-  kubectl config rename-context kind-kind "${TEST_KUBECONTEXT}" --kubeconfig "${TEST_KUBECONFIG}"
+  if [[ "${USE_EXISTING_CLUSTER,,}" == "yes" ]]; then
```
rsevilla87, line 37: Are these commas correct?

rsevilla87 (approved):
lgtm!

PR #733, opened 2024-11-12
Reuse Executor and object setup code

ygalblum:
## Type of change

- [x] Refactor

## Description

Consolidate the setup and creation of object and executor instances into a generic code.
For object, keep create with the old implantation as it is different from the rest

### DIFF
--- a/pkg/burner/job.go
+++ b/pkg/burner/job.go
@@ -34,37 +34,12 @@ import (
 	"github.com/kube-burner/kube-burner/pkg/util/metrics"
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/time/rate"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
-
-type object struct {
-	gvr        schema.GroupVersionResource
-	objectSpec []byte
-	kind       string
-	namespace  string
-	config.Object
-}
-
-// Executor contains the information required to execute a job
-type ItemHandler func(ex *Executor, obj object, originalItem unstructured.Unstructured, iteration int, wg *sync.WaitGroup)
-type ObjectFinalizer func(obj object)
-
-type Executor struct {
-	config.Job
-	objects         []object
-	uuid            string
-	runid           string
-	limiter         *rate.Limiter
-	nsRequired      bool
-	itemHandler     ItemHandler
-	objectFinalizer ObjectFinalizer
-}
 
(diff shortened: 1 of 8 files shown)


pkg/burner/object.go
```diff
@@ -0,0 +1,71 @@
...
+		log.Fatal(err)
+	}
+	obj.Namespaced = mapping.Scope.Name() == meta.RESTScopeNameNamespace
+
+	o := object{
+		Object: obj,
+		gvr:    mapping.Resource,
+	}
+
+	if len(obj.ObjectTemplate) != 0 {
```
rsevilla87, line 57: I think this way is more native
```suggestion
	if obj.ObjectTemplate != "" {
```

pkg/burner/object.go
```diff
@@ -0,0 +1,71 @@
...
+
+type object struct {
+	config.Object
+	gvr        schema.GroupVersionResource
+	objectSpec []byte
+	kind       string
+	namespace  string
+}
+
+func newObject(obj config.Object, mapper meta.RESTMapper) object {
```
rsevilla87, line 36: We probably don't need to pass mapper as argument for each object

ygalblum: Since the method needs a mapper, it can either create it use get it an argument. Getting it as a argument allows creating it only once instead of for each object.
Generally speaking, passing such elements as arguments, also allows simpler unit testing (I know we don't have them here, but still it's a good practice).

pkg/burner/job.go
```diff
@@ -329,42 +295,45 @@ func indexMetrics(uuid string, executedJobs []prometheus.Job, returnMap map[stri
...
+func verifyJobDefaults(job *config.Job, defaultTimeout time.Duration) {
+	verifyJobTimeout(job, defaultTimeout)
+	verifyQPSBurst(job)
+}
+
 // newExecutorList Returns a list of executors
-func newExecutorList(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider, uuid string, timeout time.Duration) []Executor {
-	var ex Executor
-	var executorList []Executor
+func newExecutorList(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider, defaultTimeout time.Duration) []*Executor {
```
rsevilla87, line 324: Can you move this function to executor.go?

pkg/burner/object.go
```diff
@@ -0,0 +1,71 @@
...
+	if obj.APIVersion == "" {
+		obj.APIVersion = "v1"
+	}
+
+	if len(obj.LabelSelector) == 0 {
```
rsevilla87, line 41:
```suggestion
	if obj.LabelSelector == "" {
```

pkg/burner/job.go
ygalblum: You are right. But, I wanted to limit the changes in this PR to the code consolidation. There are other moves this this (e.g. `RunJob`). So, I think it's best to do that in a followup PR

pkg/burner/object.go
ygalblum: Done

PR #734, opened 2024-11-12
Add example for moving Pods with PVCs between Nodes

ygalblum:
## Type of change

- [x] New feature

## Description
Test the CSI's ability to move volumes between nodes by forcing nodeSelector


rsevilla87 (approved):
lgtm

PR #737, opened 2024-11-15
Copy the labels to the underlying template even if it does not have any

### LINKED ISSUE #736: [BUG] VMI measurements are not collected if the VMI template does not include any labels
# Bug Description

If the VMI template does not include any labels, VMI measurements are not collected

Version: 1.10.9
Git Commit: 14132166a0f6c276454a529cabfbf5b0ad36720a
Build Date: 2024-10-16T11:09:49Z
Go Version: go1.21.13
OS/Arch: darwin arm64

## **Describe the bug**

If the VMI template does not include any labels, VMI measurements are not collected

## **To Reproduce**

1. Use `kube-burner` to create a VM without setting any labels in the VMI template object
2. Make sure to add `vmiLatency` messurement

## **Expected behavior**

VMI measurements exist

## **Screenshots or output**

No VMI measurements

ygalblum:
## Type of change

- [x] Bug fix

## Description

kube-burner copies the labels of its created objects into the template of the underlying object (e.g. pod template of a deployment). These labels are then used in other places in the code. For example, metrics measurements rely on the existence of the `kube-burner-runid` label on the pod (or VMI in case of VMs).

In the original code, the labels were copied only if the labels field already existed. 

With these fix, for specific kinds, the labels are copied even if none were added.

- Closes #736

### DIFF
--- a/pkg/burner/utils.go
+++ b/pkg/burner/utils.go
@@ -43,6 +43,18 @@ const (
 	objectLimit = 500
 )
 
+var (
+	commonUnderlyingObjectLabelsPath = []string{"spec", "template", "metadata", "labels"}
+
+	kindToLabelPaths = map[string][][]string{
+		DaemonSet:      {commonUnderlyingObjectLabelsPath},
+		Deployment:     {commonUnderlyingObjectLabelsPath},
+		ReplicaSet:     {commonUnderlyingObjectLabelsPath},
+		StatefulSet:    {commonUnderlyingObjectLabelsPath, []string{"spec", "selector", "matchLabels"}},
+		VirtualMachine: {commonUnderlyingObjectLabelsPath},
+	}
+)
+
 func prepareTemplate(original []byte) ([]byte, error) {
 	// Removing all placeholders from template.
 	// This needs to be done due to placeholders not being valid yaml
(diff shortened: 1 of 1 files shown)


pkg/burner/utils.go
```diff
@@ -75,12 +87,12 @@ func setMetadataLabels(obj *unstructured.Unstructured, labels map[string]string)
 	// objects (i.e Pods under deployment/replicastes). So this function should help
 	// us achieve that without breaking any of our labeling functionality.
 	templatePath := []string{"spec", "template", "metadata", "labels"}
-	setLabels(obj, labels, templatePath)
+	_, force := kindToForceLabels[obj.GetKind()]
+	setLabels(obj, labels, templatePath, force)
```
rsevilla87, line 91: The "force" logic is a bit weird, maybe it's better to use a conditinal?

```golang
	if _, exists := kindsToLabel["foo"]; exists {
        setLabels(obj, labels, templatePath)
         if obj.GetKind() == StatefulSet {
		     templatePath = []string{"spec", "selector", "matchLabels"}
		    setLabels(obj, labels, templatePath)

	    }
    }
```

ygalblum: I did this for backward compatibility. With your suggestion, if there's a kind I didn't think of that used this feature it will break. 
I don't mind changing the implementation if you think it's better

ygalblum: Done

ygalblum: Done

rsevilla87 (approved):
lgtm

Issue #738, opened 2024-11-15
[BUG] Pod Latency measurement might give false positive run at large scale


mukrishn:
# **Bug Description**
Pod Latency measurement misses out few pods and gives false positive results at large scale executions.

## **Version:**
kube-burner version
Version: 1.9.6
Git Commit: a51ce29e53ac6e4f022818e642295d993ae695f5
Build Date: 2024-05-14T16:57:20Z
Go Version: go1.21.9
OS/Arch: linux amd64

kube-burner-ocp version
Version: 1.4.4
Git Commit: 762ae83e3c9fee8d8c91515bbfa96ccc07df41af
Build Date: 2024-11-01T13:18:57Z
Go Version: go1.22.3
OS/Arch: linux amd64

## **Details:**

On running a large scale cluster-density-v2 test on ROSA-HCP, we noticed some of the pods took longer to became ready but they were missed/not-included in the latency measurement. The report has not captured the result of the complete execution, as it seemed to be missing some pod data. This affected overall pod latency measurement and provided a false positive result. 

Report - [dash](https://grafana.rdu2.scalelab.redhat.com:3000/d/e7788d90-8180-4690-abe5-84538b27cf52/kube-burner-report-ocp-wrapper?orgId=1&from=1731677516005&to=1731678921176&var-Datasource=ab3f14e6-a50f-4d52-93fa-a5076794f864&var-platform=AWS&var-sdn=OVNKubernetes&var-job=cluster-density-v2&var-nodes=90&var-uuid=4d4f66fa-e1dd-4876-b42e-093c3cb3bd96&var-master=&var-worker=ip-10-0-1-101.us-east-2.compute.internal&var-infra=ip-10-0-1-216.us-east-2.compute.internal&var-latencyPercentile=P99)

this namespace took more than an hr to became ready but it was not captured in the report
```
time="2024-11-15 13:52:55" level=debug msg="Waiting for replicas from deployments in ns cluster-density-v2-531 to be ready" file="waiters.go:99"
time="2024-11-15 13:52:56" level=debug msg="Waiting for replicas from deployments in ns cluster-density-v2-531 to be ready" file="waiters.go:99"
...
...
time="2024-11-15 14:50:00" level=debug msg="Waiting for replicas from deployments in ns cluster-density-v2-531 to be ready" file="waiters.go:99"
E1115 14:50:15.053862 2147632 request.go:1255] Unexpected error when reading response body: unexpected EOF
[...]

PR #739, opened 2024-11-18
Optimize waiters

rsevilla87:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

Multiple improvements to the waiters logic:

- Remove code duplication with the new function `genericWait()`, which receives as argument the YAML path of the conditions to wait for.
- These YAML paths are in the new variable `waitStatusMap`
- The setup<action>Job functions are now methods of the Executor struct, thanks to that, some duplication on these functions is removed
- The executor has now a `waitLimiter` field that is used across all waiters w/o having to pass this limiter as argument across the different functions.
- In the previous implementation, then a error in the clientSet happens (API error, timeout, etc), kube-burner returned `return false, err`, leading the `wait.PollUntilContextTimeout` loop to exit w/o providing any kind of information to the user, now the error is printed and the loop is never scaped on an error condition (`return false, nil`). This improvement should help to mitigate #738
- We now capture and return errors from `wait.PollUntilContextTimeout(context.TODO(), time.Second, ex.MaxWaitTimeout, true, func(ctx context.Context) (done bool, err error)`. Not very useful as there's not any error able to cause this loop to finish, but at least the logic is more correct.
- Consolidate the code to wait for VMIs and VMs into a single function `waitForVMorVMI()`

- Related Issue #738
- Closes #


rsevilla87:
cc @mukrishn @jtaleric @dry923 @ygalblum

pkg/burner/job.go
```diff
@@ -48,6 +48,7 @@ type object struct {
 	objectSpec []byte
 	kind       string
 	namespace  string
+	ready      bool
```
rsevilla87, line 51: This new field is now used to prevent kube-burner to wait N times (as many as namespaces) for objects with a prefixed namespace or with the runOnce flag. as these objects only need to be waited once

pkg/burner/job.go
```diff
@@ -331,33 +333,33 @@ func indexMetrics(uuid string, executedJobs []prometheus.Job, returnMap map[stri
...
 		if ex.MaxWaitTimeout == 0 {
 			log.Debugf("job.MaxWaitTimeout is zero in %s, override by timeout: %s", ex.Name, timeout)
 			ex.MaxWaitTimeout = timeout
-		} else {
-			log.Debugf("job.MaxWaitTimeout is non zero in %s: %s", ex.Name, ex.MaxWaitTimeout)
 		}
-		// Limits the number of workers to QPS and Burst
-		ex.limiter = rate.NewLimiter(rate.Limit(ex.QPS), ex.Burst)
-		ex.uuid = uuid
-		ex.runid = configSpec.GlobalConfig.RUNID
```
rsevilla87, line 360: The job config and these parameters are now configured at the beginning of the loop in a more elegant way

pkg/burner/types/types.go
```diff
@@ -14,20 +14,6 @@
 
 package types
 
-const (
```
rsevilla87, line 17: All this stuff is already automatically discovered and configured in obj.gvr by kube-burner

pkg/burner/waiters.go
```diff
@@ -21,135 +21,102 @@ import (
...
+	} else {
+		log.Info("Actions completed")
+	}
 }
 
-func waitForRS(ns string, maxWaitTimeout time.Duration, obj object, limiter *rate.Limiter) {
-	wait.PollUntilContextTimeout(context.TODO(), time.Second, maxWaitTimeout, true, func(ctx context.Context) (done bool, err error) {
-		limiter.Wait(context.TODO())
-		rss, err := ClientSet.AppsV1().ReplicaSets(ns).List(context.TODO(), metav1.ListOptions{
+func (ex *Executor) genericWait(ns string, obj object, waitPath statusPath) error {
```
ygalblum, line 82: This method focuses on replica readiness. So, consider renaming to something with `replica` in its name instead of `generic`

pkg/burner/consts.go
```diff
@@ -15,3 +15,24 @@ const (
...
 )
+
+type statusPath struct {
+	expectedReplicasPath []string
+	readyReplicasPath    []string
+}
+
+var waitStatusMap map[string]statusPath = map[string]statusPath{
+	StatefulSet: {
+		expectedReplicasPath: []string{"spec", "replicas"}, readyReplicasPath: []string{"status", "readyReplicas"}},
```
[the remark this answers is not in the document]
rsevilla87: fixed

pkg/burner/waiters.go
rsevilla87: updated

pkg/burner/job.go
```diff
@@ -90,7 +89,7 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
 	uuid := configSpec.GlobalConfig.UUID
 	globalConfig := configSpec.GlobalConfig
 	globalWaitMap := make(map[string][]string)
-	executorMap := make(map[string]*Executor)
+	executorMap := make(map[string]Executor)
```
rsevilla87, line 92: Since the setup functions are now Executor methods, we don't need this to be an array of pointers anymore, which simplifies the code

pkg/burner/waiters.go
```diff
@@ -21,135 +21,102 @@ import (
...
 		if err != nil {
-			return false, err
+			log.Fatalf("Error waiting for objects in namespace %s: %v", ns, err)
 		}
-		for _, dep := range deps.Items {
-			if *dep.Spec.Replicas != dep.Status.ReadyReplicas {
-				log.Debugf("Waiting for replicas from deployments in ns %s to be ready", ns)
-				return false, nil
-			}
+		if obj.namespace != "" || obj.RunOnce {
```
vishnuchalla, line 71: Is this check intended for light weight object like secrets, configmaps etc?

pkg/burner/consts.go
```diff
@@ -15,3 +15,27 @@ const (
...
+
+type statusPath struct {
+	expectedReplicasPath []string
+	readyReplicasPath    []string
+}
+
+var pathSpecReplicas = []string{"spec", "replicas"}
+var statusReplicas = []string{"status", "readyReplicas"}
+
+var waitStatusMap map[string]statusPath = map[string]statusPath{
```
vishnuchalla, line 27: simplifies things a lot.

vishnuchalla (approved):
lgtm

pkg/burner/waiters.go
rsevilla87: No, it's intended to only wait for these type of objects (objects with a pre-defined namespace or with the runOnce flag enabled) only once

PR #740, opened 2024-11-20
Remove global variables from job

ygalblum:
## Type of change

- [x] Refactor
- [x] Optimization

## Description

Global variables are in general a cause for issues; they can be set unexpectedly (or not set at all), they can be misused, etc. Instead, local variables, method parameters and instance fields should be used.


ygalblum:
@rsevilla87 I didn't squash the commits to make it easier to review how each variable is removed

rsevilla87:
> @rsevilla87 I didn't squash the commits to make it easier to review how each variable is removed

seems lt needs some rebase after the latest commits

ygalblum:
> > @rsevilla87 I didn't squash the commits to make it easier to review how each variable is removed
> 
> seems lt needs some rebase after the latest commits

Done

rsevilla87 (approved):
lgtm

PR #741, opened 2024-11-20
Pod and VMI Measurements: Add information about iteration and replica

ygalblum:
## Type of change

- [x] Optimization

## Description

Add the job iteration and replica count into the measurements JSON to allow post processing jobs to differentiate between iterations


rsevilla87 (approved):
lgtm

PR #748, opened 2024-12-02
New Job Type - Kubevirt

ygalblum:
## Type of change

- [x] New feature

## Description

Add support for kubevirt operations, similar to running `virtctl`.


ygalblum:
@rsevilla87 I've never worked with the License Compliance tool. I can't seem to get the information why it is now failing. It didn't fail before. Maybe you get a different view as a maintainer of the project

pkg/burner/kubevirt.go
```diff
@@ -0,0 +1,286 @@
...
+
+	for _, o := range ex.Objects {
+		if len(o.KubeVirtOp) == 0 {
+			log.Fatalln("Empty kubeVirtOp not allowed")
+		}
+		if _, ok := supportedOps[o.KubeVirtOp]; !ok {
+			log.Fatalf("Unsupported KubeVirtOp: %s", o.KubeVirtOp)
+		}
+
+		if len(o.Kind) == 0 {
```
rsevilla87, line 75: If I understood correctly, kind is a parameter of this type of objects

pkg/burner/kubevirt.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+func kubeOpHandler(ex *Executor, obj object, item unstructured.Unstructured, iteration int, wg *sync.WaitGroup) {
+	defer wg.Done()
+
+	var err error
+	switch obj.KubeVirtOp {
+	case config.KubeVirtOpStart:
+		options := kubevirtV1.StartOptions{}
+		startPaused := util.GetBoolValue(obj.InputVars, "startPaused")
```
rsevilla87, line 90: Thoughts on extending the object struct to add these flags?

pkg/burner/kubevirt.go
```diff
@@ -0,0 +1,286 @@
...
+		if retry < maxRetries {
+			time.Sleep(time.Duration(retry*(rand.IntN(5))) * time.Millisecond)
+		}
+	}
+
+	if err != nil && retry == maxRetries {
+		return fmt.Errorf("error removing volume after %d retries", maxRetries)
+	}
+
+	fmt.Printf("Successfully submitted remove volume request to VM %s for volume %s\n", vmiName, *volumeName)
```
rsevilla87, line 283:
```suggestion
	log.Debugf("Successfully submitted remove volume request to VM %s for volume %s", vmiName, *volumeName)
```

docs/reference/configuration.md
```diff
@@ -280,6 +280,82 @@ jobs:
...
+### Kubevirt
+
+This type of job can be used to execute `virtctl` commands described in the object list. This object list has the following structure:
+
+```yaml
+objects:
+- kubeVirtOp: start
+  labelSelector: {kube-burner-job: cluster-density}
+  inputVars:
+    force: true
```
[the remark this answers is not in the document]
ygalblum: No, it doesn't. I followed the same scheme as the `patch` `jobType` using `labelSelectors` for filtering the objects to run the command on

pkg/burner/kubevirt.go
ygalblum: I don't think so. The `Object` structure is used everywhere so I rather make minimal changes to it. I think using the `inputVars` served the purpose. Using it also allows adding support for other parameters (if and when needed) without changing the `Object`

pkg/burner/kubevirt.go
ygalblum: Not really. But, it is part of the `Object`. So, I'm letting users set it, though they probably won't. We can also change it to always set it overriding whatever the user sets

rsevilla87 (approved):
lgtm

rsevilla87 (commented):
I don't consider this PR an invasive change as it doesn't touch any of the current critical functionality

PR #750, opened 2024-12-03
fix localport in port forwarding

### LINKED ISSUE #749: [BUG] port forwarding code unable to parse local port information
# Bug Description

kube-burner doesn't use any specific local port for port forwarding. So a random port will be selected by client-go portforward code. Later kube-burner uses function parsePort() to parse the output and retrive the local port. However this function is expecting the output to parse as
"Forwarding from 127.0.0.1:8000 -> 4000"

However in this case, the output is
time="2024-12-03 15:16:53" level=info msg="Forwarding from 127.0.0.1:38709 -> 9002\nForwarding from [::1]:38709 -> 9002\n" file="port_forward.go:77"

So parsePort() is unable to fetch the local port as output has both ipv4 and ipv6 info.

[root@ip-172-31-32-243 kube-burner]# ss -tulnpa | grep kube-burner
tcp   LISTEN   0      4096            127.0.0.1:38709        0.0.0.0:*     users:(("kube-burner",pid=393077,fd=8))                  
tcp   ESTAB    0      0           172.31.32.243:58474 44.230.203.236:6443  users:(("kube-burner",pid=393077,fd=7))                  
tcp   LISTEN   0      4096                [::1]:38709           [::]:*     users:(("kube-burner",pid=393077,fd=9))    

Two solutions we can work on
1) parsePort() should ignore ipv6 data while parsing the output
2) or use only IPV4 for client-go portforward code

venkataanil:
On some environments, client port forwarding can use both ipv4 and ipv6 addresses for port forwarding. kube-burner has to consider this scenario as well while parsing the port forwarding output and fetch the local port.

- Closes #749

### DIFF
--- a/pkg/measurements/netpol_latency.go
+++ b/pkg/measurements/netpol_latency.go
@@ -54,7 +54,7 @@ const (
 	netpolLatencyQuantilesMeasurement = "netpolLatencyQuantilesMeasurement"
 )
 
-var proxyPortForwarder util.PodPortForwarder
+var proxyPortForwarder *util.PodPortForwarder
 var nsPodAddresses = make(map[string]map[string][]string)
 var proxyEndpoint string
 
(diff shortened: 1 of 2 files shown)


mohit-sheth (approved):
lgtm

venkataanil:
My intenrtion was to use a random port for local port. But let me try using the  networkPolicyProxyPort (9002) for the local port as well. Thaks for the suggestion.

venkataanil:
> My intenrtion was to use a random port for local port. But let me try using the networkPolicyProxyPort (9002) for the local port as well. Thaks for the suggestion.

Pushed a new commit with the suggested change.

pkg/measurements/netpol_latency.go
```diff
@@ -488,13 +488,13 @@ func (n *netpolLatency) start(measurementWg *sync.WaitGroup) error {
 			return err
 		}
 	}
-	if proxyPortForwarder.LocalPort == "" {
+	if proxyPortForwarder == nil {
```
rsevilla87, line 491: In which cases this conditional can be true?

venkataanil: This will be helpful when the test has multiple jobs.

venkataanil: The intention is to create port forwarder only once for the kube-burner run and reuse the same in all the jobs of the kube-burner run.

rsevilla87: ok, makes sense

rsevilla87 (approved):
lgtm

PR #753, opened 2024-12-09
Template - add ReadFile to template map

ygalblum:
## Type of change

- [x] New feature

## Description

Allow reading data from a file into the rendered template

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -368,7 +368,19 @@ spec:
 
 ## Template functions
 
-In addition to the default [golang template semantics](https://golang.org/pkg/text/template/), kube-burner is compiled with the [sprig library](http://masterminds.github.io/sprig/), which adds over 70 template functions for Go’s template language.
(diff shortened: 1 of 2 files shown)


rsevilla87 (changes requested):
Can you update the docs with this new function?

https://kube-burner.github.io/kube-burner/latest/reference/configuration/#template-functions

ygalblum:
@rsevilla87 I've updated the documentation. Since none of the functions that were added manually were documented, I changed the section.
Not sure I understand what `GetSubnet24` receives or does. So, either let me know and I'll update this PR or just fix it later

rsevilla87:
> @rsevilla87 I've updated the documentation. Since none of the functions that were added manually were documented, I changed the section. Not sure I understand what `GetSubnet24` receives or does. So, either let me know and I'll update this PR or just fix it later

Thanks for the updates, some of those functions are old and probably need to be removed, we can address that in the future, I don't want to delay this PR more time

rsevilla87 (approved):
lgtm

PR #755, opened 2024-12-09
Use -c when running the beforeCleanup shell command

ygalblum:
## Type of change

- [x] Bug fix

## Description

In order to run a shell script using `/bin/sh` the `-c` flag should be used


ygalblum:
I'm closing this PR because I have a broader fix that allows running scripts also from the embedded filesystem. The fix depends on other PRs, so I will open a PR for it once its dependencies are merged

PR #757, opened 2024-12-10
Preload vm images

rsevilla87:
## Type of change

- [x] New feature

## Description

Preload Virtual machines images folowing the same mechanism we already do with pods

cc @jeniferh @ygalblum @jtaleric

### DIFF
--- a/pkg/burner/pre_load.go
+++ b/pkg/burner/pre_load.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/kube-burner/kube-burner/pkg/util"
 	log "github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v3"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


pkg/burner/pre_load.go
```diff
@@ -42,6 +43,25 @@ type NestedPod struct {
 	} `json:"spec"`
 }
 
+type VMorVMI struct {
```
ygalblum, line 46: VM and VMI do not have the same structure. The path `.spec.template.spec` of the VirtualMachine
 and VirtualMachineInstanceReplicaSet resources equals to the `.spec` of the VirtualMachineInstance

ygalblum (changes requested):

pkg/burner/pre_load.go
rsevilla87: correct, but Im reusing the same struct to unmarshal all of the those kinds, this struct defines .spec.volumes (valid for VMIs) and .spec.template.spec.volumes (valid for VMs and VMIRS)

ygalblum: Since the handling is also separate, why not create separate types (and cases in the switch)? What's the advantage of combining it to the same struct?

rsevilla87: fixed

ygalblum (approved):
LGTM

josecastillolema (approved):

PR #761, opened 2024-12-16
Consolidate file read to a single function

ygalblum:
## Type of change

- [x] Refactor
- [x] Bug fix

## Description
Create a single GetReader function for all sources
Adjust all Reader operations to use the new function
Adjust APIs where needed
Remove the redundant embedConfig flag from Prometheus and Alert Manager
Make the EmbedFS field a pointer to allow simple check

When using via `kube-burner-ocp` patch jobType could not read from the embedded file system.


rsevilla87 (approved):
Thanks!, good addition!

PR #764, opened 2024-12-31
Support running shell scripts from the embedded filesystem

ygalblum:
## Type of change

- [x] Bug fix

## Description

Instead of calling `/bin/sh` on the script file, read the content of the file and run it from stdin. 
Note, this works only when all the script code is in a single file

Fail to execute shell scripts when using embedded file system

### DIFF
--- a/pkg/util/bash.go
+++ b/pkg/util/bash.go
@@ -1,0 +1,54 @@
+// Copyright 2024 The Kube-burner Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


pkg/util/bash.go
```diff
@@ -0,0 +1,54 @@
...
+	var c []string
+	if len(parts) > 1 {
+		c = append([]string{"-s", "-"}, parts[1:]...)
+	}
+
+	// Create a command
+	cmd := exec.Command("/bin/sh", c...)
+
+	// Run the shell script from STDIN
+	cmd.Stdin = shellScriptReader
```
rsevilla87, line 44: Very cool!

rsevilla87:
@ygalblum I assume your idea is to have the embed the shell scripts as well

ygalblum:
> @ygalblum I assume your idea is to embed the shell scripts together with the workload, isn't it?

Yes. I'm working on something in `kube-burner-ocp` with a verification script that requires this change

PR #765, opened 2024-12-31
Template - pass the job run id to the template

ygalblum:
## Type of change

- [x] New feature

## Description

Add the RunID to the template param to allow marking sub-resources for metrics


rsevilla87 (approved):
lgtm

PR #788, opened 2025-01-08
Verify - limit the objects list to the ones created in the current run

ygalblum:
## Type of change

- [x] Optimization

## Description
To allow running the same test sequence with the same UUID, generate the RUNID every time the configuration spec is loaded and verify objects that were created under the specific runid

### DIFF
--- a/pkg/burner/utils.go
+++ b/pkg/burner/utils.go
@@ -92,7 +92,7 @@ func (ex *Executor) Verify() bool {
 	log.Info("Verifying created objects")
 	for objectIndex, obj := range ex.objects {
 		listOptions := metav1.ListOptions{
-			LabelSelector: fmt.Sprintf("kube-burner-uuid=%s,kube-burner-job=%s,kube-burner-index=%d", ex.uuid, ex.Name, objectIndex),
+			LabelSelector: fmt.Sprintf("kube-burner-uuid=%s,kube-burner-runid=%s,kube-burner-job=%s,kube-burner-index=%d", ex.uuid, ex.runid, ex.Name, objectIndex),
 			Limit:         objectLimit,
 		}
 		err := util.RetryWithExponentialBackOff(func() (done bool, err error) {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -36,7 +36,6 @@ import (
 
 var configSpec = Spec{
 	GlobalConfig: GlobalConfig{
-		RUNID:            uid.NewString(),
 		GC:               false,
 		GCMetrics:        false,
 		GCTimeout:        1 * time.Hour,
@@ -181,6 +180,7 @@ func ParseWithUserdata(uuid string, timeout time.Duration, configFileReader, use
 	}
 	configSpec.GlobalConfig.Timeout = timeout
 	configSpec.GlobalConfig.UUID = uuid
+	configSpec.GlobalConfig.RUNID = uid.NewString()
 	return configSpec, nil
 }
 


pkg/burner/utils.go
```diff
@@ -92,7 +92,7 @@ func (ex *Executor) Verify() bool {
 	log.Info("Verifying created objects")
 	for objectIndex, obj := range ex.objects {
 		listOptions := metav1.ListOptions{
-			LabelSelector: fmt.Sprintf("kube-burner-uuid=%s,kube-burner-job=%s,kube-burner-index=%d", ex.uuid, ex.Name, objectIndex),
+			LabelSelector: fmt.Sprintf("kube-burner-uuid=%s,kube-burner-runid=%s,kube-burner-job=%s,kube-burner-index=%d", ex.uuid, ex.runid, ex.Name, objectIndex),
```
vishnuchalla, line 95: I am okay with adding `runid` here, but I do not understand the intention behind [exposing runid](https://github.com/kube-burner/kube-burner/pull/765/files) to the end user. Because it was meant to be an internal thing which shouldn't be provided by a user (i.e for user we already have `uuid`). The idea around this is to be able to uniquely identify run specific resources even after user provides a `uuid` which clashes with existing resources of a cluster by mistake. But having both the IDs exposed is kind of confusing and doesn't justify the value add.

ygalblum: First, for this PR, it is needed for the same reason you gave here. If the same test is executed with the same UUID without a cleanup the verification fails because the request finds the resources created during the previous execution.

As for the other PR, the need to expose the RunID to the user was to allow users to mark sub-resources for the metrics collection. I'll explain. I have a followup PR that adds support for `DataVolume` latency. DVs can either be created explicitly (in which case `kube-burner` marks them automatically) or implicitly, for example by using the `DataVolumeTemplates` array in the `VirtualMachine` spec. In this case, `kube-burner` does not mark the resources and as a result the metrics collection disregards it. Now, one way is to make `kube-burner` mark also the volumes. But, this means that every time such a need arises, a new code changes will be required. Instead, exposing the `RunID` solves the issue in one go.

vishnuchalla: okay. The idea seems reasonable. In your follow up PR, can we get the parameter that overrides `RUNID` renamed something as `--override-runid` or something and documented as well? So that users will not run into the same confusion as me.

ygalblum: I don't understand what you mean by `the parameter that overrides RUNID`. Which parameter are you referring to?

vishnuchalla: Sorry for not being clear enough earlier. I was referring to [this](https://github.com/kube-burner/kube-burner/pull/765/files#diff-5812cf0d04ccaf9334dcc4dde74c0fa3d3e17af5b915531afdb71bc6b837138cR407). At present, description says its internal at the same time user has an option to specify it through the job yaml as an input, which creates a confusion. Instead, I was suggesting to unexpose it from job configuration yaml and manage it through a CLI param that goes by `--override-runid` flag. That way a user understands what `runid` is all about and is sure of what they are doing.

ygalblum: But I don't think that the user should override the value of `runid` just be able to put it in the object labels. Also, let's move the discussion to a new issue as it is not really related to this PR

vishnuchalla (approved):
lgtm

pkg/burner/utils.go
vishnuchalla: ohh I see. I misread the code earlier. Yes, its not technically exposing the field at the moment. All good now.

PR #793, opened 2025-01-16
CI - Run on latest patch version of the latest 3 minor version

ygalblum:
## Type of change

- [x] CI Optimization

## Description

Fetch the latest patch version of the latest 3 minor versions of the kindest/node image and use them for testing


vishnuchalla (approved):
lgtm

PR #795, opened 2025-01-17
Allow passing additional variables when using as a package

ygalblum:
## Type of change

- [x] Optimization

## Description

When using kube-burner as a package passing parameters as environment variables may be cumbersome. Instead, allow passing the values via a method parameter


vishnuchalla (approved):
lgtm. Thanks for this change

PR #807, opened 2025-01-30
Allow waiting for status between objects in the same job iteration

ygalblum:
## Type of change

- New feature
- Bug fix

## Description

Add a new job parameter to wait between objects
In Sequential run check and wait at the object and at the job iteration level
Adjust testing
Add to doc


vishnuchalla (approved):
lgtm

PR #808, opened 2025-02-05
Kubevirt operations - add waiter

ygalblum:
## Type of change

- New feature

## Description

Implement predefined wait conditions for the completion of kubevirt operations so that users do not have to do explicitly define them


docs/reference/configuration.md
```diff
@@ -373,6 +373,22 @@ Additional parameters should be set using the `inputVars` field:
 - `volumeName` - Name of the volume to remove. Mandatory
 - `persist` - if set, the added volume will be persisted in the VM spec (if it exists). Default `false`
 
+#### Wait for completion
+
+Wait is supported for the following operations:
+
+- `start` - Wait for the `Ready` state of the `VirtualMachine`  to become `True`
+- `stop` - Wait for the  `Ready` state of the `VirtualMachine` state to become `False`
```
vishnuchalla, line 381: Looks like this isn't a desired state when we stop a VM? Also can you cross check if are any other errors in the docs?

ygalblum: First, the doc aligns with the implementation. Second, unfortunately, there is no defined condition that states that a VM is stopped. Theoretically, I could use the `printableStatus` field. However, using it is discouraged by KubeVirt.

I am pushing KubeVirt to provide better and reliable information. But, currently, this the best we can do

ygalblum: I did some more digging and found that it's better to add another check on the `reason` of the status. So, I added it to the conditions list.
I've also generalized the condition check utility and moved it to a separate file

vishnuchalla (approved):
lgtm

PR #809, opened 2025-02-06
Waiters - do not fail if query jq failed

ygalblum:
## Type of change

- Bug fix

## Description

Fix race condition in which the waiter checks for a field in the status that does not yet exist.
For example, waiting on a condition before there are any


vishnuchalla (approved):
lgtm

PR #811, opened 2025-02-14
Waiter - reduce verbosity on successful checks

ygalblum:
## Type of change

- Optimization

## Description

Since all items are fetched on each try, the log becomes too verbose. Change success to debug message


vishnuchalla (approved):
lgtm

PR #812, opened 2025-02-19
Update go-commons dependency to v2

ygalblum:
## Type of change

- Optimization

## Description

Update the go-common dependency to v2. This is required to avoid version dependency conflicts.


vishnuchalla (approved):
lgtm

Issue #433, opened 2023-08-14
[RFE] Ship grafana dashboards used to represent indexed with `--reporting` mode


smalleni:
**Is your feature request related to a problem? Please describe.**
It would be good to ship the Grafana dashboards we use for visualizing metric data shipped to ES.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

rsevilla87:
Agree, but, maybe we should consider using a different repo to avoid having an all-in-one repo?, not sure though, having the dashboards here would also bound them to an specific kube-burner version.

vishnuchalla:
Now that we also opensearch integration in CI, I think we can setup a make target to ship grafana with some example dashboards using ES and prometheus as data sources.

PR #817, opened 2025-03-03
Ship grafana dashboard in CI

vishnuchalla:
## Type of change
- New feature
- CI

## Description
Adding a option to ship grafana dashboards in our CI.

Closes # https://github.com/kube-burner/kube-burner/issues/433

## Testing
Tested in local using `DEPLOY_GRAFANA=true make test-k8s`
Screenshot From 2025-03-03 15-18-47

### DIFF
--- a/test/test-k8s.bats
+++ b/test/test-k8s.bats
@@ -15,20 +15,30 @@ setup_file() {
   export CHURN_CYCLES=100
   export TEST_KUBECONFIG; TEST_KUBECONFIG=$(mktemp -d)/kubeconfig
   export TEST_KUBECONTEXT=test-context
+  export ES_SERVER=${PERFSCALE_PROD_ES_SERVER:-"http://localhost:9200"}
+  export ES_INDEX="kube-burner"
+  export DEPLOY_GRAFANA=${DEPLOY_GRAFANA:-false}
   if [[ "${USE_EXISTING_CLUSTER,,}" != "yes" ]]; then
     setup-kind
   fi
   create_test_kubeconfig
   setup-prometheus
   if [[ -z "$PERFSCALE_PROD_ES_SERVER" ]]; then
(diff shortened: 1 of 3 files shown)


smalleni:
Nice, I like the fact that we can even test our dashboards locally with this. Currently the way it is setup, it is always going to test the one dashboard in `tests/grafana`. Is the idea that we would move all our kube-burner based grafana dashboards also in tree and test them in CI?

vishnuchalla:
> Is the idea that we would move all our kube-burner based grafana dashboards also in tree

Yes correct. The more we add, the easier it becomes for us to demo, test while development and customer usage perspective as well.

>  and test them in CI?

It's not exactly about on boarding this to CI (i.e. for which we have other better options), but more about providing a quick flexibility on testing and verifying the entire workflow.

test/test-k8s.bats
```diff
@@ -15,20 +15,30 @@ setup_file() {
...
   fi
   create_test_kubeconfig
   setup-prometheus
   if [[ -z "$PERFSCALE_PROD_ES_SERVER" ]]; then
+    $OCI_BIN rm -f opensearch
+    $OCI_BIN network rm -f monitoring
+    setup-shared-network
     setup-opensearch
+    if [ "$DEPLOY_GRAFANA" = "true" ]; then
+      $OCI_BIN rm -f grafana
```
ygalblum, line 32: Sorry, I missed this. You need to run this before deleting the network. Since you're using `-f` it will noop in case the container does not exist

vishnuchalla: Actually order doesn't matter here. Script will always work because we do not care about broken containers that are only created by this script in a specific way.

ygalblum (approved):
LGTM

smalleni (approved):

PR #819, opened 2025-03-04
VMI Latency - watch only for virt-launcher pods

ygalblum:
## Type of change

- Bug fix

## Description

Filter the pod watcher for vmi latency to those marked as virt-launcher to avoid warning of unrelated pods


vishnuchalla (approved):
lgtm

PR #822, opened 2025-03-07
Expect items from a single iteration when RunOnce is set

### LINKED ISSUE #821: [BUG] Creation verification fails when runOnce is true and jobIterations is bigger than 1
# Bug Description

## **Describe the bug**

When `runOnce` is set at the object level, create will process it only once. As a result, the number of items will be the number of replicas set in the object. The verify method however multiplies the expected number by the number of `jobIteration`.
As a result, when `runOnce` is set and `jobIteration` is larger than 1 the verification fails.

## **To Reproduce**

Add a create job with `jobIterations` larger than 1 and include an object with `runOnce` set to `true`.
For example, set `jobIterations` to 2 and an object to create a secret with `replicas` set to 1.

## **Expected behavior**

Only one item is created and the verification passes

## **Screenshots or output**

Only one item is created but the verification fails:
```
time="2025-03-07 11:07:26" level=error msg="secrets found: 1 Expected: 2" file="utils.go:149"
```

ygalblum:
## Type of change

- Bug fix

## Description

When `runOnce` is set the number of created items is the number is in `replicas` without multiplying it be the number of iterations.

- Closes #821

### DIFF
--- a/pkg/burner/utils.go
+++ b/pkg/burner/utils.go
@@ -144,7 +144,12 @@ func (ex *Executor) Verify() bool {
 			success = false
 			continue
 		}
-		objectsExpected := ex.JobIterations * obj.Replicas
+		var objectsExpected int
+		if obj.RunOnce {
+			objectsExpected = obj.Replicas
+		} else {
+			objectsExpected = obj.Replicas * ex.JobIterations
+		}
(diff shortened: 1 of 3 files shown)


vishnuchalla (approved):
lgtm

PR #823, opened 2025-03-10
DV Latency - reduce log verbosity

ygalblum:
## Type of change

- Optimization

## Description

DV Latency - reduce log verbosity of ready to debug


vishnuchalla (approved):
lgtm

PR #824, opened 2025-03-11
Allow changing the content of the objects stored in the objects array

ygalblum:
## Type of change

- Optimization

## Description

Change objects array to hold pointers and pass by pointer
Remove duplicate kind field in object struct
Fix data volume latency test


vishnuchalla (approved):
lgtm

PR #831, opened 2025-03-19
Metrics - allow aggregating metrics from consecutive jobs

ygalblum:
## Type of change

- Optimization

## Description

Allow aggregating metrics from consecutive jobs so that metrics accumulation can span over multiple jobs


vishnuchalla (approved):
lgtm

PR #837, opened 2025-04-03
Delete job - always set waitWhenFinished to false

ygalblum:
## Type of change

- Bug fix
- Documentation

## Description

waitWhenFinished searches for resources while a delete job removes them. As a result, waitWhenFinished should never be used. Instead WaitForDeletion should be used


vishnuchalla (approved):
lgtm

PR #845, opened 2025-04-09
Prevent aborting execution on garbage collection timeout

### LINKED ISSUE #842: [BUG] If cleanup times out, indexing of all results does not occur.
# Bug Description
When kube-burner's cleanup fails, indexing of results gets bypassed

Starting kube-burner (1.6.3@41f0a2c8d0c05a68729f77f577f01bf6c84d77f2)

## **Describe the bug**

When kube-burner fails to cleanup, we see a message like : 

```
time="2025-04-07 21:07:13" level=info msg="Indexing collected data from measurement: pprof" file="factory.go:151"
time="2025-04-07 22:07:12" level=fatal msg="Timeout cleaning up namespaces: context deadline exceeded" file="namespaces.go:94"
```

and all the measurement data isn't indexed locally or to OpenSearch -- even though the benchmark completed. 

## **To Reproduce**

Cause cleanup not to succeed.

## **Expected behavior**

Cleanup or any other failure happens, we should have a best-effort on indexing results.

rsevilla87:
## Type of change

- Bug fix

## Description

With this new logic, when garbage collection times out, execution is not aborted (so indexing won't be interrupted) and kube-burner's rc will be 2, as per documented rc for timeout cases.

- Related Issue #
- Closes #842

### DIFF
--- a/pkg/burner/job.go
+++ b/pkg/burner/job.go
@@ -250,14 +250,16 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider, 
 		indexMetrics(uuid, executedJobs, returnMap, metricsScraper, configSpec, false, utilerrors.NewAggregate(errs).Error(), true)
 	}
 	// When GC is enabled and GCMetrics is disabled, we assume previous GC operation ran in background, so we have to ensure there's no garbage left
-	if globalConfig.GC && !globalConfig.GCMetrics {
-		log.Info("Garbage collecting jobs")
-		gcWg.Wait()
-	}
-	// When GC times out and job execution has finished successfully, return timeout
-	if gcCtx.Err() == context.DeadlineExceeded && rc == 0 {
-		errs = append(errs, fmt.Errorf("garbage collection timeout reached"))
-		rc = rcTimeout
+	if globalConfig.GC {
+		if !globalConfig.GCMetrics {
+			log.Info("Garbage collecting jobs")
+			gcWg.Wait()
+		}
+		// When GC times out and job execution has finished successfully, return timeout
+		if gcCtx.Err() == context.DeadlineExceeded && rc == 0 {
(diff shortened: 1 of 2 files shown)


pkg/burner/job.go
```diff
@@ -72,6 +72,8 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
 	var executedJobs []prometheus.Job
 	var jobList []Executor
 	var msWg, gcWg sync.WaitGroup
+	var gcCtx context.Context
```
ygalblum, line 75: Why do you need to define these variables?

rsevilla87: They are used in line 243 and need to be predefined in order to evaluate them outside the conditional

ygalblum: Right, I missed it below

pkg/burner/job.go
```diff
@@ -230,24 +232,34 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
...
+			gcCtx, cancelGC = context.WithTimeout(context.Background(), globalConfig.GCTimeout)
+			defer cancelGC()
+			for _, job := range jobList[:len(executedJobs)-1] {
+				gcWg.Add(1)
+				go garbageCollectJob(gcCtx, job, fmt.Sprintf("kube-burner-job=%s", job.Name), &gcWg)
+			}
+		}
 		indexMetrics(uuid, executedJobs, returnMap, metricsScraper, configSpec, false, utilerrors.NewAggregate(errs).Error(), true)
 	}
 	// When GC is enabled and GCMetrics is disabled, we assume previous GC operation ran in background, so we have to ensure there's no garbage left
```
ygalblum, line 252: This comment should now go inside the if

ygalblum (commented):
Small nit

ygalblum (approved):
LGTM

PR #848, opened 2025-04-14
DataVolume Latency - disregard true conditions at creation timestamp

ygalblum:
## Type of change

- Bug fix

## Description

Workaround an known issue in CDI causing a DV to show as ready before it actually is


vishnuchalla (approved):
lgtm

Issue #846, opened 2025-04-09
[RFE] Optinal inclusion of jobPause in kube-burner metrics


vishnuchalla:
**Is your feature request related to a problem? Please describe.**
If we have a workload that has multiple jobs and at the end while capturing metrics we do not want to include the numbers related to `jobPause` stage as it doesn't make much sense in some of the usecases.

**Describe the solution you'd like**
Simple solution. Make a note of every `jobPause` stage (start, end) time pairs and exclude them while performing aggregations at the end for measurements.

**Describe alternatives you've considered**
Also consider capturing measurements in chunks as an alternative and excluding every `jobPause` phase to avoid unnecessary data in such cases.

**Additional context**
This should help us in avoiding noise in our metrics in almost all our workloads.

cc: @jtaleric @mohit-sheth @venkataanil

PR #852, opened 2025-05-05
MetricsCollectionEndpoint config to control metrics end timing

venkataanil:
This introduces a new job configuration option `MetricsCollectionEndpoint` to define when the metrics collection should stop. The option supports three values:

- AfterJob:        collect metrics after the job completes
- AfterJobPause:   collect metrics after the jobPause duration ends
- AfterMeasurements: collect metrics after all measurements are finished

This gives users fine-grained control over when to end metrics collection, based on their workload requirements:

1. Users who want to monitor idle periods post resource creation can use AfterJobPause to include the pause in metrics.
2. UDN jobs may want metrics collected immediately after the job, before the pause begins.
3. BGP workloads, which perform route import/export during measurements, benefit from collecting metrics until AfterMeasurements completes.


docs/reference/configuration.md
```diff
@@ -525,3 +526,11 @@ jobs:
...
     replicas: 10
 ```
+
+## MetricsCollectionEndpoint
+
+This config defines when the metrics collection should stop. The option supports three values:
+
+- `AfterJob` - collect metrics after the job completes
+- `AfterJobPause` - collect metrics after the jobPause duration ends
+- `AfterMeasurements` - collect metrics after all measurements are finished
```
rsevilla87, line 536: We use camelCase notation for configuration parameters

```suggestion
- `afterJob` - collect metrics after the job completes
- `afterJobPause` - collect metrics after the jobPause duration ends
- `afterMeasurements` - collect metrics after all measurements are finished
```

pkg/config/config.go
```diff
@@ -84,23 +84,24 @@ func (o *Object) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+		WaitForDeletion:           true,
+		PreLoadImages:             true,
+		PreLoadPeriod:             1 * time.Minute,
+		Churn:                     false,
+		ChurnCycles:               100,
+		ChurnPercent:              10,
+		ChurnDuration:             1 * time.Hour,
+		ChurnDelay:                5 * time.Minute,
+		ChurnDeletionStrategy:     "default",
+		MetricsCollectionEndpoint: "AfterJob",
```
rsevilla87, line 104: Default should be `afterPause` to keep current functionality, but thinking it twice, it makes more sense to me to skip pause stage by default. I think there's not any critical workload using jobPause at the moment, so Im good with this `afterJob` value, anyway, I leave this message here in case you find out any risk.

pkg/config/types.go
```diff
@@ -205,6 +205,8 @@ type Job struct {
 	ObjectWait bool `yaml:"objectWait" json:"objectWait,omitempty"`
 	// MetricsAggregate aggregate the metrics of this job with the next one
 	MetricsAggregate bool `yaml:"metricsAggregate" json:"metricsAggregate,omitempty"`
+	// MetricsCollectionEndpoint defines when to stop metrics collection
+	MetricsCollectionEndpoint string `yaml:"metricsCollectionEndpoint" json:"metricsCollectionEndpoint,omitempty"`
```
rsevilla87, line 209: This variable name is very confusing, endpoint in IT normally refers to an URL, let's use something like `collectMetrics`, which fits with the possible values nicely.

rsevilla87 (changes requested):

rsevilla87 (commented):
Can we also validate the value configured `metricsCollectionEndpoint`  is correct?, config validations are usually performed in  https://github.com/kube-burner/kube-burner/blob/0e60a622a5c71a404d32e9ba071d889aa7da8cb1/pkg/config/config.go#L152

pkg/config/types.go
vishnuchalla: I am not an expert in naming either. But something that aligns with the definition would make better sense. Maybe something like a `metricsStopper/metricsPeriod/metricsCease/metricsClosing/metricsFinish`?

vishnuchalla:
I think this change should close: https://github.com/kube-burner/kube-burner/issues/846

rsevilla87:
> I think this change should close: #846

yeah, makes sense to me

docs/reference/configuration.md
venkataanil: my bad

pkg/config/config.go
venkataanil: I will keep "afterJob".

pkg/config/types.go
venkataanil: shall I go with metricsClosing?

rsevilla87: ok, sgtm

docs/reference/configuration.md
```diff
@@ -129,6 +129,7 @@ This section contains the list of jobs `kube-burner` will execute. Each job can
 | `objectDelay`                | How long to wait between each object in a job                                                                                         | Duration | 0s       |
 | `objectWait`                 | Wait for each object to complete before processing the next one - not for Create jobs                                                 | Boolean  | 0s       |
 | `metricsAggregate`           | Aggregate the metrics collected for this job with those of the next one                                                               | Boolean  | false    |
+| `metricsClosing`  | To define when the metrics collection should stop. More details at [MetricsClosing](#MetricsClosing)                       | String   | "afterJob"       |
```
rsevilla87, line 132: Last nit, can you align this line with the previous ones?

PR #855, opened 2025-05-06
Fix churnEnd calculation

### LINKED ISSUE #854: [BUG] churn Start and End timestamps are the same
# Bug Description

Latest versions.

## **Describe the bug**

jobSummary document has two timestamps that are being set to the same value: `churnStartTimestamp` and `churnEndTimestamp`.

Looks like because we are only calling Now() once.
A consequence of over-refactoring.

## **Expected behavior**

The two timestamps should be different.

## **Screenshots or output**

Image

rsevilla87:
## Type of change

- Bug fix

## Description

Fixing the expression

- Related Issue #
- Closes #854


afcollins (approved):
LGTM

Issue #647, opened 2024-05-31
[RFE] Support to optionally add watches during the testing


josecastillolema:
**Is your feature request related to a problem? Please describe.**
Kubernetes supports efficient change notifications on resources via watches. Several k8s applications leverage them, i.e.: ArgoCD, Knative, etc. Watches are known to add load to KAPI in some scenarios.

**Describe the solution you'd like**
While kube-burner creates a watch for all the pods that are created during the testing, it would be nice to be able to optionally create additional watches for a number of resources (i.e.: namespaces with certain labels).

cc @williamcaban @smalleni @Noreen21

rsevilla87:
Hey @josecastillolema , can you provide an example of what exactly are you going to get with this feature? is this related to create load in the API by adding watchers to some subset of objects?

josecastillolema:
Exactly, @williamcaban was mentioning that some watchers are known to cause a severe overhead on KAPI resource consumption.

williamcaban:
For more context, two actions are known to stress the K8s API: WATCH and LIST. In both cases, any K8s with a version < 1.30 (where we should see some fixes with API streaming). In the case of many LIST requests (which is what an application like Argo, or an audit tool does), it forces the API server to allocate enough memory to build the structure in memory to represent all the objects requested by the LIST action, which can be equivalent to all objects in etcd when is a LIST of all the objects in the cluster, to generate the answer to that request. Because of this, when a cluster has multiple applications or agents with regular large LIST requests, it tries to allocate all the available memory to the point where the control plane node resources are exhausted. There are features like APF that can indirectly help. The intent is to be able to replicate these unfavorable conditions to determine scalability, recoverability, and stability, which we can maintain in the control plane under stress.

In the case of WATCH, it directly impacts the scalability of how many Operators or custom controllers the system can handle. The intent of generating a WATCH load condition is to measure operator scalability, and stability under stressed conditions, etc. 

At the end, having both of these, WATCH and LIST, would also allow us to replicate conditions to the control plane that otherwise will require instantiating large number of workloads requiring large number of worker nodes (e.g., 8k Tekton pipelines, or 100 instances of Argo) when the fact is we need to generate the WATCH and LIST load, not to install the applications and create applications trying to achieve the same goal.

PR #866, opened 2025-05-21
Adds option to run a script in background before workload execution

vishnuchalla:
## Type of change
- Refactor
- New feature
- Optimization
- Documentation

## Description
Adds support to run a daemon script in the background, which provides us with the flexibility to act as an observer while the workload is running.

- Closes # https://github.com/kube-burner/kube-burner/issues/647

## Testing
Tested integrating with PR: https://github.com/kube-burner/kube-burner/pull/865


pkg/burner/job.go
```diff
@@ -274,6 +269,28 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
...
+func runLifecycleHook(name string, script string, background bool, configSpec config.Spec, errs *[]error, rc *int) {
+	if script == "" {
+		return
+	}
+
+	logMsg := fmt.Sprintf("Running %s command %s", name, script)
+	if !background {
+		logMsg = fmt.Sprintf("Waiting for %s command %s to finish", name, script)
+	}
+	log.Info(logMsg)
```
ygalblum, line 282: The extra variable does not add much here. If anything it adds complexity to the runtime
```suggestion
	if background {
	    log.Info("Running %s command %s", name, script)
	else {
		log.Info("Waiting for %s command %s to finish", name, script)
	}
```

pkg/burner/job.go
```diff
@@ -274,6 +269,28 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
...
+
+	stdOut, stdErr, err := util.RunShellCmd(script, configSpec.EmbedFS, configSpec.EmbedFSDir, background)
+	if err != nil {
+		err = fmt.Errorf("%s failed: %v", name, err)
+		log.Error(err.Error())
+		*errs = append(*errs, err)
+		*rc = 1
+	}
+	log.Infof("%s out: %v, err: %v", name, stdOut.String(), stdErr.String())
```
ygalblum, line 291: Is this print relevant in case of `background`?

pkg/util/bash.go
```diff
@@ -48,6 +49,15 @@ func RunShellCmd(shellCmdLine string, configEmbedFS *embed.FS, configEmbedFSDir
 	cmd.Stdout = &outb
 	cmd.Stderr = &errb
 
+	if background {
+		// Fully detach from parent so the child keeps running if Go process exits
```
ygalblum, line 53: Do we really want to allow escaping outside of the process? What if the script just runs forever?

josecastillolema:
Nice addition @vishnuchalla !
Are there some docs on how to use this new feature?

cc @sferlin this could become handy for dp+cp orquestration

pkg/util/bash.go
vishnuchalla: Yes we want it to be running in a detached mode. It's up to user to make sure that it is killed if required.

pkg/burner/job.go
vishnuchalla: Removed it as suggested. User can always look at log files to figure things out if running workloads in parallel.

pkg/util/bash.go
ygalblum: ok

ygalblum:
The linter is complaining. I think it's my suggestion, it should be `log.Infof` instead of `log.Info`

docs/reference/configuration.md
```diff
@@ -106,6 +106,7 @@ This section contains the list of jobs `kube-burner` will execute. Each job can
 | `maxWaitTimeout`             | Maximum wait timeout per namespace                                                                                                    | Duration | 4h       |
 | `jobIterationDelay`          | How long to wait between each job iteration. This is also the wait interval between each delete operation                             | Duration | 0s       |
 | `jobPause`                   | How long to pause after finishing the job                                                                                             | Duration | 0s       |
+| `beforeExecution`              | Allows to run a bash script in the background before the workload starts                                                                            | String   | ""       |
```
rsevilla87, line 109: Please align this line the the surroundings

docs/reference/configuration.md
```diff
@@ -106,6 +106,7 @@ This section contains the list of jobs `kube-burner` will execute. Each job can
...
 | `beforeCleanup`              | Allows to run a bash script before the workload is deleted                                                                            | String   | ""       |
```
rsevilla87, line 110:
```suggestion
| `beforeCleanup`              | Allows to run a bash script in foreground before the job is deleted                                                                            | String   | ""       |
```

docs/reference/configuration.md
vishnuchalla: Aligned according to surroundings.

docs/reference/configuration.md
vishnuchalla: Updated as suggested.

pkg/burner/job.go
```diff
@@ -274,6 +269,27 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
 	return rc, utilerrors.NewAggregate(errs)
 }
 
+// runLifecycleHook runs a user-defined shell script for a specific lifecycle hook
+func runLifecycleHook(name string, script string, background bool, configSpec config.Spec, errs *[]error, rc *int) {
```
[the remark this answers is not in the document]
vishnuchalla: `background` is just to toogle execution in between foreground and background. You should be able to see two fucntion calls in this code which show the difference.

vishnuchalla:
> Overall code looks good to me except some nits, what I don't understand its relation with #865 and how does running a script in foreground helps to that feature.

Now with the new idea that you have proposed in #865 it might not be. But still this feature has its own value in terms of usage.

pkg/burner/job.go
```diff
@@ -275,6 +270,27 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
...
+		return
+	}
+
+	if background {
+		log.Infof("Running %s command %s", name, script)
+	} else {
+		log.Infof("Waiting for %s command %s to finish", name, script)
+	}
+
+	_, _, err := util.RunShellCmd(script, embedCfg, background)
```
ygalblum, line 285: You should still print stderr in case of error

ygalblum:
> But still this feature has its own value in terms of usage.

What's the point in this feature now? Do you have an example?

pkg/burner/job.go
vishnuchalla: I am not sure what is being referred here, isn't it already being printed in the following lines? CMIIW.

vishnuchalla:
> > But still this feature has its own value in terms of usage.
> 
> What's the point in this feature now? Do you have an example?

I have provided an usage example in the description on how to run things in the background. I do not have a use-case for myself handy at the moment but I think this feature will be useful for some of the users who are trying to run some control plane + data plane workloads in parallel.

ygalblum:
This change was added to support the ability to run a script while setting watchers. But, @rsevilla87 provided a better design that was chosen and being implemented. As a result, I don't see the need for this addition and it doesn't seem like we have a different example. 
Code that is not used becomes stale, so we need to make sure we don't add code that is not needed.

pkg/burner/job.go
ygalblum: The call to `RunShellCmd` that was removed, used to print stdout and strerr in case of error. But, here you disregard the returned values. As a result, all you will print is the return value of the script and not its output

vishnuchalla:
> This change was added to support the ability to run a script while setting watchers. But, @rsevilla87 provided a better design that was chosen and being implemented. As a result, I don't see the need for this addition and it doesn't seem like we have a different example. Code that is not used becomes stale, so we need to make sure we don't add code that is not needed.

Thanks for putting down your thoughts @ygalblum . @rsevilla87 considering different groups of users that you have seen (.i.e probably more than me and Ygal), let me know if you feel the same about this. I am okay with discarding this PR if that is the case.

rsevilla87:
> > This change was added to support the ability to run a script while setting watchers. But, @rsevilla87 provided a better design that was chosen and being implemented. As a result, I don't see the need for this addition and it doesn't seem like we have a different example. Code that is not used becomes stale, so we need to make sure we don't add code that is not needed.
> 
> Thanks for putting down your thoughts @ygalblum . @rsevilla87 considering different groups of users that you have seen (.i.e probably more than me and Ygal), let me know if you feel the same about this. I am okay with discarding this PR if that is the case.

ok, thinking about it again, I think @ygalblum is correct, we should prevent adding features w/o potential users, like this, as they require maintenance and increase technical debt.

vishnuchalla:
Closing this as not required for now.

PR #874, opened 2025-06-04
[DNM] Improve pod latency measurement granularity

rsevilla87:
## Type of change

- Optimization

## Description

Thanks to this optimization we can have much better timestamps granularity.

- Related Issue #
- Closes #

### DIFF
--- a/pkg/measurements/pod_latency.go
+++ b/pkg/measurements/pod_latency.go
@@ -95,8 +95,13 @@ func (plmf podLatencyMeasurementFactory) NewMeasurement(jobConfig *config.Job, c
 func (p *podLatency) handleCreatePod(obj interface{}) {
 	pod := obj.(*corev1.Pod)
 	podLabels := pod.GetLabels()
+	creationTs := pod.CreationTimestamp.Time.UTC()
+	// If current timestamp is before the creation timestamp, use it
+	if time.Now().UTC().Before(pod.CreationTimestamp.Time.UTC()) {
+		creationTs = time.Now().UTC()
+	}
 	p.metrics.LoadOrStore(string(pod.UID), podMetric{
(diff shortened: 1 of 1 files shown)


rsevilla87:
This is affecting pod_latency for now, but same logic can be applied to other measurements wdyt @afcollins @jtaleric @josecastillolema ?

pkg/measurements/pod_latency.go
```diff
@@ -95,8 +95,13 @@ func (plmf podLatencyMeasurementFactory) NewMeasurement(jobConfig *config.Job, c
 func (p *podLatency) handleCreatePod(obj interface{}) {
 	pod := obj.(*corev1.Pod)
 	podLabels := pod.GetLabels()
+	creationTs := pod.CreationTimestamp.Time.UTC()
+	// If current timestamp is before the creation timestamp, use it
+	if time.Now().UTC().Before(pod.CreationTimestamp.Time.UTC()) {
```
vishnuchalla, line 100: This might hide the incorrect timestamps reported by kubernetes JSON status spec if there are any by always choosing the best lowest timestamp. Which will create a possibility for us to miss out on some errors that needs to be reported. Not sure if we are okay with that?

rsevilla87: The goal of this measurement is not to detect those kind of errors, IMO that's out of scope and blocking enhancements like this

ygalblum:
I don't understand how the code change addresses the subject of the PR. How does taking the current time of the machine running the test improves the granularity of the measurement? 

To me, it looks like the code change overcomes a timestamp issue on the cluster. But, I think that if such an issue exists, it's better to address it before running the test as it might be causing other issues, and not hide it

rsevilla87:
Seeems like it's not worth to work more on this PR, closing it.

PR #875, opened 2025-06-05
Public methods

rsevilla87:
## Description

These methods must be public so they can be used in external measurements implementations,

- Related Issue #
- Closes #

### DIFF
--- a/pkg/alerting/alert_manager.go
+++ b/pkg/alerting/alert_manager.go
@@ -94,6 +94,7 @@ func NewAlertManager(alertProfileCfg, uuid string, prometheusClient *prometheus.
 		uuid:       uuid,
 		indexer:    indexer,
 		metadata:   metadata,
+		embedCfg:   embedCfg,
 	}
 	if err := a.readProfile(alertProfileCfg); err != nil {
 		return &a, err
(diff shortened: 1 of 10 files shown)


venkataanil (approved):
LGTM

pkg/alerting/alert_manager.go
```diff
@@ -94,6 +94,7 @@ func NewAlertManager(alertProfileCfg, uuid string, prometheusClient *prometheus.
 		uuid:       uuid,
 		indexer:    indexer,
 		metadata:   metadata,
+		embedCfg:   embedCfg,
```
rsevilla87, line 97: This bugfix is also required to update kube-burner-ocp with to latest version

ygalblum:
I'm not sure these are the correct methods to expose. I think exposing `stopMeasurement` is better for this case

ygalblum:
General question I should have asked earlier, why are we supporting external measurements? Why not have them here? It's not like there's a three month release cadence that blocks whoever needs the new code

rsevilla87:
> General question I should have asked earlier, why are we supporting external measurements? Why not have them here? It's not like there's a three month release cadence that blocks whoever needs the new code

B/c we try have OpenShift related stuff in the kube-burner-ocp repository

rsevilla87:
> I'm not sure these are the correct methods to expose. I think exposing `stopMeasurement` is better for this case

Thanks, good suggestion, code changes are now much simpler

PR #880, opened 2025-06-08
feat(build): add security hardening flags to binary releases

### LINKED ISSUE #851: [RFE] security hardening for binary releases
as the package maintainer for kube-burner-bin in the arch user repository [1], I've identified that the pre-built binaries in official releases don't have a few security hardening features. this results in package warnings during installation and reduces the overall security posture of the application for end users who install via package managers rather than building from source.

Specifically, when performing package quality checks, the following security issues are flagged:
- ELF file lacks FULL RELRO
- Binary is unstripped
- ELF file lacks Position Independent Executable
- ELF file lacks GNU_PROPERTY_X86_FEATURE_1_SHSTK (Shadow Stack protection)

I would like the build process for official releases to include the following security hardening measures:
- Enable `FULL RELRO` by adding `-Wl,-z,relro,-z,now` to `LDFLAGS` during compilation
- Enable `PIE` by compiling with `-fPIE` -pie flags
- Enable Shadow Stack protection (`GNU_PROPERTY_X86_FEATURE_1_SHSTK`) if the target architecture supports it
- Consider providing stripped binaries for production releases (or both stripped and unstripped versions)

These changes would make the pre-built binaries more secure against common exploitation techniques without requiring end users to recompile from source.

the main alternative is to maintain a non-binary package where we compile from source with the appropriate security flags enabled. However, this adds complexity for users, increases build time, and defeats the purpose of providing pre-built binaries.

another option would be to create a post-processing step in my package to strip the binary, but this only addresses one of the issues. the more critical security features (RELRO, PIE, Shadow Stack) must be enabled at compile time and cannot be retrofitted to an existing binary AFAIK.

**Additional context**
these changes would benefit not just arch users but all users who download pre-built binaries directly from your GitHub releases.
these security features are becoming standard best practices in binary distribution, and many users and organizations are now expecting them for production software deployment.

[1] https://aur.archlinux.org/packages/kube-burner-bin

pfeifferj:
## Type of change

- Optimization

## Description

Pre-built binaries in official releases were missing several security hardening features, resulting in package warnings during installation and reduced security posture for end users.

The missing protections included:
- FULL RELRO (Read-Only Relocations)
- PIE (Position Independent Executable)
- Symbol stripping
- Build path information exposure

This commit updates both the Makefile and GoReleaser configuration to enable these security features:
- Enable CGO for Linux builds to support PIE and external linking
- Add `-buildmode=pie` for Position Independent Executable
- Add `-s -w ldflags` for symbol and DWARF stripping
- Add `-linkmode=external` with RELRO linker flags
- Add `-trimpath` to remove file system paths from binaries

Windows and macOS builds retain partial hardening (symbol stripping) as they don't support PIE without CGO.

Component: build

- Closes #851


ygalblum:
I'm not sure if we should add this hardening also to the Makefile that is used for local builds. Is there a downside to this hardening for local builds?

pfeifferj:
good and valid question.. well it could be a downside for devs since `-s -w` strip debug symbols and that would break debugging the build with tools like delve [1]. also `-trimpath` removes file system paths, which makes it harder for debuggers to map binary code back to source files and complicating stack trace analysis.

maybe the better approach would be just keep those flags in `.goreleaser.yml`, and add a make build-release target for end users who want the hardened local builds without disturbing the dev workflow.

wdyt?

[1] https://github.com/go-delve/delve

ygalblum:
A separate make target SGTM

pfeifferj:
I added the build target as discussed and reverted the flags from the normal build (but kept the line breaks because I think it makes the code more readable ^^)

ygalblum (approved):
LGTM

ygalblum:
@rsevilla87 @vishnuchalla are you ok with the new flags in the released binaries?

vishnuchalla:
> @rsevilla87 @vishnuchalla are you ok with the new flags in the released binaries?

doesn't seem to be breaking anything and the explanation makes sense. lgtm

PR #881, opened 2025-06-09
Fix network policy measurement

rsevilla87:
## Type of change

- Bug fix

## Description

We need to pass an instance of the embedCfg so it can read the network policy templates

- Related Issue #
- Closes #


pkg/measurements/netpol_latency.go
```diff
@@ -434,6 +434,7 @@ func (n *netpolLatency) processResults() {
 
 // Read network policy object template
 func readTemplate(o kconfig.Object, embedCfg *fileutils.EmbedConfiguration) ([]byte, error) {
+	fmt.Println(embedCfg)
```
ygalblum: Remove this line

ygalblum (approved):

PR #883, opened 2025-06-09
feat(job): add support for job-specific measurements

### LINKED ISSUE #715: [RFE] Activate per job measurements

At the moment, measurements are enabled globally and run during all jobs from a benchmark. 

To provide further flexibility we should provide the logic to enable measurements in a per-job basis.

This new feature can be a replacement of the current approach, or a new addition, that's something we need to evaluate though.

niladrix719:
## Type of change

- New feature

## Description

introduces job-specific measurements along side global ones

- Job-specific measurements will combine with existing global ones, ensuring backward compatibility
- Global measurements will apply to all jobs; if a job defines the same measurement, global measurements will be skipped
-  Documentation updated for measurements
- Tests added for measurements
- Logs for job-specific measurements have been added

Screenshot 2025-06-11 at 11 26 49 AM

- Related Issue
- Closes #715

### DIFF
--- a/pkg/measurements/factory.go
+++ b/pkg/measurements/factory.go
@@ -97,7 +97,6 @@ func (msf *MeasurementsFactory) NewMeasurements(jobConfig *config.Job, kubeClien
 	clientSet, restConfig := kubeClientProvider.ClientSet(jobConfig.QPS, jobConfig.Burst)
 	log.Infof("Initializing measurements for job: %s", jobConfig.Name)
 	mergedMeasurements := make(map[string]types.Measurement)
-	mergo.Merge(&mergedMeasurements, msf.ConfigSpec.GlobalConfig.Measurements, mergo.WithOverride)
 	for _, measurement := range msf.ConfigSpec.GlobalConfig.Measurements {
 		mergedMeasurements[measurement.Name] = measurement
 	}
@@ -123,8 +122,7 @@ func (msf *MeasurementsFactory) NewMeasurements(jobConfig *config.Job, kubeClien
 		}
 		mf, err := newMeasurementFactoryFunc(msf.ConfigSpec, measurement, msf.Metadata)
 		if err != nil {
-			log.Errorf("Failed to create measurement [%s]: %v", name, err)
-			continue
+			log.Fatalf("Failed to create measurement [%s]: %v", name, err)
 		}
 		ms.MeasurementsMap[name] = mf.NewMeasurement(jobConfig, clientSet, restConfig, embedCfg)
 		log.Infof("Registered measurement: %s", name)
--- a/pkg/measurements/factory.go
+++ b/pkg/measurements/factory.go
@@ -30,7 +30,6 @@ import (
 
 type MeasurementsFactory struct {
 	Metadata   map[string]any
-	Factories  map[string]MeasurementFactory
 	ConfigSpec config.Spec
 }
 
@@ -85,33 +84,10 @@ func NewMeasurementsFactory(configSpec config.Spec, metadata map[string]any, add
 			measurementFactoryMap[k] = v
 		}
 	}
-
-	measurementsFactory := MeasurementsFactory{
+	return &MeasurementsFactory{
 		Metadata:   metadata,
-		Factories:  make(map[string]MeasurementFactory, len(configSpec.GlobalConfig.Measurements)),
 		ConfigSpec: configSpec,
 	}
-	for _, measurement := range configSpec.GlobalConfig.Measurements {
-		if !isIndexerOk(configSpec, measurement) {
-			log.Fatalf("One of the indexers for measurement %s has not been found", measurement.Name)
-		}
-		if _, alreadyRegistered := measurementsFactory.Factories[measurement.Name]; alreadyRegistered {
-			log.Warnf("Measurement [%s] is registered more than once", measurement.Name)
-			continue
-		}
-		newMeasurementFactoryFunc, exists := measurementFactoryMap[measurement.Name]
-		if !exists {
-			log.Warnf("Measurement [%s] is not supported", measurement.Name)
-			continue
-		}
-		mf, err := newMeasurementFactoryFunc(configSpec, measurement, metadata)
-		if err != nil {
-			log.Fatal(err.Error())
-		}
-		measurementsFactory.Factories[measurement.Name] = mf
-		log.Infof("📈 Registered measurement: %s", measurement.Name)
-	}
-	return &measurementsFactory
 }
 
 func (msf *MeasurementsFactory) NewMeasurements(jobConfig *config.Job, kubeClientProvider *config.KubeClientProvider, embedCfg *fileutils.EmbedConfiguration) *Measurements {
@@ -119,9 +95,9 @@ func (msf *MeasurementsFactory) NewMeasurements(jobConfig *config.Job, kubeClien
 		MeasurementsMap: make(map[string]Measurement),
 	}
 	clientSet, restConfig := kubeClientProvider.ClientSet(jobConfig.QPS, jobConfig.Burst)
-
 	log.Infof("Initializing measurements for job: %s", jobConfig.Name)
 	mergedMeasurements := make(map[string]types.Measurement)
+	mergo.Merge(&mergedMeasurements, msf.ConfigSpec.GlobalConfig.Measurements, mergo.WithOverride)
 	for _, measurement := range msf.ConfigSpec.GlobalConfig.Measurements {
 		mergedMeasurements[measurement.Name] = measurement
 	}
@@ -137,9 +113,8 @@ func (msf *MeasurementsFactory) NewMeasurements(jobConfig *config.Job, kubeClien
 		}
 	}
 	for name, measurement := range mergedMeasurements {
-		if _, duplicate := ms.MeasurementsMap[name]; duplicate {
-			log.Warnf("Measurement [%s] is already registered", name)
-			continue
+		if !isIndexerOk(msf.ConfigSpec, measurement) {
+			log.Fatalf("One of the indexers for measurement %s has not been found", measurement.Name)
 		}
 		newMeasurementFactoryFunc, exists := measurementFactoryMap[name]
 		if !exists {
--- a/pkg/measurements/factory.go
+++ b/pkg/measurements/factory.go
@@ -17,6 +17,7 @@ package measurements
 import (
 	"sync"
 
+	"dario.cat/mergo"
 	"github.com/cloud-bulldozer/go-commons/v2/indexers"
 	"github.com/kube-burner/kube-burner/pkg/config"
 	"github.com/kube-burner/kube-burner/pkg/measurements/types"
@@ -28,8 +29,9 @@ import (
 )
 
 type MeasurementsFactory struct {
-	Metadata  map[string]any
-	Factories map[string]MeasurementFactory
+	Metadata   map[string]any
+	Factories  map[string]MeasurementFactory
+	ConfigSpec config.Spec
 }
 
 type Measurements struct {
@@ -85,8 +87,9 @@ func NewMeasurementsFactory(configSpec config.Spec, metadata map[string]any, add
 	}
 
 	measurementsFactory := MeasurementsFactory{
-		Metadata:  metadata,
-		Factories: make(map[string]MeasurementFactory, len(configSpec.GlobalConfig.Measurements)),
+		Metadata:   metadata,
+		Factories:  make(map[string]MeasurementFactory, len(configSpec.GlobalConfig.Measurements)),
+		ConfigSpec: configSpec,
 	}
 	for _, measurement := range configSpec.GlobalConfig.Measurements {
 		if !isIndexerOk(configSpec, measurement) {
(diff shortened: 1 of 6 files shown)


niladrix719:
I think this log might be confusing now, after addition of job specific logs (need some help on this)

https://github.com/kube-burner/kube-burner/blob/7b833b172d33e56fd6a12e0ee6087afd32ad3f66/pkg/measurements/factory.go#L108-L110

pkg/measurements/factory.go
```diff
@@ -111,11 +113,34 @@ func NewMeasurementsFactory(configSpec config.Spec, metadata map[string]any, add
...
+			newMeasurementFactoryFunc, exists := measurementFactoryMap[measurement.Name]
+			if !exists {
+				log.Warnf("Measurement [%s] is not supported", measurement.Name)
+				continue
+			}
+			mf, err := newMeasurementFactoryFunc(msf.ConfigSpec, measurement, msf.Metadata)
+			if err != nil {
+				log.Fatal(err.Error())
+				continue
+			}
```
pfeifferj: `log.Fatal()` [1] calls `os.Exit(1)` [2] which makes the continue unreachable

[1] https://pkg.go.dev/log#Fatal
[2] https://pkg.go.dev/os#Exit

pkg/measurements/factory.go
```diff
@@ -111,11 +113,34 @@ func NewMeasurementsFactory(configSpec config.Spec, metadata map[string]any, add
...
 	for name, factory := range msf.Factories {
 		ms.MeasurementsMap[name] = factory.NewMeasurement(jobConfig, clientSet, restConfig)
+		log.Infof("📈 Measurement: %s (global)", name)
+	}
+	if len(jobConfig.Measurements) > 0 {
+		for _, measurement := range jobConfig.Measurements {
+			// Skip if measurement is already registered through global config
+			if _, exists := ms.MeasurementsMap[measurement.Name]; exists {
+				continue
+			}
```
pfeifferj: this silently skips job-specific measurements when a global measurement with the same name exists

not logging or documenting this behaviour could lead to user confusion when they define job-specific measurements that mysteriously don't take effect.

pkg/measurements/factory.go
```diff
@@ -111,11 +113,34 @@ func NewMeasurementsFactory(configSpec config.Spec, metadata map[string]any, add
...
 	ms := Measurements{
-		MeasurementsMap: make(map[string]Measurement, len(msf.Factories)),
+		MeasurementsMap: make(map[string]Measurement),
 	}
 	clientSet, restConfig := kubeClientProvider.ClientSet(jobConfig.QPS, jobConfig.Burst)
+
+	log.Infof("Initializing measurements for job: %s", jobConfig.Name)
 	for name, factory := range msf.Factories {
 		ms.MeasurementsMap[name] = factory.NewMeasurement(jobConfig, clientSet, restConfig)
+		log.Infof("📈 Measurement: %s (global)", name)
```
pfeifferj: the logging here doesn't follow the project's established patterns. 
regular operational logs should avoid emojis for better log parsing and consistency

pkg/measurements/factory.go
```diff
@@ -111,11 +113,34 @@ func NewMeasurementsFactory(configSpec config.Spec, metadata map[string]any, add
...
+				log.Warnf("Measurement [%s] is not supported", measurement.Name)
+				continue
+			}
+			mf, err := newMeasurementFactoryFunc(msf.ConfigSpec, measurement, msf.Metadata)
+			if err != nil {
+				log.Fatal(err.Error())
+				continue
+			}
+			ms.MeasurementsMap[measurement.Name] = mf.NewMeasurement(jobConfig, clientSet, restConfig)
+			log.Infof("📈 Measurement: %s (job-specific)", measurement.Name)
```
pfeifferj: same as above

pfeifferj (changes requested):
left some comments

pls squash your commits into #51d11b3
it would also be good to update the documentation, and add tests

pkg/measurements/factory.go
```diff
@@ -112,9 +114,26 @@ func NewMeasurementsFactory(configSpec config.Spec, metadata map[string]any, add
...
+				continue
+			}
+			mf, err := newMeasurementFactoryFunc(msf.ConfigSpec, measurement, msf.Metadata)
+			if err != nil {
+				log.Fatal(err.Error())
+			}
+			ms.MeasurementsMap[measurement.Name] = mf.NewMeasurement(jobConfig, clientSet, restConfig, embedCfg)
+			log.Infof("Registered job-specific measurement: %s", measurement.Name)
+		}
+	}
```
pfeifferj: job-specific measurements (lines 122-136) create NEW factory instances instead of using configuration from job-specific settings

when a measurement exists both globally and job-specific, the code skips the global one (line 139) but doesn't merge configurations

pkg/measurements/factory.go
```diff
@@ -112,9 +114,26 @@ func NewMeasurementsFactory(configSpec config.Spec, metadata map[string]any, add
...
+	if len(jobConfig.Measurements) > 0 {
+		for _, measurement := range jobConfig.Measurements {
+			newMeasurementFactoryFunc, exists := measurementFactoryMap[measurement.Name]
+			if !exists {
+				log.Warnf("Measurement [%s] is not supported", measurement.Name)
+				continue
+			}
+			mf, err := newMeasurementFactoryFunc(msf.ConfigSpec, measurement, msf.Metadata)
+			if err != nil {
+				log.Fatal(err.Error())
```
pfeifferj: `log.Fatal` should generally only be used in main or init functions
I would log an error message here. this follows the go proverb: "don't panic" (and by extension, don't `log.Fatal` in libraries/packages)

pkg/measurements/factory.go
```diff
@@ -112,9 +114,26 @@ func NewMeasurementsFactory(configSpec config.Spec, metadata map[string]any, add
...
+
+	log.Infof("Initializing measurements for job: %s", jobConfig.Name)
+	if len(jobConfig.Measurements) > 0 {
+		for _, measurement := range jobConfig.Measurements {
+			newMeasurementFactoryFunc, exists := measurementFactoryMap[measurement.Name]
+			if !exists {
+				log.Warnf("Measurement [%s] is not supported", measurement.Name)
+				continue
+			}
+			mf, err := newMeasurementFactoryFunc(msf.ConfigSpec, measurement, msf.Metadata)
```
pfeifferj: consider adding a duplicate check

pkg/measurements/factory.go
```diff
@@ -112,11 +114,33 @@ func NewMeasurementsFactory(configSpec config.Spec, metadata map[string]any, add
...
+			if err != nil {
+				log.Fatal(err.Error())
+			}
+			ms.MeasurementsMap[measurement.Name] = mf.NewMeasurement(jobConfig, clientSet, restConfig, embedCfg)
+			log.Infof("Registered job-specific measurement: %s", measurement.Name)
+		}
+	}
 	for name, factory := range msf.Factories {
+		if _, exists := ms.MeasurementsMap[name]; exists {
```
vishnuchalla: What's the point of doing this? Looks like the only ones in the local scope that overlaps will be ignored but rest of the measurements in the global scope might end up running.

niladrix719: Yeah right, I wanted to give the flexibility to override global measurements at the job level while maintaining a default behavior. The current implementation follows a hierarchical configuration pattern where global measurements serve as defaults that can be selectively overridden by job-specific measurements. Sometimes we need all jobs to have default global values with few exceptions, and in those cases, this approach makes it easier.

So that was my intuition - do you think it would be better to:

- Keep this hierarchical override pattern, or
- Have completely separate global vs job-specific measurements with no override behavior?

pkg/measurements/factory.go
pfeifferj: this was not addressed

pfeifferj (changes requested):

pkg/measurements/factory.go
niladrix719: Okay, so we want merging at the configuration level. In that case, we can use some libraries to merge the objects

pkg/measurements/factory.go
```diff
@@ -112,11 +115,43 @@ func NewMeasurementsFactory(configSpec config.Spec, metadata map[string]any, add
...
+				log.Errorf("Failed to merge measurement [%s]: %v", measurement.Name, err)
+				continue
+			}
+			mergedMeasurements[measurement.Name] = globalMeasurement
+		} else {
+			mergedMeasurements[measurement.Name] = measurement
+		}
+	}
+	for name, measurement := range mergedMeasurements {
+		if _, duplicate := ms.MeasurementsMap[name]; duplicate {
```
vishnuchalla: Do you think we will even run into this case where a duplicate will exist as a value for same key in a map?

pkg/measurements/factory.go
```diff
@@ -112,11 +115,43 @@ func NewMeasurementsFactory(configSpec config.Spec, metadata map[string]any, add
...
-		ms.MeasurementsMap[name] = factory.NewMeasurement(jobConfig, clientSet, restConfig, embedCfg)
+
+	log.Infof("Initializing measurements for job: %s", jobConfig.Name)
+	mergedMeasurements := make(map[string]types.Measurement)
+	for _, measurement := range msf.ConfigSpec.GlobalConfig.Measurements {
+		mergedMeasurements[measurement.Name] = measurement
+	}
+	for _, measurement := range jobConfig.Measurements {
+		if globalMeasurement, exists := mergedMeasurements[measurement.Name]; exists {
+			if err := mergo.Merge(&globalMeasurement, measurement, mergo.WithOverride); err != nil {
```
vishnuchalla, line 105: what's the point of merge here? that too being overridden in a hash map?

niladrix719: The merge is intentional — it's not just an overwrite. I am using `mergo.Merge` to combine fields from the global and job-specific configs, allowing job-specific fields to override and global ones to act as defaults. Since `globalMeasurement` is a copy, we need to reassign the merged result back into the map to persist it, I think I should also add this part in the readme for clarity, do you agree with this approach

test/k8s/kube-burner-measurements.yml
```diff
@@ -0,0 +1,47 @@
+---
+global:
+  functionTemplates:
+    - objectTemplates/envs.tpl
+  measurements:
```
vishnuchalla, line 5: Lets add one more here, for example `nodeLatency` and make sure that the `precedence-measurements` is only executing `podLatency` when it comes to that job.

niladrix719: If we add `nodeLatency` in the global and keep the job-specific as it is now, then the final result would be `precedence-measurements` will have `[nodeLatency, podLatency]` and `merge-measurements` will have `[nodeLatency, podLatency, serviceLatency]`, as current implementation combines the global with job specific (docs - https://github.com/kube-burner/kube-burner/blob/de9bf1738e564fe3fd6cb75d751203c71ef94921/docs/measurements/index.md?plain=1#L7-L10)

```
global:
  measurements:
    - name: nodeLatency
    - name: podLatency

jobs:
  - name: precedence-measurements
    measurements:
      - name: podLatency
    # Result: [nodeLatency, podLatency]

  - name: merge-measurements
    measurements:
      - name: serviceLatency
        svcTimeout: 2s
    # Result: [nodeLatency, podLatency, serviceLatency]
```

And one more thing - as stated in the docs if both global and job have the same measurements, let's say both have `serviceLatency` with different `svcTimeout` like `3s` and `5s` respectively, only a single `serviceLatency` will be applied to the job with the `5s` `svcTimeout`, as job-specific `serviceLatency` overwrites the global `serviceLatency` with `3s`

So the current implementation will work like this. I made it this way, but if we want to proceed with only job-specific measurements without combining them with global, then what you said will work. Do we want it that way? I might have implemented it differently than what you were expecting

vishnuchalla: Thanks for the clarification. My thought is simple, if a user specifies a list of measurements at the job level, then they should be the only thing to be executed regardless of what is specified at the global level (i.e something like a override). That way it will provide user flexibility to make a choice between job level or leave it up to the list specified at global level.

Let's see what others think
cc: @rsevilla87 @ygalblum

ygalblum: Obviously there are arguments for both sides. To me, I like the ability to set a global that is always used and then add per job. For example, always track podLatency and sometimes track additional measurements (as was implemented here)

rsevilla87: I like @vishnuchalla's point, job level measurements when defined, should be the only ones executed regardless what is specified globally (that will make the logic more intuitive and simple).

We can maybe improve the logic in another follow-up PR, in case we see a potential use case, but the initial intention of this issue was to skip measurements in jobs not needing them. (hence no declaring them in the global section).

i.e:

```yaml

global
  gc: false
jobs:
  - name: pre-reqs
    jobIterations: 1
    objects:
      - objectTemplate: rolebinding.yml
        replicas: 1

   - name: job2
     jobIterations: 10
     measurements:
       - name: podLatency
     objects:
        - objectTemplate: deployment.yml
          replicas: 10
```

ygalblum: > skip measurements in jobs not needing them. (hence no declaring them in the global section).

Sure. But this means that users might need to define the same measurement twice. See my previous example:
> always track podLatency and sometimes track additional measurements

rsevilla87: Thanks for your suggestions  @ygalblum , I'm gonna take over this one and wrap up pending details, agree on your duplicated measurements suggestion, I wanted to simplify the logic before moving forward.

rsevilla87:
Just updated the code @niladrix719 @vishnuchalla to remove some duped logic, should be ready to go now

pkg/measurements/factory.go
```diff
@@ -83,41 +84,50 @@ func NewMeasurementsFactory(configSpec config.Spec, metadata map[string]any, add
...
 	}
-	for _, measurement := range configSpec.GlobalConfig.Measurements {
-		if !isIndexerOk(configSpec, measurement) {
-			log.Fatalf("One of the indexers for measurement %s has not been found", measurement.Name)
+	clientSet, restConfig := kubeClientProvider.ClientSet(jobConfig.QPS, jobConfig.Burst)
+	log.Infof("Initializing measurements for job: %s", jobConfig.Name)
+	mergedMeasurements := make(map[string]types.Measurement)
+	mergo.Merge(&mergedMeasurements, msf.ConfigSpec.GlobalConfig.Measurements, mergo.WithOverride)
+	for _, measurement := range msf.ConfigSpec.GlobalConfig.Measurements {
+		mergedMeasurements[measurement.Name] = measurement
```
ygalblum: I'm probably missing something here. Doesn't the loop do the same as when `mergo.Merge` does?

pkg/measurements/factory.go
ygalblum: Sounds good to me

pkg/measurements/factory.go
```diff
@@ -83,41 +84,50 @@ func NewMeasurementsFactory(configSpec config.Spec, metadata map[string]any, add
...
 		if !exists {
-			log.Warnf("Measurement [%s] is not supported", measurement.Name)
+			log.Warnf("Measurement [%s] is not supported", name)
 			continue
 		}
-		mf, err := newMeasurementFactoryFunc(configSpec, measurement, metadata)
+		mf, err := newMeasurementFactoryFunc(msf.ConfigSpec, measurement, msf.Metadata)
 		if err != nil {
-			log.Fatal(err.Error())
+			log.Errorf("Failed to create measurement [%s]: %v", name, err)
```
ygalblum: Not sure we shouldn't fail here, or at least propagate the error. If the user wanted to get measurements, not sure they want to continue with the test if they won't be collected

test/k8s/kube-burner-measurements.yml
vishnuchalla: I see, I am okay with the idea of merging measurements. But , I think we need to provide users with flexibility to execute only job level measurements if required with an override flag. That way they don't have to deal with noise configured at global level and also it's going to be less error prone (.i.e per say if a measurement in global is broken and they did not want to use it all, they can stick to their job level measurements alone and don't have to run into errors with measurements at global level).

ygalblum: If they don't define anything at the global level but only on the job level, won't that fit your requirement?

vishnuchalla: Unfortunately no. If I want global measurements to be as a fallback when there aren't any overrides in local for some of the jobs in the list representing my workload.

pkg/measurements/factory.go
rsevilla87: That's right, removing the Merge line, which is actually useless

pkg/measurements/factory.go
rsevilla87: correct, let's use Fatal instead

rsevilla87: Correct, let's use Fatalf instead

ygalblum (approved):
LGTM

rsevilla87 (approved):

PR #887, opened 2025-06-10
Improve metricsClosing logic and update default value

rsevilla87:
## Type of change

- Optimization

## Description

Let's try to use constants and log the metricsClosing option selected. And also let's update the default metricsClosing value to `afterJobPause` to keep previous behaviour. 

- Related Issue #
- Closes #

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -187,8 +187,8 @@ func ParseWithUserdata(uuid string, timeout time.Duration, configFileReader, use
 		if job.JobIterations < 1 && (job.JobType == CreationJob || job.JobType == ReadJob) {
 			log.Fatalf("Job %s has < 1 iterations", job.Name)
 		}
-		if job.MetricsClosing != "afterJob" && job.MetricsClosing != "afterJobPause" && job.MetricsClosing != "afterMeasurements" {
-			log.Fatalf("Invalid value %s for MetricsClosing", job.MetricsClosing)
+		if _, ok := metricsClosing[job.MetricsClosing]; !ok {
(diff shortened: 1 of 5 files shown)


pkg/config/types.go
```diff
@@ -49,8 +50,17 @@ const (
 	KubeVirtOpMigrate      KubeVirtOpType = "migrate"
 	KubeVirtOpAddVolume    KubeVirtOpType = "add-volume"
 	KubeVirtOpRemoveVolume KubeVirtOpType = "remove-volume"
+	AfterJobPause          MetricsClosing = "afterJobPause"
```
ygalblum: Since this is a different type, to make it more readable either move to a separate const section, or add a newline

ygalblum (dismissed):
(very) small nit. Otherwise LGTM. I'm approving it, just decide if you want to address or merge as is

pkg/config/config.go
```diff
@@ -104,7 +104,7 @@ func (j *Job) UnmarshalYAML(unmarshal func(any) error) error {
 		ChurnDuration:          1 * time.Hour,
 		ChurnDelay:             5 * time.Minute,
 		ChurnDeletionStrategy:  "default",
-		MetricsClosing:         "afterJob",
+		MetricsClosing:         AfterJobPause,
```
rsevilla87, line 107: This line is changing the default value

pkg/config/types.go
rsevilla87: done

jtaleric (approved):
let

vishnuchalla (approved):
lgtm

PR #889, opened 2025-06-11
Feature/image pull latency

### LINKED ISSUE #882: [RFE]  Add ImagePullLatency Measurement to Capture Container Image Pull Delays
Most of the time taken in pod startup latency is due to pulling images from remote registries. Currently `podLatency` offers a lot of insights but does not include image pull time. It would be good if I could get a `imagePullLatency` measurement which I can use to know when it's better to cache images and apply other optimizations in cluster. Does this measurement seem like a viable metric that we can implement in kube-burner? what do you guys think?

To introduce a new measurement type named imagePullLatency, which captures the time duration specifically for pulling container images during pod creation. This can be derived by monitoring Kubernetes events such as:
- "Pulling" (start of image pull)
- "Pulled" or "Created" (end of image pull)
And then calculating.

`podLatency` does provide initialisation and container-ready time but it's not granular enough. The time between them contains other operations than pulling images too.

**Additional context**

VijetaPriya47:
## Type of change

-
- New feature

## Description
I implemented a feature in image_pull_latency.go that listens to pod events and tracks how long it takes to pull container images. Specifically, I monitor when a container enters the "Pulling" state and when it transitions to "Running." By recording the timestamps for these events, I calculate the image pull latency — essentially the time taken to download the image.

I store this latency data so it can be aggregated and used for reporting, giving clear insights into image pull delays across the system.

I also wrote a test to verify that this logic works as expected and accurately measures the image pull time.

I wanted you to review this, and if the implementation is fien, then, I will add documentation update also

- Related Issue #
- Closes #882


image-pull-latency.yml
```diff
@@ -0,0 +1,19 @@
+---
```
ygalblum: This file should either be removed or moved somewhere else

pkg/measurements/image_pull_latency.go
```diff
@@ -0,0 +1,176 @@
...
+func newImagePullLatencyMeasurementFactory(configSpec config.Spec, measurement types.Measurement, metadata map[string]any) (MeasurementFactory, error) {
+	return imagePullLatencyMeasurementFactory{
+		BaseMeasurementFactory: NewBaseMeasurementFactory(configSpec, measurement, metadata),
+	}, nil
+}
+
+func (iplmf imagePullLatencyMeasurementFactory) NewMeasurement(jobConfig *config.Job, clientSet kubernetes.Interface, restConfig *rest.Config) Measurement {
+	return &imagePullLatency{
+		BaseMeasurement: iplmf.NewBaseLatency(jobConfig, clientSet, restConfig, imagePullLatencyMeasurement, imagePullLatencyQuantilesMeasurement),
+		Ctx:             context.Background(),
```
ygalblum: Why do you need to keep one at the struct level? Just call `context.Background()` when you need it

pkg/measurements/image_pull_latency.go
```diff
@@ -0,0 +1,176 @@
...
+	imagePullLatencyQuantilesMeasurement = "imagePullLatencyQuantilesMeasurement"
+)
+
+type imagePullMetric struct {
+	Timestamp     time.Time `json:"timestamp"`
+	PodName       string    `json:"podName"`
+	ContainerName string    `json:"containerName"`
+	Image         string    `json:"image"`
+	PullStartTime time.Time `json:"pullStartTime"`
+	PullEndTime   time.Time `json:"pullEndTime"`
```
ygalblum: Fields with json tags will be exported to the index. In other measurements we do not expose the times, only the latency.

pkg/measurements/image_pull_latency.go
```diff
@@ -0,0 +1,176 @@
...
+		NodeName:     pod.Spec.NodeName,
+	})
+}
+
+func (ipl *imagePullLatency) handleUpdatePod(obj any) {
+	pod := obj.(*corev1.Pod)
+	if value, exists := ipl.metrics.Load(string(pod.UID)); exists {
+		ipm := value.(imagePullMetric)
+		for _, containerStatus := range pod.Status.ContainerStatuses {
+			if containerStatus.State.Waiting != nil && containerStatus.State.Waiting.Reason == "Pulling" {
```
ygalblum: How does this work with a multi-container pod? Maybe the right thing to do here is to manage each container separately

pkg/measurements/image_pull_latency.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+}
+
+func (ipl *imagePullLatency) Start(measurementWg *sync.WaitGroup) error {
+	defer measurementWg.Done()
+	ipl.startMeasurement(
+		[]MeasurementWatcher{
+			{
+				restClient:    ipl.ClientSet.CoreV1().RESTClient().(*rest.RESTClient),
+				name:          "podWatcher",
```
ygalblum: Please give it a different name, `podWatcher` is already used by pod_latency

pkg/measurements/image_pull_latency.go
```diff
@@ -0,0 +1,176 @@
...
+		if ipm.PullLatency > 0 {
+			totalLatency += float64(ipm.PullLatency)
+			count++
+		}
+		return true
+	})
+	if count == 0 {
+		return 0
+	}
+	return totalLatency / float64(count)
```
ygalblum: The return value is the error percentage, not the average latency. See other implementations for example. Alternatively, you can return 0

pkg/measurements/image_pull_latency_test.go
```diff
@@ -0,0 +1,103 @@
...
+				{
+					Name:  "test-container",
+					Image: "test-image",
+				},
+			},
+		},
+	}
+
+	// Test pod creation
+	measurementInstance.(*imagePullLatency).handleCreatePod(pod)
```
ygalblum: Since you cast several times, consider doing it once after creating the instance

pkg/measurements/image_pull_latency_test.go
```diff
@@ -0,0 +1,103 @@
...
+				Waiting: &corev1.ContainerStateWaiting{
+					Reason: "Pulling",
+				},
+			},
+		},
+	}
+	measurementInstance.(*imagePullLatency).handleUpdatePod(pod)
+
+	// Simulate time passing between pulling and running states
+	time.Sleep(100 * time.Millisecond)
```
ygalblum: Instead of simulating the time, just pass a larger number to `StartedAt`, this will allow you to also verify the actual value

pod-template.yml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
```
ygalblum: This file does not belong here

VijetaPriya47: removed it

VijetaPriya47:
can you check again @ygalblum

pkg/measurements/image_pull_latency_test.go
```diff
@@ -0,0 +1,106 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-pod",
+			Namespace: "default",
+			UID:       "test-uid",
+			Labels: map[string]string{
+				"kube-burner-runid": "test-uuid",
+			},
+		},
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
```
ygalblum, line 49: I think having a second container will improve the test without a lot of effort

pkg/measurements/image_pull_latency_test.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	ipl.handleUpdatePod(pod)
+
+	// Set a specific StartedAt value to simulate time passing
+	startedAt := metav1.NewTime(time.Now().Add(-200 * time.Millisecond))
+
+	// Test pod update with running state
+	pod.Status.ContainerStatuses[0].State = corev1.ContainerState{
+		Running: &corev1.ContainerStateRunning{
+			StartedAt: startedAt,
```
ygalblum: Unless I missed it, I didn't notice this field is looked at. Now that I see that you don't have timestamps on the status and you use local time, I guess you will need to actually wait in the test

ygalblum (commented):
OK for the code, some nits on the test code

VijetaPriya47:
@ygalblum hyy can you also check once, why is it showing linter error?

pkg/measurements/image_pull_latency_test.go
VijetaPriya47: well instead of using sleep, I just refractored the code to use `StartedAt` . I've also updated the tests to cover multi-container pods and to assert the exact pull latency for each container.

ygalblum:
@VijetaPriya47 I'm not sure why the lint if failing. But, you need to rebase on main in any case because the API changed. So, please rebase and fix the implementation and then we'll see again

ygalblum:
@VijetaPriya47 I guess some of the functions changed since you started this. You'll need to adjust your code

vishnuchalla:
Closing as this has been stale for a while.

PR #896, opened 2025-06-14
Fix : validate QPS and Burst to prevent panic and improve logging

Ritish134:
## Description

This change ensures the application handles these situations gracefully.

---

### Before (example failure cases)

**Case: QPS invalid → panic**

Image

**Case: QPS valid, Burst negative → panic**

Image

---

### After (example with fix applied)

**No panic, invalid values are corrected at runtime** Image


Ritish134:
Done

ygalblum (approved):
Thanks for your contribution

PR #895, opened 2025-06-12
precreate-namespace

SamFajobi513:
## Type of change

- New feature

## Description
This pull request introduces a new feature in kube-burner that allows namespaces to be pre-created and optionally waited upon before the main resource objects are rendered and applied.

Key Features
   -preCreateNamespaces: true — When set, the namespaces listed in the job's namespace field are created before rendering      and applying objects.

   -waitBeforeCreateObjects: 10s (or any duration) — Optional field that allows users to introduce a delay after namespace creation and before object creation begins.

This addition is useful for cases where other systems (like operators or admission controllers) need the namespace to exist before objects are applied.

## How to test
To test this feature:

1.Ensure kube-burner is built and a Kubernetes cluster is running.

2.Use the sample job file test-examples/test-job.yml to run a simple test

3.Then run:
bin/amd64/kube-burner init -c test-examples/test-job.yml --log-level debug

4.Observe the log output:
 -Namespace Pre-Creation and Reuse Logic:
The job now pre-creates namespaces (test-ns) and attempts to reuse them, which is evident from repeated log entries about namespace creation and object duplication (e.g., Pod already exists)

 -Finally, it will proceed to create the pod or other specified objects

-Object Verification After Job Execution:
  A verification step now runs after object creation to compare expected vs. actual objects, reporting mismatches (e.g., "pods     found: 1 Expected: 2"), helping to catch deployment inconsistencies

-  A delay matching waitBeforeCreateObjects (implementation in progress)

## Related Tickets & Documents

- Related Issue #
- Closes #850

PR #898, opened 2025-06-16
Pre create namespaces

### LINKED ISSUE #850: [RFE] Pre-create namespaces

There're some cases where a namespace needs to be created some seconds upfront before adding objects to it, since some controllers take some time to reconcile and process those namespaces

Having an option to pre-create namespaces for the specified jobs.

**Additional context**
For example, the BGP UDN workload https://github.com/kube-burner/kube-burner-ocp/pull/200/ needs the namespaces to be created beforehand

rsevilla87:
## Type of change

- New feature

## Description

Implementation of the pre-create namespaces functionallity

- Related Issue #
- Closes #850


ygalblum:
We need some coordination on work items. This PR is the same as #895

ygalblum:
I don't understand how this change is helpful if you're not waiting for any condition after the namespace is created. If you run a single iteration (and hence a single namespace), will there actually be any change? Maybe I'm missing something

Issue #901, opened 2025-06-16
[BUG] CI failing on deploy-docs job


pfeifferj:
# Bug Description

## **Describe the bug**

The `deploy-docs` CI job is failing with exit code 1 when trying to delete old documentation versions:

```
error: unable to delete nonexistant identifier 'v1.2'
Error: Process completed with exit code 1.
```

e.g. [latest run](https://github.com/kube-burner/kube-burner/actions/runs/15684256973/job/44187900892)

PR #902, opened 2025-06-16
fix(ci): fix deploy-docs CI job failing on mike delete with substring matching

pfeifferj:
## Type of change

- CI

## Description

the deploy-docs workflow was failing when trying to delete old documentation versions due to a grep pattern matching issue. the script used substring matching (`grep -q "$tag"`) which would incorrectly match partial version strings - for example, checking for "v1.2" would match "v1.20.6" and assume the documentation existed

this caused the workflow to attempt deleting non-existent versions like "v1.2" when only "v1.20.6" actually existed, resulting in: "error: unable to delete nonexistant identifier 'v1.2'"

the fix changes the grep pattern to use exact line matching (`grep -q "^$tag$"`) and updates mkdocs.yml to use the new material emoji extensions instead of the deprecated materialx.emoji extensions

also added error handling to prevent individual delete failures from breaking the entire ci workflow

- Closes: #901 
- Component: ci/docs

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -33,9 +33,9 @@ jobs:
           tags_to_keep=$(echo "$all_tags" | head -n 3)
           for tag in $all_tags; do
             if [[ $tags_to_keep != *"$tag"* ]]; then
-              if mike list | grep -q "$tag"; then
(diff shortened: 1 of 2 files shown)


vishnuchalla (approved):
Thanks for this fix.

PR #903, opened 2025-06-16
Waiters - limit resource list to ones created by the current run

ygalblum:
## Type of change

- Bug fix
- CI

## Description
When waiting for resources to reach the desired state, list calls did not limit the list to the resources created by the current execution causing wait to hang on unrelated resources.
Add the current runid to the label selector when listing resources for wait

- Related Issue https://github.com/kube-burner/kube-burner-ocp/issues/263

### DIFF
--- a/pkg/burner/waiters.go
+++ b/pkg/burner/waiters.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"maps"
 	"time"
 
 	"github.com/itchyny/gojq"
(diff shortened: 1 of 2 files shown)


pkg/burner/waiters.go
```diff
@@ -123,11 +129,11 @@ func (ex *Executor) waitForObject(ns string, obj *object) {
 	}
 }
 
-func (ex *Executor) waitForReplicas(ns string, obj object, waitPath statusPath) error {
+func (ex *Executor) waitForReplicas(ns string, obj *object, waitPath statusPath, labelSelector string) error {
```
rsevilla87: Why are you using a pointer now?

ygalblum: To avoid having to dereference it in everywhere it is called. Probably should have done it earlier (or separately)

rsevilla87: np, sounds good!

rsevilla87 (dismissed):
Good change, it helps to improve the waiters logic in scenarios reusing namespaces

vishnuchalla (approved):
lgtm

PR #906, opened 2025-06-17
Fix(cli): Clarify default UUID behavior with help

### LINKED ISSUE #905: [BUG] CLI: Misleading default value for --uuid flag
# Bug Description

## **Describe the bug**

The --uuid flag for several kube-burner subcommands (init, check-alerts, index, measure) displays a randomly generated UUID as the default value in the help text. This is misleading as it suggests the value is hardcoded or that this specific UUID will be used, when in reality a new one is generated for each benchmark run if not provided. This can cause confusion for new users.

$ kube-burner version
Version: main
Git Commit: 5626b9f4a21b3219079df02f8362133aa1654a55
Build Date: 2025-06-10-15:46:46
Go Version: go1.23.9
OS/Arch: linux amd64

## **To Reproduce**

1. Open a terminal.
2. Run kube-burner init --help to see the misleading default UUID.
3. Run the command again and observe that the "default" UUID has changed.
4. This can be observed for other commands as well, like kube-burner check-alerts --help.

## **Expected behavior**

The help text for the --uuid flag should not display a random, hardcoded-looking value. Instead, the default value should be empty, and the help description should clearly state that a UUID will be generated automatically if not provided by the user.

For example:
--uuid string Benchmark UUID (generated automatically if not provided)

## **Screenshots or output**

kube-burner init --help
Launch benchmark

Usage:
  kube-burner init [flags]

Flags:
      --uuid string               Benchmark UUID (default "3b46c153-bd56-4042-bc37-aed74e7011a9")

## **Additional context**

The issue stems from the flag's default value being set to the result of uid.NewString() at command initialization time. This function is executed whenever the help text is generated, leading to a new random UUID appearing as the default each time.

Sandeepyadav93:
## Type of change

- Bug fix

## Description

The --uuid flag for the `init`, `check-alerts`, `index` commands displayed a randomly-generated UUID as its default value in the help text. This was misleading to users, suggesting the value was hardcoded or that they needed to override a specific default.

This commit addresses the issue in two ways:

1.  The default value for the flag definition is changed to an empty string. This provides a clean and predictable --help output.

2.  Logic is added to the beginning of each command's execution to generate a new UUID if one was not provided by the user.

This approach results in a more intuitive user experience while ensuring that a unique benchmark UUID is always present at runtime, preserving the original functionality.

Old behavior:-
~~~
$kube-burner init --help
Launch benchmark

Usage:
  kube-burner init [flags]

Flags:
      --uuid string               Benchmark UUID (default "a22992bc-e0f4-4301-8340-25b5ef4be94b")
~~~

New behavior:-
~~~
$kube-burner init --help
Launch benchmark

Usage:
  kube-burner init [flags]

Flags:
      --uuid string               Benchmark UUID (generated automatically if not provided)
~~~

- Closes #905

### DIFF
--- a/cmd/kube-burner/kube-burner.go
+++ b/cmd/kube-burner/kube-burner.go
@@ -91,6 +91,11 @@ func initCmd() *cobra.Command {
 			os.Exit(rc)
 		},
 		Args: cobra.NoArgs,
+		PreRun: func(cmd *cobra.Command, args []string) {
+			if uuid == "" {
+				uuid = uid.NewString()
+			}
+		},
 		Run: func(cmd *cobra.Command, args []string) {
 			if configMap != "" {
 				metricsProfile, alertProfile, err = config.FetchConfigMap(configMap, namespace)
@@ -137,7 +142,7 @@ func initCmd() *cobra.Command {
 			}
 		},
 	}
-	cmd.Flags().StringVar(&uuid, "uuid", uid.NewString(), "Benchmark UUID")
+	cmd.Flags().StringVar(&uuid, "uuid", "", "Benchmark UUID (generated automatically if not provided)")
 	cmd.Flags().StringVarP(&metricsEndpoint, "metrics-endpoint", "e", "", "YAML file with a list of metric endpoints")
 	cmd.Flags().BoolVar(&skipTLSVerify, "skip-tls-verify", true, "Verify prometheus TLS certificate")
 	cmd.Flags().DurationVarP(&timeout, "timeout", "", 4*time.Hour, "Benchmark timeout")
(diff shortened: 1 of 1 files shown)


venkataanil (approved):
LGTM

cmd/kube-burner/kube-burner.go
```diff
@@ -92,6 +92,9 @@ func initCmd() *cobra.Command {
 		},
 		Args: cobra.NoArgs,
 		Run: func(cmd *cobra.Command, args []string) {
+			if uuid == "" {
+				uuid = uid.NewString()
+			}
```
chentex, line 97: Thinking out lout here, this might be a good piece of code to include in the `PreRun` execution of Cobra.
I am going on my logic that previous to this change the init of flags is executed before any of the cobra methods are run.
So just to keep the order and separate a "required" flag validation, I think `PreRun`, would be the right spot.

chentex (approved):
/lgtm

cmd/kube-burner/kube-burner.go
Sandeepyadav93: Thank you, https://github.com/spf13/cobra/blob/6dec1ae26659a130bdb4c985768d1853b0e1bc06/site/content/user_guide.md?plain=1#L620-L628 confirms PreRun executes before Run.

I have moved the code to assign a random uuid if not defined in PreRun, worked locally in my testing.

PR #910, opened 2025-06-23
Added Object based deletion strategy

### LINKED ISSUE #908: [RFE] Object based deletion strategy
Yes, We need a object based deletion strategy for all the objects that are created as a part of kube-burner workloads. Previously we used to have this only for churn: https://github.com/kube-burner/kube-burner/pull/492/files, but having it for the entire execution level would be more useful.

To have a global flag that cleans up object either based on the `default` or `gvr` strategy. Because even if some one wants to delete objects with `gvr` strategy in churn would sure be interested in having the same strategy applied to the entire benchmark run instead of having one strategy for churn and another for the rest of the workload.

Having one flag to control benchmark level deletion strategy and keep the existing `chrunDeletionStrategy` flag as is. But I think having deletion strategy same across the benchmark including churn makes more sense.

**Additional context**
More context: https://redhat-internal.slack.com/archives/C0780C0J6HJ/p1750162586836599

niladrix719:
## Type of change

- Refactor
- New feature
- Documentation

## Description

- Added Object based deletion strategy `deletionStrategy`
- Removed `churnDeletionStrategy`
- Updated documentation

- Related Issue #
- Closes #908


pkg/burner/executor.go
```diff
@@ -52,6 +52,7 @@ type Executor struct {
 	kubeVirtClient    kubecli.KubevirtClient
 	functionTemplates []string
 	embedCfg          *fileutils.EmbedConfiguration
+	DeletionStrategy  string
```
ygalblum: I don't think you need to expose this field
```suggestion
	deletionStrategy  string
```

pkg/burner/job.go
```diff
@@ -405,16 +406,29 @@ func garbageCollectJob(ctx context.Context, jobExecutor Executor, labelSelector
...
-	err := util.CleanupNamespaces(ctx, jobExecutor.clientSet, labelSelector)
-	// Just report error and continue
-	if err != nil {
-		log.Error(err.Error())
+	if jobExecutor.DeletionStrategy == config.GVRDeletionStrategy {
+		namespaces, err := jobExecutor.clientSet.CoreV1().Namespaces().List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
+		if err != nil {
+			log.Error(err.Error())
+		} else {
+			var namespacesToDelete []string
```
ygalblum:
```suggestion
			namespacesToDelete := make([]string, 0, len(namespaces.Items)
```

pkg/burner/job.go
```diff
@@ -405,16 +406,29 @@ func garbageCollectJob(ctx context.Context, jobExecutor Executor, labelSelector
...
+		if err != nil {
+			log.Error(err.Error())
+		}
 	}
 	for _, obj := range jobExecutor.objects {
 		jobExecutor.limiter.Wait(ctx)
 		if !obj.namespaced {
 			CleanupNonNamespacedResourcesUsingGVR(ctx, jobExecutor, obj, labelSelector)
-		} else if obj.namespace != "" { // When the object has a fixed namespace not generated by kube-burner
+		} else if obj.namespace != "" && jobExecutor.DeletionStrategy != config.GVRDeletionStrategy { // When the object has a fixed namespace not generated by kube-burner
```
ygalblum: Not sure I understand this addition. Can you please explain?

niladrix719: Ah, I think that was a wrong assumption on my part

niladrix719:
Update: Ordered the `indexByName` entries in `test/grafana/test-dashboard.json`

vishnuchalla:
Thanks for this change. @niladrix719 can you fix the conflicts please?

ygalblum (approved):
LGTM

vishnuchalla (approved):
lgtm

vishnuchalla:
@rsevilla87 do you wanna take a look as well?

rsevilla87 (commented):
If I've understood correctly, when using a gvr strategy, when the workload finishes it will leave empty all the created namespaces, is that what we want?, shouldn't we delete the namespaces as well once they are emptied? wdyt @vishnuchalla @niladrix719 ?

ygalblum:
That's what the RFE in question is about, no?

rsevilla87:
> That's what the RFE in question is about, no?

If I understand correctly, the RFE is about using the same deletion strategy for garbage collection as is used for churning, correct, but we have to keep that one of the garbage collection goals is to left the cluster in its original state, including namespaces created by the workloads.

ygalblum:
Yes and no. If that's not the default behaviour and the user is aware of their choice, then maybe it's ok. The question is what's the point? Will users really want to delete all resources and leave the namespace?
BTW how does this affect cases in which the namespace was pre-created by the user? Maybe it's beneficial in these cases?

vishnuchalla:
> > That's what the RFE in question is about, no?
> 
> If I understand correctly, the RFE is about using the same deletion strategy for garbage collection as is used for churning, correct, but we have to keep that one of the garbage collection goals is to left the cluster in its original state, including namespaces created by the workloads.

Yeah once all the resources are deleted using gvr, its better to cleanup the empty namespaces as well as they are part of kube-burner too. wdyt @niladrix719 ?

ygalblum:
Please rebase on main once https://github.com/kube-burner/kube-burner/pull/935 is merged

PR #912, opened 2025-06-23
Added achieved QPS to jobSummary report

### LINKED ISSUE #660: [RFE] Log and report achieved job's QPS
When a job finishes, report the achieved average number of request per second observed during the benchmark. This number can be indexed along with the jobSummary document

Multiple ways to calculate this number, one of them can be to summarize the total amount of objects created/deleted/patched and divide them by the number of seconds the job took

**Additional context**
This feature can be useful in high latency scenarios in which the user is not sure about the own client's performance (kube-burner) or benchmarks using a high number of QPS/Burst values.

niladrix719:
## Type of change

- New feature

## Description

Added achieved QPS to jobSummary report

- Related Issue #
- Closes #660


pkg/burner/job.go
```diff
@@ -62,6 +63,8 @@ var (
 		config.ExecutionModeParallel:   {},
 		config.ExecutionModeSequential: {},
 	}
+
+	objectOperations int32 = 0
```
ygalblum: IIUC, the calculation is per job. But, `objectOperations` is never reset to 0. So the information is incorrect.
Instead of resetting it, add a field to the `Executor`. This way you will start the calculation for each job. 
Then also add it to the `prometheus.Job` and copy it there at the end of the job before the metrics are calculated.
In any case, avoid using global variables

ygalblum (changes requested):
Shouldn't we also take into account calls in `read` and `kubevirt` job types?

PR #916, opened 2025-06-25
feat(build): add missing security hardening flags for binary releases

### LINKED ISSUE #851: [RFE] security hardening for binary releases

pfeifferj:
## Type of change

- Optimization

## Description

following up on #880 I found a way to add the missing security hardening features elegantly:

**1. FULL RELRO (relocation read-only)**
**2. position independent executable (PIE)**
**3. binary stripping**
**4. reproducible builds**

- updated `.goreleaser.yml` to split builds into linux (with full hardening) and other platforms
- added `build-hardened` target to Makefile for local hardened builds
- maintained backward compatibility with existing `build` and `build-release` targets
- security flags only applied to Linux builds as other platforms have different requirements

Component: build

- Closes #851

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 
-.PHONY: build lint clean test help images push manifest manifest-build all
+.PHONY: build build-release build-hardened build-hardened-cgo lint clean test help images push manifest manifest-build all
 
 
 ARCH ?= $(shell uname -m | sed s/aarch64/arm64/ | sed s/x86_64/amd64/)
@@ -8,6 +8,13 @@ BIN_DIR = bin
 BIN_PATH = $(BIN_DIR)/$(ARCH)/$(BIN_NAME)
 CGO = 0
 TEST_BINARY ?= $(CURDIR)/$(BIN_PATH)
+# Security hardening flags for static builds
+HARDENED_FLAGS = -buildmode=pie -tags=netgo,osusergo,static_build
+HARDENED_LDFLAGS = -extldflags '-static'
+
+# Full security hardening flags with CGO
+HARDENED_CGO_FLAGS = -buildmode=pie
+HARDENED_CGO_LDFLAGS = -linkmode=external -extldflags "-Wl,-z,relro,-z,now"
 
 GIT_COMMIT = $(shell git rev-parse HEAD)
 VERSION ?= $(shell hack/tag_name.sh)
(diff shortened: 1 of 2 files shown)


ygalblum (approved):
LGTM

.goreleaser.yml
```diff
@@ -3,22 +3,48 @@
 before:
   hooks:
 builds:
-  - env:
-      - CGO_ENABLED=0
+  # Linux builds with security hardening
+  - id: linux-secure
```
rsevilla87: Compiling linux with CGO_ENABLED=1 produces a dynamically linked binary, that can have portability problems.

pfeifferj:
good point @rsevilla87 I'll do some more research into best practices here. I think golang 1.23+ might also support some of these features with `CGO_ENABLED=0`

pfeifferj:
ok I dug through the docs and it seems like it's really an either or decision:

* CGO_ENABLED=1 with security hardening flags =>  better security but portability issues
* CGO_ENABLED=0  => maximum portability but can't achieve full security hardening

I guess the tradeoff would again be offering a make target for this and let people compile from source if they care about these flags

vishnuchalla:
lgtm

vishnuchalla (approved):
lgtm

PR #917, opened 2025-07-01
Added achieved QPS to jobSummary report

### LINKED ISSUE #660: [RFE] Log and report achieved job's QPS

niladrix719:
## Type of change

- New feature

## Description

Added achieved QPS to jobSummary report

- Related Issue #
- Closes #660


vishnuchalla:
Thanks for this change. Calculation looks good to me, @rsevilla87 @ygalblum can you also take a look?

ygalblum (dismissed):
LGTM

pkg/burner/delete.go
```diff
@@ -59,6 +60,8 @@ func deleteHandler(ex *JobExecutor, obj *object, item unstructured.Unstructured,
 	}
 	if err != nil {
 		log.Errorf("Error found removing %s/%s: %s", item.GetKind(), item.GetName(), err)
+	} else {
+		atomic.AddInt32(&ex.objectOperations, 1)
```
rsevilla87: We should also count errors as requests, I think this else condition shouldn't exist.

pkg/burner/read.go
```diff
@@ -51,5 +52,7 @@ func readHandler(ex *JobExecutor, obj *object, item unstructured.Unstructured, i
 	}
 	if err != nil {
 		log.Errorf("Error found reading %s/%s: %s", item.GetKind(), item.GetName(), err)
+	} else {
+		atomic.AddInt32(&ex.objectOperations, 1)
```
rsevilla87: Same as above, failed requests should be counted as well IMO

pkg/burner/patch.go
```diff
@@ -95,6 +96,7 @@ func patchHandler(ex *JobExecutor, obj *object, originalItem unstructured.Unstru
 				originalItem.GetName(), ns, err)
 		}
 	} else {
+		atomic.AddInt32(&ex.objectOperations, 1)
```
rsevilla87: Same

pkg/burner/kubevirt.go
```diff
@@ -187,6 +188,7 @@ func kubeOpHandler(ex *JobExecutor, obj *object, item unstructured.Unstructured,
 	if err != nil {
 		log.Errorf("Failed to execute op [%s] on the VM [%s]: %v", obj.KubeVirtOp, item.GetName(), err)
 	} else {
+		atomic.AddInt32(&ex.objectOperations, 1)
```
rsevilla87: Same

pkg/burner/delete.go
vishnuchalla: @rsevilla87 Can we consider errors being counted as `achievedQPS` as well 🤔 ? Wasn't it meant to see only the requests that successfully went through?

rsevilla87: An API error also consumes 1 QPS

vishnuchalla (commented):
lgtm

rsevilla87 (approved):
lgtm

PR #922, opened 2025-07-05
feat: Add support for creating CRDs and their CRs in same workflow

### LINKED ISSUE #863: [RFE] Ability to create CRs of a CRD that does not yet exist on the cluster until the job is running

With kube-burner it is not possible to create CRDs dynamically and then create the CRs.

Example Job:

```yaml
---
global:
  gc: false

jobs:
- name: test-crd-cr
  namespace: test
  jobType: create
  jobIterations: 2
  namespacedIterations: true
  cleanup: false
  podWait: false
  qps: 10
  burst: 10
  objects:
  - objectTemplate: crd.yml
    replicas: 1

  - objectTemplate: cr0.yml
    replicas: 1

  - objectTemplate: cr1.yml
    replicas: 1
```

Example crd.yml
```yaml
---
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: testcrds{{.Iteration}}.testcrd.example.com
spec:
  group: testcrd.example.com
  versions:
    - name: v1
      served: true
      storage: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                workload:
                  type: string
                iterations:
                  type: integer
  scope: Namespaced
  names:
    plural: testcrds{{.Iteration}}
    singular: testcrd{{.Iteration}}
    kind: TestCRD{{.Iteration}}
    shortNames:
    - tcrd{{.Iteration}}
```

Example cr0.yml and cr1.yml
```yaml
---
apiVersion: testcrd.example.com/v1
kind: TestCRD0
metadata:
  name: test{{.Replica}}
spec:
  workload: {{randAlpha 5}}
  iterations: {{.Iteration}}

```

```yaml
---
apiVersion: testcrd.example.com/v1
kind: TestCRD1
metadata:
  name: test{{.Replica}}
spec:
  workload: {{randAlpha 5}}
  iterations: {{.Iteration}}

```

And when it is ran:
```console
$ kube-burner init -c job-crd-cr.yml 
time="2025-05-19 09:30:43" level=info msg="🔥 Starting kube-burner (1.16.0@ee79f622d88168e20fc2cd14fd7bc12d64c998f0) with UUID 1b07215f-a7e5-4000-bd46-ceb993752cef" file="job.go:84"
time="2025-05-19 09:30:43" level=info msg="QPS: 10" file="job.go:340"
time="2025-05-19 09:30:43" level=info msg="Burst: 10" file="job.go:341"
time="2025-05-19 09:30:44" level=info msg="Job test-crd-cr: 2 iterations with 1 CustomResourceDefinition replicas" file="create.go:78"
time="2025-05-19 09:30:44" level=fatal msg="no matches for kind \"TestCRD0\" in group \"testcrd.example.com\"" file="create.go:64"
```

The ability for kube-burner to recognize when a job creates the CRD that the cr yaml files are defined from.

The current alternative here is to run two separate jobs, one in which CRDs are created on a cluster prior to the CRs being created.

niladrix719:
## Type of change

- New feature
- Bug fix

## Description

- Objects whose mapping fails initially, retrying after object creation allows CRDs to register before validating their CRs 
- `DeferredDiscoveryRESTMapper` enables dynamic discovery of CRDs during runtime

- Closes #863


pkg/burner/utils.go
```diff
@@ -111,6 +113,21 @@ func yamlToUnstructured(fileName string, y []byte, uns *unstructured.Unstructure
 	return o, gvk
 }
 
+// validateObject resets the REST mapper and resolves the object's fields before creation
+func (ex *JobExecutor) validateObject(obj *object) {
```
ygalblum: This method doesn't really validate the object. Please rename it

pkg/burner/create.go
```diff
@@ -121,17 +118,19 @@ func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterati
...
 		}
 		for objectIndex, obj := range ex.objects {
+			if obj.gvr == (schema.GroupVersionResource{}) {
+				ex.validateObject(obj)
+				if ex.nsRequired {
+					if !ex.NamespacedIterations {
+						ns = ex.createNamespace(ex.Namespace, nsLabels, nsAnnotations, waitListNamespaces, namespacesCreated)
+					} else {
+						ns = ex.createNamespace(ex.generateNamespace(i), nsLabels, nsAnnotations, waitListNamespaces, namespacesCreated)
+					}
```
ygalblum:
```suggestion
                    nsName := ex.Namespace
					if ex.NamespacedIterations {
					    nsName = ex.generateNamespace(i)
					}
					ns = ex.createNamespace(ex.generateNamespace(i), nsLabels, nsAnnotations, waitListNamespaces, namespacesCreated)
```

pkg/burner/create.go
```diff
@@ -66,15 +70,13 @@ func (ex *JobExecutor) setupCreateJob(mapper meta.RESTMapper) {
...
-		if err != nil {
-			log.Fatal(err)
-		}
-		obj := &object{
-			gvr:        mapping.Resource,
-			objectSpec: t,
-			Object:     o,
-			namespace:  uns.GetNamespace(),
-			namespaced: mapping.Scope.Name() == meta.RESTScopeNameNamespace,
+		obj.namespace = uns.GetNamespace()
```
ygalblum:
```suggestion
        obj := &object{
            objectSpec: t,
            Object: o,
		    namespace = uns.GetNamespace()
        }
```
And remove the creation of `obj` before

pkg/burner/executor.go
```diff
@@ -73,19 +74,19 @@ func newExecutor(configSpec config.Spec, kubeClientProvider *config.KubeClientPr
 	ex.dynamicClient = dynamic.NewForConfigOrDie(ex.restConfig)
 
 	_, setupRestConfig := kubeClientProvider.ClientSet(100, 100) // Hardcoded QPS/Burst
-	mapper := newRESTMapper(discovery.NewDiscoveryClientForConfigOrDie(setupRestConfig))
+	ex.mapper = newRESTMapper(setupRestConfig)
 
 	switch job.JobType {
 	case config.CreationJob:
-		ex.setupCreateJob(mapper)
+		ex.setupCreateJob(ex.mapper)
```
ygalblum: If the `mapper` is now part of the `ex`, then you don't need pass it to the method, it can access it directly

pkg/burner/utils.go
```diff
@@ -111,6 +113,21 @@ func yamlToUnstructured(fileName string, y []byte, uns *unstructured.Unstructure
...
+	ex.mapper.(*restmapper.DeferredDiscoveryRESTMapper).Reset()
```
ygalblum: Since the `mapper` belongs to the `ex` it should know its type and not have to cast it at runtime

pkg/burner/create.go
```diff
@@ -121,17 +118,19 @@ func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterati
...
-				}
-				namespacesCreated[ns] = true
-				*waitListNamespaces = append(*waitListNamespaces, ns)
-			}
+			ns = ex.createNamespace(ex.generateNamespace(i), nsLabels, nsAnnotations, waitListNamespaces, namespacesCreated)
 		}
 		for objectIndex, obj := range ex.objects {
+			if obj.gvr == (schema.GroupVersionResource{}) {
+				ex.validateObject(obj)
+				if ex.nsRequired {
```
ygalblum, line 127: Please add a comment explaining the `ex.nsRequired` may change in `validateObject`. I had to look into what chnages

ygalblum (approved):
LGTM

pkg/burner/utils.go
```diff
@@ -111,6 +113,21 @@ func yamlToUnstructured(fileName string, y []byte, uns *unstructured.Unstructure
 	return o, gvk
 }
 
+// resolveObjectMapping resets the REST mapper and resolves the object's resource mapping and namespace requirements
+func (ex *JobExecutor) resolveObjectMapping(obj *object) {
+	ex.mapper.Reset()
+	mapping, err := ex.mapper.RESTMapping(obj.gvk.GroupKind())
```
rsevilla87, line 119: Just for the record, IIRC this operation performs several API calls to identify the resource mapping of the provide GK,  that can introduce an undesired delay on the operation

rsevilla87: operation == RestMapping()

pkg/burner/create.go
```diff
@@ -121,17 +118,20 @@ func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterati
...
+			if obj.gvr == (schema.GroupVersionResource{}) {
+				// resolveObjectMapping may set ex.nsRequired to true if the object is namespaced but doesn't have a namespace specified
+				ex.resolveObjectMapping(obj)
+				if ex.nsRequired {
+					nsName := ex.Namespace
+					if ex.NamespacedIterations {
+						nsName = ex.generateNamespace(i)
+					}
+					ns = ex.createNamespace(nsName, nsLabels, nsAnnotations, waitListNamespaces, namespacesCreated)
+				}
```
rsevilla87, line 133: Why are you creating the namespace again?, namespace should be already either in line 106 or 121

rsevilla87: The information that kube-burner was missing to create this kind of "dynamic" object was the gvr, which is set now by the resolveObjectMapping method

rsevilla87: wdyt @ygalblum ?

rsevilla87 (changes requested):
Hey @niladrix719 , can you also add some test(s) to cover this functionality?

pkg/burner/create.go
niladrix719: I think the namespace creation here is necessary because objects with missing GVR also don't get their `namespaced` field set to true during the initial setup phase due to missing `mapping`:
https://github.com/kube-burner/kube-burner/blob/20b7922a05bde70ba74a85dd38ce7195f49f1cbb/pkg/burner/create.go#L74-L80
Thus, the `nsRequired` field is also not set to true here:
https://github.com/kube-burner/kube-burner/blob/20b7922a05bde70ba74a85dd38ce7195f49f1cbb/pkg/burner/create.go#L83-L85
So the checks before lines 106 or 121 are not satisfied

test/k8s/kube-burner-cr-crd.yml
rsevilla87, line 1: This is not what we need to test, we need to test a single job creating the two objects,  a crd and then a cr associated with the previous crd.

rsevilla87: Similar to what is reported in the original issue https://github.com/kube-burner/kube-burner/issues/863#issue-3073882296

rsevilla87:
Hey @niladrix719, I've raised a question on the original issue https://github.com/kube-burner/kube-burner/issues/863#issuecomment-3154625748, let's wait for the reporter's answer before continuing working on this PR, thanks

pkg/burner/create.go
ygalblum: Yes, this is the reason I asked for the comment above

ygalblum:
@niladrix719 also see that you need to rebase on main and resolve the conflict

pkg/burner/create.go
```diff
@@ -121,17 +118,20 @@ func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterati
...
-			}
+			ns = ex.createNamespace(ex.generateNamespace(i), nsLabels, nsAnnotations, waitListNamespaces, namespacesCreated)
 		}
 		for objectIndex, obj := range ex.objects {
+			if obj.gvr == (schema.GroupVersionResource{}) {
+				// resolveObjectMapping may set ex.nsRequired to true if the object is namespaced but doesn't have a namespace specified
+				ex.resolveObjectMapping(obj)
+				if ex.nsRequired {
+					nsName := ex.Namespace
+					if ex.NamespacedIterations {
```
vishnuchalla, line 129: looks like this logic might get inline with pre-creating namespaces logic that we have in the other https://github.com/kube-burner/kube-burner/pull/898

pkg/burner/executor.go
```diff
@@ -52,6 +52,7 @@ type JobExecutor struct {
 	kubeVirtClient    kubecli.KubevirtClient
 	functionTemplates []string
 	embedCfg          *fileutils.EmbedConfiguration
+	mapper            *restmapper.DeferredDiscoveryRESTMapper
```
vishnuchalla, line 55: Interesting detail.

vishnuchalla:
@rsevilla87 @ygalblum seems like good to go. wdyt?

ygalblum:
@niladrix719 looking at the test you added, I understand that the CRD and the CRs are created in separate jobs. Is that a limitation of this solution? If not, please update the test to make sure that the CRs can be created in the same job as the CRD as a different object but probably with order.
Regardless to which mode is supported, I think we should explain it in the documentation

vishnuchalla:
Closing in favor of https://github.com/kube-burner/kube-burner/pull/999

PR #927, opened 2025-07-11
Add support for VirtualMachineInstanceMigration latency

ygalblum:
## Type of change

- New feature
- Optimization
- Documentation

## Description
Allow monitoring live migration via the Virtual Machine Instance Migration resource
Replace the call to KubeVirt Migrate with creating the resource to allow controlling its labels

### DIFF
--- a/docs/measurements/index.md
+++ b/docs/measurements/index.md
@@ -697,6 +697,161 @@ Where `quantileName` matches with the pvc phases and can be:
 
 And the metrics, error rates, and their thresholds work the same way as in the other latency measurements.
 
+## VirtualMachineInstanceMigration Latency
+
+Collects latencies from different VirtualMachineInstanceMigration phases on the cluster, these **latency metrics are in ms**. It can be enabled with:
+
+```yaml
+  measurements:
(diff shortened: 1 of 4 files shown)


docs/measurements/index.md
```diff
@@ -697,6 +697,161 @@ Where `quantileName` matches with the pvc phases and can be:
 
 And the metrics, error rates, and their thresholds work the same way as in the other latency measurements.
 
+## VirtualMachineInstanceMigration Latency
```
vishnuchalla, line 700: I guess, adding tests to this measurement is a bit complicated?

ygalblum: Adding the tests is simple. The problem is that since the CI is running on a single node cluster, migration never really happens. So, there will not be any data to show:
https://github.com/kube-burner/kube-burner/blob/476df7c040f553c2d324c7ddf2bf43a7196688db/test/k8s/kube-burner-virt-operations.yml#L71

vishnuchalla: Understood.

PR #928, opened 2025-07-13
feat: add --overlay flag support to modify base configurations

### LINKED ISSUE #924: [RFE] Add support for configuration overlays to modify base configurations

Currently, kube-burner requires separate configuration files for different test scenarios, leading to duplication and maintenance overhead. This enhancement would add overlay support similar to Kustomize, allowing users to define base configurations and apply targeted modifications through overlay files

Add support for overlay files that can modify base configurations, similar to Kustomize. Users should be able to run `kube-burner run -c base.yaml --overlay overlays/high-load.yaml` to merge configurations, reducing duplication across different test environments

niladrix719:
## Type of change

- New feature

## Description

Adds a new `--overlay` CLI flag that enables users to merge overlay configurations into a base setup, allowing for flexible customization and environment-specific modifications

- Related Issue #
- Closes #924

### DIFF
--- a/test/k8s/kube-burner-base.yml
+++ b/test/k8s/kube-burner-base.yml
@@ -8,6 +8,7 @@ global:
 jobs:
   - name: overlay
     jobIterations: 1
+    namespacedIterations: false
     podWait: false
     cleanup: true
     objects:


cmd/kube-burner/kube-burner.go
```diff
@@ -108,9 +109,18 @@ func initCmd() *cobra.Command {
 			util.SetupFileLogging(uuid)
 			kubeClientProvider := config.NewKubeClientProvider(kubeConfig, kubeContext)
 			clientSet, _ = kubeClientProvider.DefaultClientSet()
-			configFileReader, err := fileutils.GetWorkloadReader(configFile, nil)
-			if err != nil {
-				log.Fatalf("Error reading configuration file %s: %s", configFile, err)
+			var configFileReader io.Reader
+			if len(overlayFiles) > 0 {
```
ygalblum: The separation between with or without overlay is an implementation detail and should not bother the caller. There should be a single reader object that gets the configFile and a list (that can be nil or empty) or overlayFiles that handles the read requests.

pkg/util/overlay.go
```diff
@@ -0,0 +1,234 @@
...
+	for _, overlayPath := range o.OverlayPaths {
+		overlayReader, err := fileutils.GetWorkloadReader(overlayPath, o.EmbedCfg)
+		if err != nil {
+			return 0, fmt.Errorf("failed to read overlay %s: %w", overlayPath, err)
+		}
+		overlayContent, err := io.ReadAll(overlayReader)
+		if err != nil {
+			return 0, fmt.Errorf("failed to read overlay %s: %w", overlayPath, err)
+		}
+		renderedOverlayContent, err := RenderTemplate(overlayContent, EnvToMap(), MissingKeyZero, []string{})
```
ygalblum: kube-burner allows the user to pass a YAML file with values (using the `--user-data` flag). How will it be reflected here? I think you need to pass it to the reader object. See https://github.com/kube-burner/kube-burner/blob/a2da6c4ea60bdedba4666d670461b70418131be8/pkg/config/config.go#L135

pkg/util/overlay.go
```diff
@@ -0,0 +1,234 @@
...
+	return &OverlayReader{
+		BaseReader:   baseReader,
+		OverlayPaths: overlayPaths,
+		EmbedCfg:     embedCfg,
+	}, nil
+}
+
+// mergeYAML merges two YAML documents, applying the specified merge options
+func mergeYAML(baseYAML, overlayYAML []byte) ([]byte, error) {
+	var baseMap map[string]interface{}
```
ygalblum:
```suggestion
	var baseMap map[string]any
```
Here and in other places

pkg/util/overlay.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	return mergedYAML, nil
+}
+
+// deepMergeMap recursively merges two maps, applying merge options
+func deepMergeMap(base, overlay map[string]interface{}, currentPath []string) map[string]interface{} {
+	result := make(map[string]interface{})
+	for k, v := range base {
+		result[k] = v
+	}
```
ygalblum:
```suggestion
	maps.Copy(result, base)
```

pkg/util/overlay.go
```diff
@@ -0,0 +1,234 @@
...
+				if overlayMap, ok := v.(map[string]interface{}); ok {
+					result[k] = deepMergeMap(baseMap, overlayMap, newPath)
+				} else {
+					result[k] = v
+				}
+			} else {
+				result[k] = v
+			}
+		} else {
+			result[k] = v
```
ygalblum: Is there a cleaner way to do this (maybe with continue, or something like it)?

pkg/util/overlay.go
```diff
@@ -0,0 +1,234 @@
...
+
+type MergeOption struct {
+	Path     []string
+	MergeKey string
+}
+
+var mergeOptions = []MergeOption{
+	{
+		Path:     []string{"jobs"},
+		MergeKey: "name",
```
ygalblum: I don't understand the `MergeKey`. Does this mean that users can only override what's defined here? If yes, the doc is not clear about it. If not, then what is it?

pkg/util/overlay.go
```diff
@@ -0,0 +1,234 @@
...
+	overlayArray, ok := overlay.([]interface{})
+	if !ok {
+		return overlay
+	}
+	baseItemsMap := make(map[string]map[string]interface{})
+	baseOrder := make([]string, 0)
+	for _, item := range baseArray {
+		if itemMap, ok := item.(map[string]interface{}); ok {
+			keyValue := ""
+			if mergeKey != "" {
```
ygalblum: IIUC if `mergeKey == ""`, then `keyValue` will remain `""` and then `baseItemsMap` and `baseOrder` are never changed. If I'm right, then should the code just skip the loop?
If there any point in calling this function with `mergeKey == ""`?

pkg/util/overlay.go
```diff
@@ -0,0 +1,234 @@
...
+					}
+				}
+			}
+			if keyValue != "" {
+				baseItemsMap[keyValue] = itemMap
+				baseOrder = append(baseOrder, keyValue)
+			}
+		}
+	}
+	for _, overlayItem := range overlayArray {
```
ygalblum: Same question as the previous loop

pkg/util/overlay.go
niladrix719: The `MergeKey` doesn't limit what we can override; it enables nested merging based on a particular key instead of just replacing the entire array

pkg/util/reader/reader.go
```diff
@@ -0,0 +1,87 @@
...
+	UserDataFileReader io.Reader
+	AllowMissingKeys   bool
+	AdditionalVars     map[string]any
+	processedContent   []byte
+	readOffset         int
+}
+
+// NewReader creates a new Reader with the given base and overlay paths
+func NewReader(baseLocation string, overlayFilePaths []string, embedCfg *fileutils.EmbedConfiguration, userDataFileReader io.Reader, allowMissingKeys bool, additionalVars map[string]any) (*Reader, error) {
+	baseReader, err := fileutils.GetWorkloadReader(baseLocation, embedCfg)
```
ygalblum, line 26: Why do you get the `BaseReader` for the `baseLocation` but not for the `overlayFilePaths`?

pkg/util/reader/reader.go
```diff
@@ -0,0 +1,87 @@
...
+		EmbedCfg:           embedCfg,
+		UserDataFileReader: userDataFileReader,
+		AllowMissingKeys:   allowMissingKeys,
+		AdditionalVars:     additionalVars,
+	}, nil
+}
+
+// Read reads data from the base configuration and applies overlays
+func (o *Reader) Read(p []byte) (n int, err error) {
+	if o.processedContent != nil {
```
ygalblum, line 42: Since you're using memory, should we protect this from multi-threading? Currently, it seems that it's called only from the main thread, but, maybe it's better to be future-ready

ygalblum (commented):
Sorry for the delay. I've just notices I did not submit my comments

vishnuchalla:
Closing this as stale.

PR #933, opened 2025-07-18
feat: add multi-shell completion support (bash, zsh, fish, powershell) and enhance error messaging

mdxabu:
## Type of change

- Enhancement

## Description

I've made two meaningful improvements to the kube-burner CLI tool,
- Enhanced Shell Completion Support
   - Upgraded the completion command from bash-only to support multiple shells
   - Before only supported bash completion with limited usage instructions
   - Now supports bash, zsh, fish, and PowerShell with comprehensive documentation for each shell
   
These improvements enhance the user experience and make kube-burner more accessible to developers using different shell environments, which is especially important for a tool used across diverse development setups in the cloud-native ecosystem.

### DIFF
--- a/cmd/kube-burner/kube-burner.go
+++ b/cmd/kube-burner/kube-burner.go
@@ -57,18 +57,58 @@ Tool aimed at stressing a kubernetes cluster by creating or deleting lots of obj
 }
 
 var completionCmd = &cobra.Command{
-	Use:   "completion",
-	Short: "Generates completion scripts for bash shell",
-	Long: `To load completion in the current shell run
-. <(kube-burner completion)
+	Use:   "completion [bash|zsh|fish|powershell]",
+	Short: "Generate completion script for your shell",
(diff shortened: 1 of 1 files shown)


cmd/kube-burner/kube-burner.go
```diff
@@ -110,13 +150,13 @@ func initCmd() *cobra.Command {
 			clientSet, _ = kubeClientProvider.DefaultClientSet()
 			configFileReader, err := fileutils.GetWorkloadReader(configFile, nil)
 			if err != nil {
-				log.Fatalf("Error reading configuration file %s: %s", configFile, err)
+				log.Fatalf("Error reading configuration file %s: %s\nPlease ensure the file exists and is accessible", configFile, err)
```
ygalblum, line 153: I don't like multiline log messages even if they are the last lines. Why is this addition needed? How does it relate to this PR?

PR #940, opened 2025-08-01
feat(pod_latency): Add high-precision metrics support for pod latency…

### LINKED ISSUE #930: How to get podLatency in ms
### What happened
I enabled the built-in `podLatency` metrics collection expecting millisecond-level granularity as described in the documentation:

> *"Collects latencies from the different pod startup phases, these latency metrics are in **ms**."*

However, it seems that kube-burner get the latencies from the status in pod, these fields don't contain milliseconds. 

### Environment
- Kubernetes 1.25.0  
- containerd 1.7.10  
- kube-burner v1.16.4 

### How to reproduce

Example pod status that kube-burner uses to compute latencies:

```yaml
status:
  conditions:
  - type: PodScheduled
    lastTransitionTime: "2025-07-11T07:25:38Z"
  - type: Initialized
    lastTransitionTime: "2025-07-11T07:26:59Z"
  startTime: "2025-07-11T07:25:29Z"
  containerStatuses:
  - state:
      running:
        startedAt: "2025-07-11T07:26:59Z"
```

All relevant fields (`startTime`, `PodScheduled`, `Initialized`) are RFC3339 with **second** precision.

### Questions
Is there an additional configuration flag or feature gate that makes kube-burner collect higher-resolution timestamps?  

Thanks!

devsubid:
# Add High-Precision Pod Latency Measurements

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation
- [ ] CI

## Description

This PR introduces **high-precision pod latency measurements** to address timing precision limitations in Kubernetes API timestamps.

### Problem

Current pod latency measurements suffer from second-only precision due to Kubernetes API constraints:

- API server only supports RFC3339 format (second precision), not RFC3339NANO
- Kubelet purposely discards microseconds/nanoseconds due to API limitations
- Fast pod operations (< 1 second) result in zero or negative latencies
- Sub-second pod startup scenarios cannot be measured accurately

### Solution

Implements client-side timestamp capture using `time.Now().UTC()` with nanosecond precision when kube-burner observes pod events in real-time.

### Key Features

- **Dual measurement system**: Provides both standard (API-based) and high-precision (client-based) metrics
- **Sub-millisecond accuracy**: Float64 values in milliseconds with decimal precision (e.g., `123.456ms`)
- **Client-side timestamps**: Captures exact moment events are observed, not API timestamps
- **Backward compatibility**: Existing measurements remain unchanged

### Configuration

```yaml
global:
  measurements:
    - name: podLatency
      highPrecisionMetrics: true # Enable high-precision mode
```

### Output Metrics

When enabled, adds high-precision metrics with "Client" prefix:

- `ClientPodScheduled`: 123.456ms
- `ClientPodReady`: 456.789ms
- `ClientPodInitialized`: 234.567ms
- `ClientContainersReady`: 345.678ms

### Implementation

- Added `HighPrecisionMetrics bool` field to Measurement struct
- Enhanced `podMetric` with client-side timestamp and latency fields
- Updated event handlers to capture client-side timestamps
- Modified metric calculation to include high-precision measurements
- Uses nanosecond resolution converted to milliseconds with float64

### Benefits

- Accurate SLO/SLA tracking for pod startup performance
- Better bottleneck identification for sub-second operations
- Improved granularity with microsecond-level timing
- Eliminates negative latency measurement errors
- Enhanced Kubernetes cluster performance observability

- Related Issue: Pod latency measurements limited to second precision
- Addresses: Inaccurate measurements due to Kubernetes API timestamp constraints
- Closes #930

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -79,7 +79,7 @@ require (
 	github.com/x448/float16 v0.8.4 // indirect
 	golang.org/x/crypto v0.36.0 // indirect
 	golang.org/x/net v0.38.0 // indirect
-	golang.org/x/oauth2 v0.26.0 // indirect
+	golang.org/x/oauth2 v0.27.0 // indirect
 	golang.org/x/sys v0.31.0 // indirect
 	golang.org/x/term v0.30.0 // indirect
 	golang.org/x/text v0.23.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -573,8 +573,8 @@ golang.org/x/oauth2 v0.0.0-20190226205417-e64efc72b421/go.mod h1:gOpvHmFTYa4Iltr
 golang.org/x/oauth2 v0.0.0-20190604053449-0f29369cfe45/go.mod h1:gOpvHmFTYa4IltrdGE7lF6nIHvwfUNPOp7c8zoXwtLw=
 golang.org/x/oauth2 v0.0.0-20191202225959-858c2ad4c8b6/go.mod h1:gOpvHmFTYa4IltrdGE7lF6nIHvwfUNPOp7c8zoXwtLw=
 golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d/go.mod h1:gOpvHmFTYa4IltrdGE7lF6nIHvwfUNPOp7c8zoXwtLw=
-golang.org/x/oauth2 v0.26.0 h1:afQXWNNaeC4nvZ0Ed9XvCCzXM6UHJG7iCg0W4fPqSBE=
-golang.org/x/oauth2 v0.26.0/go.mod h1:XYTD2NtWslqkgxebSiOHnXEap4TF09sJSc7H1sXbhtI=
+golang.org/x/oauth2 v0.27.0 h1:da9Vo7/tDv5RH/7nZDz1eMGS/q1Vv1N/7FCrBhI9I3M=
+golang.org/x/oauth2 v0.27.0/go.mod h1:onh5ek6nERTohokkhCD/y2cV4Do3fxFHFuAejCkRWT8=
 golang.org/x/sync v0.0.0-20180314180146-1d60e4601c6f/go.mod h1:RxMgew5VJxzue5/jJTE5uejpjVlOe/izrB70Jof72aM=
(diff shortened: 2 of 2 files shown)


pkg/measurements/pod_latency.go
```diff
@@ -112,7 +128,14 @@ func (p *podLatency) handleCreatePod(obj any) {
 		Metadata:     p.Metadata,
 		JobIteration: getIntFromLabels(podLabels, config.KubeBurnerLabelJobIteration),
 		Replica:      getIntFromLabels(podLabels, config.KubeBurnerLabelReplica),
-	})
+	}
+
+	if p.highPrecisionMetrics {
+		pm.Timestamp = clientTimestamp
```
ygalblum: Even when this flag is on, you still keep track of the server side timestamps. So, the base timestamp should be separated as well so that each group (server and client side) is calculated based on their corresponding timestamps.

pkg/measurements/pod_latency.go
```diff
@@ -70,10 +70,23 @@ type podMetric struct {
 	Name                          string `json:"podName"`
 	NodeName                      string `json:"nodeName"`
 	Metadata                      any    `json:"metadata,omitempty"`
+	// High-precision client-side timestamps
+	clientScheduled              time.Time
```
ygalblum: These states should be added to the `supportedPodConditions` map so that users will be able to define failure thresholds on them

pkg/measurements/pod_latency.go
```diff
@@ -112,7 +128,14 @@ func (p *podLatency) handleCreatePod(obj any) {
 		Metadata:     p.Metadata,
 		JobIteration: getIntFromLabels(podLabels, config.KubeBurnerLabelJobIteration),
 		Replica:      getIntFromLabels(podLabels, config.KubeBurnerLabelReplica),
-	})
+	}
+
+	if p.highPrecisionMetrics {
```
ygalblum: Consider keeping track of the client side timestamps regardless to `highPrecisionMetrics` and only decide whether or not to report it in `getLatency`. It shouldn't add to much to the runtime (you fetch the local time anyhow, so it's just assignments) and it will remove a lot of `if` conditions from the code

devsubid:
moved #953

PR #943, opened 2025-08-03
Add install command to docs

### LINKED ISSUE #942: [RFE] Add the built binary to local bin directory for global support of kube-burner command
We need to add the built binary to the local bin directory manually to get it working from anywhere in the host system (Globally).

We can modify the make command to automate this, or we can add a command in the docs for convenience of the new contributors.

**Additional context**
After moving the built directory to local bin directory we can use it globally

Pratik-050:
## Type of change
- Documentation

## Description
Added commands in docs to copy the binary to the local bin directory. Also updated the make build command to do the same.

- Related Issue #942
- Closes #942

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,7 @@ $(BIN_PATH): $(SOURCES)
 		-X $(KUBE_BURNER_VERSION).BuildDate=$(BUILD_DATE) \
 		-X $(KUBE_BURNER_VERSION).Version=$(VERSION)" \
 		-o $(BIN_PATH) ./cmd/kube-burner
+	sudo cp $(BIN_PATH) /usr/local/bin/kube-burner
 
 build-release: $(SOURCES)
 	@echo -e "\033[2mBuilding secure release binary $(BIN_PATH)\033[0m"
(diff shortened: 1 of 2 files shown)


pfeifferj:
we really shouldn't make this change imo because it will break a lot of other parts:

firstly it breaks the principle of separation of concerns where building and installing should be separate operations. users may want to build without installing.

the current design builds to `bin/$(ARCH)/kube-burner` which allows for multi-architecture builds in the same workspace. 
   
the test infrastructure depends on the binary being at `$(TEST_BINARY)` which defaults to `$(CURDIR)/$(BIN_PATH)`. changing the build location would break the testing workflow.
  
the make images target expects the binary at the specific architecture path for building container images.

Pratik-050:
Hello @pfeifferj thanks for your response.
We can avoid the auto install part then, but I think we can add the installation line in the docs as I faced this issue while setting up the repo locally. Let me know if I should remove the makefile changes. Thanks!

Pratik-050:
Hello @pfeifferj We can also add another field for installation in the docs, if this sounds good then I can work on that. Thanks!

Pratik-050:
Sorry for the confusion @ygalblum 
I've updated the docs to use the existing make target named `install`. Thanks!

ygalblum (approved):
LGTM

PR #946, opened 2025-08-05
Add back logic to detect config file in CWD

rsevilla87:
## Type of change

- Bug fix

## Description

This logic was mistakenly deleted in #868, it works great now, as it allows to combine local and embedded configuration files in the same run, including alerts and metrics files too.

- Related Issue #
- Closes #

### DIFF
--- a/pkg/util/fileutils/file_reader.go
+++ b/pkg/util/fileutils/file_reader.go
@@ -73,13 +73,18 @@ func GetScriptsReader(location string, embedCfg *EmbedConfiguration) (io.Reader,
 }
 
 func getEmbedReader(location string, embedFS *embed.FS, embedDir string) (io.Reader, error) {
-	log.Debugf("Looking for file %s in embed fs", location)
-	f, err := embedFS.Open(filepath.Join(embedDir, location))
-	if err != nil {
-		log.Infof("File %s not found in the embedded filesystem. Falling back to original path", location)
+	if _, err := os.Stat(location); err == nil {
+		log.Infof("Config file %v available in the current directory, using it", location)
 		return getReader(location)
+	} else {
+		log.Debugf("Looking for file %s in embed fs", location)
+		f, err := embedFS.Open(filepath.Join(embedDir, location))
+		if err != nil {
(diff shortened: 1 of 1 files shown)


ygalblum (commented):
What will happen if the same file exists in both the local and the embedded FS? I guess there's a path within the repo that will cause this

rsevilla87:
> What will happen if the same file exists in both the local and the embedded FS? I guess there's a path within the repo that will cause this

It uses the local one

ygalblum:
> It uses the local one

Is this the behavior we want? In any case, I think it should be documented (I didn't find any reference to this)

rsevilla87:
> > It uses the local one
> 
> Is this the behavior we want? In any case, I think it should be documented (I didn't find any reference to this)

As this feature is not directly available in kube-burner, is currently documented in kube-burner-ocp docs:

https://kube-burner.github.io/kube-burner-ocp/latest/#customizing-workloads

ygalblum:
> As this feature is not directly available in kube-burner, is currently documented in kube-burner-ocp docs

I didn't look for it in `kube-burner`. My comment was not about the intentional use (which, we you linked, is documented); it was about the unintentional one. If someone runs `kube-burner-ocp` from a directory that has a file expected by the tests, this file will be picked up without the user's knowledge.

rsevilla87:
> > As this feature is not directly available in kube-burner, is currently documented in kube-burner-ocp docs
> 
> I didn't look for it in `kube-burner`. My comment was not about the intentional use (which, we you linked, is documented); it was about the unintentional one. If someone runs `kube-burner-ocp` from a directory that has a file expected by the tests, this file will be picked up without the user's knowledge.

We can add a heads up note into the kube-burner-ocp docs explaining it. Would that work for you?

ygalblum:
> We can add a heads up note into the kube-burner-ocp docs explaining it. Would that work for you?

I think so, yes.

vishnuchalla:
Also logs should make it clear where the config files are getting picked up from. So seems to be fine from usage standpoint.

vishnuchalla (approved):
lgtm

PR #947, opened 2025-08-06
feat(ci): Implement parallel testing and paths-ignore optimization

### LINKED ISSUE #897: [RFE] Optimize CI runtime

At the moment CI takes almost 1 hour to run, that can be frustrating for users opening PRs.
Also there're situations where the full CI shouldn't run, like a documentation update. 

- Save time by running tests in parallel, when possible
- Rule out PRs not touching source code to trigger CI

niladrix719:
## Type of change

- CI

## Description

Many of them rely on the same test files, which caused conflicts when accessed simultaneously. To avoid these race conditions, I moved those dependent tests into the sequential file

- Reduces time `1hr -> 30m`
- Moved Serial tests into separate bats file
- Added a new `setup_suite.bash` file to handle setup and teardown for both files together
- Added path ignores for pull requests

- Closes #897

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,18 +8,22 @@ jobs:
     uses: ./.github/workflows/ci-tests.yml
     secrets: inherit
 
-  ci-integration-tests:
-    uses: ./.github/workflows/ci-integration-tests.yml
+  ci-k8s-tests:
+    uses: ./.github/workflows/ci-k8s-tests.yml
     needs: ci-tests
     secrets: inherit
 
   release-build:
     uses: ./.github/workflows/gorelease.yml
-    needs: [ci-tests, ci-integration-tests]
+    needs:
+      - ci-tests
+      - ci-k8s-tests
 
   image-upload:
     uses: ./.github/workflows/image-upload.yml
-    needs: [ci-tests, ci-integration-tests]
+    needs:
+      - ci-tests
+      - ci-k8s-tests
     secrets: inherit
 
   deploy-docs:
--- a/.github/workflows/ci-k8s-tests.yml
+++ b/.github/workflows/ci-k8s-tests.yml
@@ -1,4 +1,4 @@
-name: CI Integration tests
+name: CI K8s tests
 on:
   workflow_call:
   workflow_dispatch:


.github/workflows/test-k8s.yml
```diff
@@ -3,6 +3,18 @@ name: Execute tests on k8s and OCP
 on:
   workflow_dispatch:
   workflow_call:
+  pull_request:
```
ygalblum: This workflow is not supposed to run on its own. It is built to be executed by `ci-tests.yml`. If you run it on its own, the `Download kube-burner binary` step will fail

test/test-k8s-serial.bats
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env bats
+# vi: ft=bash
+# shellcheck disable=SC2086,SC2030,SC2031,SC2164
+
+load helpers.bash
+export BATS_NO_PARALLELIZE_WITHIN_FILE=true
```
ygalblum: Just to make it cleaner, please put this in a `setup_file` function

ygalblum (commented):
Thanks for the PR.
Can you explain how you decided which test can be run in parallel and which can't?

niladrix719:
> Thanks for the PR. Can you explain how you decided which test can be run in parallel and which can't?

Initially, I tried running all tests in parallel, but many of them rely on the same test files, which caused conflicts when accessed simultaneously. To avoid these race conditions, I moved those tests into the sequential file

.github/workflows/ci-integration-tests.yml
```diff
@@ -0,0 +1,25 @@
...
+      - '.gitignore'
+      - 'LICENSE'
+
+jobs:
+  build:
+    uses: ./.github/workflows/builders.yml
+
+  tests:
+    needs: build
+    uses: ./.github/workflows/test-k8s.yml
```
vishnuchalla, line 24: what about serial tests in the CI?

.github/workflows/release.yml
```diff
@@ -8,13 +8,18 @@ jobs:
     uses: ./.github/workflows/ci-tests.yml
     secrets: inherit
 
+  ci-integration-tests:
+    uses: ./.github/workflows/ci-integration-tests.yml
```
vishnuchalla: can we use a different name please?  integration tests doesn't look like a fit.

.github/workflows/ci-integration-tests.yml
ygalblum: Not sure what you're asking here. The parallelism is handled by bats

.github/workflows/release.yml
```diff
@@ -8,13 +8,18 @@ jobs:
...
+    needs: ci-tests
+    secrets: inherit
+
   release-build:
     uses: ./.github/workflows/gorelease.yml
-    needs: ci-tests
+    needs: [ci-tests, ci-integration-tests]
```
ygalblum: Small nit, I'm not a fan of json arrays inside a yaml
```suggestion
    needs:
      - ci-tests
      - ci-integration-tests
```

.github/workflows/release.yml
niladrix719: I’ve changed it to `ci-k8s-tests`. Let me know if we should use something else

.github/workflows/release.yml
niladrix719: sure, done

rsevilla87:
Closing in favor of #1039

PR #949, opened 2025-08-07
Additional vars have prececence over env vars

rsevilla87:
## Type of change

- Bug fix
- Optimization

## Description

Environment variables now get overwritten by additionalVars, (which is more logical), that are can be passed from flags in wrappers and arbitrary YAML files.

This is required to avoid issues with variables using data types different than "string", which can be overridden by environment variables

Also removing the variable inputData, which is useless.

- Related Issue #
- Closes #


mohit-sheth (dismissed):
lgtm, thanks for the quick fix @rsevilla87

rsevilla87:
Hey @ygalblum, this PR updates the user variables logic a little bit, and I had to update the tests hope it doesn't affect your work PTAL

ygalblum:
Fine by me. But, the test fails, not sure why

ygalblum (approved):
LGTM

PR #953, opened 2025-08-09
feat(pod_latency): Add high-precision metrics support for pod latency

### LINKED ISSUE #930: How to get podLatency in ms

devsubid:
# Add High-Precision Pod Latency Measurements

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation
- [ ] CI

## Description

This PR introduces **high-precision pod latency measurements** to address timing precision limitations in Kubernetes API timestamps.

### Problem

Current pod latency measurements suffer from second-only precision due to Kubernetes API constraints:

- API server only supports RFC3339 format (second precision), not RFC3339NANO
- Kubelet purposely discards microseconds/nanoseconds due to API limitations
- Fast pod operations (< 1 second) result in zero or negative latencies
- Sub-second pod startup scenarios cannot be measured accurately

### Solution

Implements client-side timestamp capture using `time.Now().UTC()` with nanosecond precision when kube-burner observes pod events in real-time.

### Key Features

- **Dual measurement system**: Provides both standard (API-based) and high-precision (client-based) metrics
- **Sub-millisecond accuracy**: Float64 values in milliseconds with decimal precision (e.g., `123.456ms`)
- **Client-side timestamps**: Captures exact moment events are observed, not API timestamps
- **Backward compatibility**: Existing measurements remain unchanged

### Configuration

```yaml
global:
  measurements:
    - name: podLatency
      highPrecisionMetrics: true # Enable high-precision mode
```

### Output Metrics

When enabled, adds high-precision metrics with "Client" prefix:

- `ClientPodScheduled`: 123.456ms
- `ClientPodReady`: 456.789ms
- `ClientPodInitialized`: 234.567ms
- `ClientContainersReady`: 345.678ms

### Implementation

- Added `HighPrecisionMetrics bool` field to Measurement struct
- Enhanced `podMetric` with client-side timestamp and latency fields
- Updated event handlers to capture client-side timestamps
- Modified metric calculation to include high-precision measurements
- Uses nanosecond resolution converted to milliseconds with float64

### Benefits

- Accurate SLO/SLA tracking for pod startup performance
- Better bottleneck identification for sub-second operations
- Improved granularity with microsecond-level timing
- Eliminates negative latency measurement errors
- Enhanced Kubernetes cluster performance observability

- Related Issue: Pod latency measurements limited to second precision
- Addresses: Inaccurate measurements due to Kubernetes API timestamp constraints
- Closes #930

### DIFF
--- a/pkg/measurements/pod_latency.go
+++ b/pkg/measurements/pod_latency.go
@@ -46,6 +46,12 @@ var (
 		string(corev1.PodInitialized):  {},
 		string(corev1.PodReady):        {},
 		string(corev1.PodScheduled):    {},
+		// Client-side high-precision conditions
+		"ClientContainersReady":           {},
+		"ClientPodInitialized":            {},
+		"ClientPodReady":                  {},
+		"ClientPodScheduled":              {},
+		"ClientPodReadyToStartContainers": {},
 	}
 )
 
@@ -70,10 +76,24 @@ type podMetric struct {
 	Name                          string `json:"podName"`
 	NodeName                      string `json:"nodeName"`
 	Metadata                      any    `json:"metadata,omitempty"`
+	// High-precision client-side timestamps
+	clientCreationTimestamp      time.Time // Client-side creation timestamp for high-precision base
+	clientScheduled              time.Time
+	clientInitialized            time.Time
+	clientContainersReady        time.Time
+	clientPodReady               time.Time
+	clientReadyToStartContainers time.Time
+	// High-precision latencies (in milliseconds with sub-millisecond precision)
+	ClientSchedulingLatency             float64 `json:"clientSchedulingLatency,omitempty"`
+	ClientInitializedLatency            float64 `json:"clientInitializedLatency,omitempty"`
+	ClientContainersReadyLatency        float64 `json:"clientContainersReadyLatency,omitempty"`
+	ClientPodReadyLatency               float64 `json:"clientPodReadyLatency,omitempty"`
+	ClientReadyToStartContainersLatency float64 `json:"clientReadyToStartContainersLatency,omitempty"`
 }
 
 type podLatency struct {
 	BaseMeasurement
+	highPrecisionMetrics bool
 }
 
 type podLatencyMeasurementFactory struct {
@@ -91,7 +111,8 @@ func newPodLatencyMeasurementFactory(configSpec config.Spec, measurement types.M
 
 func (plmf podLatencyMeasurementFactory) NewMeasurement(jobConfig *config.Job, clientSet kubernetes.Interface, restConfig *rest.Config, embedCfg *fileutils.EmbedConfiguration) Measurement {
 	return &podLatency{
-		BaseMeasurement: plmf.NewBaseLatency(jobConfig, clientSet, restConfig, podLatencyMeasurement, podLatencyQuantilesMeasurement, embedCfg),
+		BaseMeasurement:      plmf.NewBaseLatency(jobConfig, clientSet, restConfig, podLatencyMeasurement, podLatencyQuantilesMeasurement, embedCfg),
+		highPrecisionMetrics: plmf.Config.HighPrecisionMetrics,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/measurements/pod_latency.go
```diff
@@ -132,22 +160,40 @@ func (p *podLatency) handleUpdatePod(obj any) {
 						if pm.scheduled.IsZero() {
 							pm.scheduled = c.LastTransitionTime.UTC()
 							pm.NodeName = pod.Spec.NodeName
+							// Always capture client-side timestamp regardless of highPrecisionMetrics flag
+							if pm.clientScheduled.IsZero() {
```
ygalblum: Is this check really needed? Is there a case where `pm.scheduled.IsZero()` will return `true` but `pm.clientScheduled.IsZero()` will return false? (same for the ones below)

pkg/measurements/pod_latency.go
```diff
@@ -211,20 +257,32 @@ func (p *podLatency) Collect(measurementWg *sync.WaitGroup) {
 				containersReady = c.LastTransitionTime.UTC()
 			case corev1.PodReady:
 				podReady = c.LastTransitionTime.UTC()
+			case corev1.PodReadyToStartContainers:
+				readyToStartContainers = c.LastTransitionTime.UTC()
 			}
 		}
+
+		// For collected metrics, use server-side timestamps as base
```
ygalblum: Sorry, I don't understand this change.

pkg/measurements/pod_latency.go
```diff
@@ -253,33 +311,91 @@ func (p *podLatency) normalizeMetrics() float64 {
 		// v2 latencies are currently under AB testing which blindly trust kubernetes as source of
 		// truth and will prevent us from those over 1s delays as well as <0 cases.
 		errorFlag := 0
-		m.ContainersReadyLatency = int(m.containersReady.Sub(m.Timestamp).Milliseconds())
+
+		// For server-side calculations, use the server-side base timestamp
+		serverBaseTimestamp := m.Timestamp
```
ygalblum: Adding this field is not needed. Please revert

pkg/measurements/pod_latency.go
```diff
@@ -253,33 +311,91 @@ func (p *podLatency) normalizeMetrics() float64 {
...
+
+			if !m.clientReadyToStartContainers.IsZero() {
+				m.ClientReadyToStartContainersLatency = float64(m.clientReadyToStartContainers.Sub(clientBaseTimestamp).Nanoseconds()) / 1e6
+				if m.ClientReadyToStartContainersLatency < 0 {
+					m.ClientReadyToStartContainersLatency = 0
+				}
+			}
+		}
+
+		if p.highPrecisionMetrics {
```
ygalblum: Why is this print needed? The method doesn't print other things. Keep in mind that when you have a load of pods this print will flood the log

pkg/measurements/pod_latency.go
```diff
@@ -46,6 +46,12 @@ var (
 		string(corev1.PodInitialized):  {},
 		string(corev1.PodReady):        {},
 		string(corev1.PodScheduled):    {},
+		// Client-side high-precision conditions
+		"ClientContainersReady":           {},
```
ygalblum: Please define a use consts for these strings (here and in `getLatency`)

devsubid:
Thanks for your valuable review, i have incorporated all your requested changes

pkg/measurements/pod_latency.go
```diff
@@ -280,6 +335,52 @@ func (p *podLatency) normalizeMetrics() float64 {
...
+			}
+
+			if !m.clientInitialized.IsZero() {
+				m.ClientInitializedLatency = float64(m.clientInitialized.Sub(m.clientCreationTimestamp).Nanoseconds()) / 1e6
+				if m.ClientInitializedLatency < 0 {
+					m.ClientInitializedLatency = 0
+				}
+			}
+
+			if !m.clientContainersReady.IsZero() {
```
ygalblum, line 362: The struct is initialized with this field set, so this statement is always true

ygalblum (changes requested):
Sorry, I missed this comment in the previous review.
As for the test, first, you added it to a `vmiLatency`, which does not support this. You should add it to `kube-burner.yml` in the same directory. You should also add a check that the data was actually reported using `check_metric_value` in one of the tests that use it in `test-k8s.bats`

pkg/measurements/pod_latency.go
devsubid: im not getting. could you provide more details about it?

test/test-k8s.bats
```diff
@@ -103,6 +103,16 @@ teardown_file() {
   check_destroyed_pods default kube-burner-job=not-namespaced,kube-burner-uuid="${UUID}"
 }
 
+@test "kube-burner init: local-indexing=true; high-precision-pod-latency=true" {
```
ygalblum, line 106: Since you added the field unconditionally to `kube-burner.yml`, the data will be included in all the tests that use this file. So, you don't need a dedicated test case, you can just add the checks (calls to `check_high_precision_metrics` and `check_high_precision_quantiles`) to an existing test. Doing so will allow testing the feature with a very small addition to the CI runtime

pkg/measurements/pod_latency.go
ygalblum: I think I commented on the wrong line. This comment is about the check `if !m.clientCreationTimestamp.IsZero()`. 
All the instances are created in `handleCreatePod` where the field `clientCreationTimestamp` is set. So, it will never be Zero

vishnuchalla:
Closing this as stale and incomplete. Please feel free to join upstream office hours if you have any further questions/concerns.

PR #956, opened 2025-08-09
Add better error logs for init command

### LINKED ISSUE #955: [RFE] Add more descriptive error logs for kube-burner init command
Currently the `init` command always returns the same error message weather there is no running cluster or there is no config file. It makes the debugging tedious. 

Image

We can add better logs to identify where the things are going wrong.

Pratik-050:
## Type of change

- Refactor

## Description
Added different error logs while checking for cluster status and config file (global .kube/config or provided config.yaml)

- Related Issue #955
- Closes #955

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -233,7 +233,7 @@ func NewKubeClientProvider(config, context string) *KubeClientProvider {
 	var err error
 	if kubeConfigPath == "" {
 		if restConfig, err = rest.InClusterConfig(); err != nil {
-			log.Fatalf("error preparing kubernetes client: %s", err)
+			log.Fatalf("no running cluster found (no kubeconfig and no in-cluster config): %v", err) // If no kubeconfig is provided or no env vars are set
 		}
 	} else {
 		kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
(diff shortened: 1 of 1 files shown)


pkg/config/config.go
```diff
@@ -233,15 +233,23 @@ func NewKubeClientProvider(config, context string) *KubeClientProvider {
...
 		kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
 			&clientcmd.ClientConfigLoadingRules{ExplicitPath: kubeConfigPath},
 			&clientcmd.ConfigOverrides{CurrentContext: context},
 		)
 		if restConfig, err = kubeConfig.ClientConfig(); err != nil {
-			log.Fatalf("error preparing kubernetes client: %s", err)
+			log.Fatalf("invalid kubeconfig or unreachable cluster: %v", err) // If kubeconfig is provided, but invalid or cluster is unreachable
+		}
+
+		clientset, err := kubernetes.NewForConfig(restConfig)
```
ygalblum, line 247: I'm guessing you added these calls for a quick failure check. If so, are there any other commands that can benefit from this? If yes, then move the check to a separate function and use it where needed

Pratik-050: Sorry for the confusion @ygalblum, it sounds confusing because of the title of the PR. Actually these changes applies for all the commands as we are checking if there is a running cluster. It adds more error logs depending on state of the cluster.

ygalblum: Sorry, I missed the fact that it's in the `config.go` file. Thanks for the explanation

PR #963, opened 2025-08-16
Add validate command to check valid config

### LINKED ISSUE #962: [RFE] Add validate command to check valid config files
If we run the `init` command with an invalid config file it creates a `.log` file every time. We don't really need to save these logs. 

We can add another command `validate` to check if our config file is valid or not to use with `init` command. It acts like the syntax check logic for nginx files using the `sudo nginx -t`

Pratik-050:
## Type of change

- New feature

## Description
Added a new command for checking the validity of `config.yaml` file before running `init` command. Utilizes already existing logic of `ParseWithUserdata` used in init command.
For valid file:
image
For invalid file:
image

- Related Issue #962
- Closes #962

### DIFF
--- a/cmd/kube-burner/kube-burner.go
+++ b/cmd/kube-burner/kube-burner.go
@@ -518,6 +518,61 @@ func alertCmd() *cobra.Command {
 	return cmd
 }
 
+func validateCmd() *cobra.Command {
+	var uuid string
+	var cfgFile string
+	var userDataFile string
+	var allowMissingKeys bool
+	var timeout time.Duration
+
+	cmd := &cobra.Command{
+		Use:   "validate",
+		Short: "Validate kube-burner configuration file",
+		PreRun: func(cmd *cobra.Command, args []string) {
+			if uuid == "" {
+				uuid = uid.NewString()
+			}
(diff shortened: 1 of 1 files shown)


ygalblum:
The code here duplicates what is already implemented for the init command. We should avoid code duplications as they add to the maintenance load.  
I have to say, I don't understand the need for this command to begin with

Pratik-050:
Hello @ygalblum we can use this command for a quick check of the config file even when there is no live cluster present. Also validation ensures configs are syntactically and structurally correct without provisioning pods, saving computation. 
As talking about breaking DRY design pattern, we can export this whole code block to a separate function. Thanks!

ygalblum:
Thanks for your response, I understand the need better now. So, yes, we should reuse code and not duplicate. In addition, can you think of other configuration issues the validation should look at? IIUC we mainly check the YAML validity, but not the content itself (or maybe I'm missing something)

Pratik-050:
Thanks for your response @ygalblum 
I'll take a look if anything else can be checked in a valid YAML via the validate command, and move the checker to a separate function.

Issue #576, opened 2024-01-16
[RFE] Parallel job execution


riuvshyn:
**Is your feature request related to a problem? Please describe.**
Is there any way to set up parallel job execution? 

**Describe the solution you'd like**
I would love to be able to run some jobs in parallel to be able to generate more real world kind of load.

**Describe alternatives you've considered**
As an alternative I am running multiple instances of kube-burner which is quite complicated when it gets to measurements and report analysing.

rsevilla87:
Interesting idea, I guess we can create something like execution groups, so we can group jobs of the same groups to be executed in parallel

vishnuchalla:
@riuvshyn while we look at this issue, along side can you share the `config.yaml` file that you are using to run your jobs? If you don't mind.

vishnuchalla:
@riuvshyn Based on your idea, here are some thoughts around its implementation.
* We enable multiple jobs execution in parallel, which will have their inbuilt measurements like `podlatency`, `vmlatency` and `pprof` etc specific to an individual job. And cluster prometheus metrics can be recorded between the start time and end time of all these jobs that executed in parallel as a whole. So that a user will be able to have detailed look at cluster level metrics while these jobs where executing in parallel and at the time taken for their relevant resources to come up.
* We can achieve this by introducing a new option in [Job](https://github.com/kube-burner/kube-burner/blob/main/pkg/config/types.go#L102) struct as `executionGroup` to group jobs that needs to be executed in parallel ensuring backward compatibility. But then all these jobs must have unique names, same type (.i.e one among of create, delete and patch) and would be better if they are in different namespaces.

Let us know if this plan aligns with your use case. Thank you.
cc: @rsevilla87 @shashank-boyapally

ygalblum:
Reopening this issue as there's an open PR addressing it

sanjay7178:
Should we consider a DAG based job execution , which could include a parallel execution as well ?

vishnuchalla:
> Should we consider a DAG based job execution , which could include a parallel execution as well ?

@sanjay7178 Can you please elaborate on this idea? We were thinking of just having tests execute parallel through go threads.

sanjay7178:
> > Should we consider a DAG based job execution , which could include a parallel execution as well ?
> 
> @sanjay7178 Can you please elaborate on this idea? We were thinking of just having tests execute parallel through go threads.

[...]

PR #968, opened 2025-08-21
feat: add execution group support for parallel job execution

niladrix719:
## Type of change

- New feature

## Description

- Add executionGroup field to job config and parser.
- Run contiguous jobs in the same executionGroup in parallel (same jobType, unique names).
- Create a synthetic `group-<name>` job to provide a single Prometheus metrics window for the group.

Issue #576


pkg/burner/job.go
```diff
@@ -100,121 +108,33 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
...
-					log.Info("Churning enabled")
-					log.Infof("Churn cycles: %v", jobExecutor.ChurnCycles)
-					log.Infof("Churn duration: %v", jobExecutor.ChurnDuration)
-					log.Infof("Churn percent: %v", jobExecutor.ChurnPercent)
-					log.Infof("Churn delay: %v", jobExecutor.ChurnDelay)
-					log.Infof("Using deletion strategy: %v", jobExecutor.deletionStrategy)
+			groupName := jobExecutor.ExecutionGroup
+			groupType := jobExecutor.JobType
+			groupMembers := []JobExecutor{jobExecutor}
+			for j := jobExecutorIdx + 1; j < len(jobExecutors); j++ {
```
ygalblum, line 123: IIUC the `break` in the `else` case makes sure that jobs that belong to the same execution group must also be consecutive in the test spec. 
I think it should be this way and if not the code should be changed.
If I am right then first It should be documented. Second, there's no need to maintain the `skip` array. Change the top `for` statement to:
```
for jobExecutorIdx := 0; jobExecutorIdx < len(jobExecutors); jobExecutorIdx++
```
Then, you can just increment `jobExecutorIdx` (or assign `j` to it) in order to skip the processed ones

pkg/burner/job.go
```diff
@@ -300,6 +220,269 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
...
+		jobExecutor.Run(ctx)
+		if ctx.Err() != nil {
+			return 0, ctx.Err()
+		}
+	}
+	return 0, nil
+}
+
+// runBeforeCleanup executes beforeCleanup command if configured
+func runBeforeCleanup(jobExecutor JobExecutor) (int, error) {
```
ygalblum, line 272: The returned `int` seems redundant. The code returns either `0, nil` or `1, err` when `err != nil`. So the caller can just check `err != nil`

pkg/burner/job.go
```diff
@@ -300,6 +220,269 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
...
+	for idx, watcher := range jobExecutor.Watchers {
+		for replica := range watcher.Replicas {
+			watcherManager.Start(watcher.Kind, watcher.APIVersion, watcher.LabelSelector, idx+1, replica+1)
+		}
+	}
+	return watcherManager, watcherManager.Wait()
+}
+
+// executeJob runs the main work for a job
+func executeJob(ctx context.Context, jobExecutor JobExecutor, waitListNamespaces *[]string) (int, error) {
```
ygalblum, line 235: Since `ctx` belongs to the caller, returning `ctx.Err()` is not needed. Furthermore, the caller looks at `err` only if the returned `int` is not 0. So, the return of this function can be simplified.
Change `return 0, ctx.Err()` to `return nil` (it will also remove the need to check and return from the `else` statement.
Change `return 1, verr` to `return verr`
Then the caller can check `err`, look at `jobExecutor.ErrorOnVerify` and decide what to do next

pkg/burner/job.go
```diff
@@ -300,6 +220,269 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
...
+			errsMu.Lock()
+			*errs = slices.Concat(*errs, watcherStopErrs)
+			errsMu.Unlock()
+			if m.ex.GC {
+				m.ex.gc(ctx, nil)
+			}
+		}(m)
+	}
+	groupWg.Wait()
+	maxMemberRC := 0
```
ygalblum, line 461: `innerRC` is not a counter it's an error code. Looking for `max` does not make any sense. There should be a decision which value to return. @rsevilla87 @vishnuchalla WDYT?

pkg/burner/job.go
```diff
@@ -300,6 +220,269 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
...
+		globalWaitMap[strconv.Itoa(jobExecutorIdx)+jobExecutor.Name] = waitListNamespaces
+		executorMap[strconv.Itoa(jobExecutorIdx)+jobExecutor.Name] = jobExecutor
+	}
+	if cleanupRC, cleanupErr := runBeforeCleanup(jobExecutor); cleanupErr != nil {
+		*errs = append(*errs, cleanupErr)
+		*innerRC = cleanupRC
+	}
+	jobEnd := time.Now().UTC()
+	if jobExecutor.MetricsClosing == config.AfterJob {
+		(*executedJobs)[len(*executedJobs)-1].End = jobEnd
```
ygalblum, line 327: Can you find a simpler way to get this instance? Maybe get a pointer to it once on top and then dereference it? (also in `runExecutionGroup`)

test/helpers.bash
```diff
@@ -172,11 +172,11 @@ check_running_pods() {
 }
 
 check_running_pods_in_ns() {
-    running_pods=$(kubectl get pod -n "${1}" -l kube-burner-job=namespaced --field-selector=status.phase==Running -o json | jq '.items | length')
-    if [[ "${running_pods}" != "${2}" ]]; then
-      echo "Running pods in namespace $1 different from expected. Expected=${2}, observed=${running_pods}"
-      return 1
-    fi
+  running_pods=$(kubectl get pod -n "${2}" -l kube-burner-job="${1}" --field-selector=status.phase==Running -o json | jq '.items | length')
```
ygalblum, line 175: Please add local variables and assign the passed arguments to them. It will make the code easier to read

pkg/burner/job.go
```diff
@@ -300,6 +220,269 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
...
+	}
+	watcherStopErrs := watcherManager.StopAll()
+	*errs = slices.Concat(*errs, watcherStopErrs)
+	if jobExecutor.GC {
+		jobExecutor.gc(ctx, nil)
+	}
+}
+
+// runExecutionGroup runs a contiguous group of jobs in parallel
+func runExecutionGroup(ctx context.Context, groupName string, groupMembers []JobExecutor, restConfig *rest.Config, measurementsFactory *measurements.MeasurementsFactory, kubeClientProvider *config.KubeClientProvider, embedCfg *fileutils.EmbedConfiguration, metricsScraper metrics.Scraper, globalConfig config.GlobalConfig, executedJobs *[]prometheus.Job, globalWaitMap map[string][]string, executorMap map[string]JobExecutor, msWg *sync.WaitGroup, errs *[]error, innerRC *int) {
```
ygalblum, line 370: Something in the naming of `groupMembers` is confusing because `members` is of type `[]groupMember` but `groupMembers` is of type `[]JobExecutor`

pkg/burner/job.go
```diff
@@ -300,6 +220,269 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
...
+			MetricsClosing: config.AfterJob,
+			JobType:        groupMembers[0].JobType,
+		},
+	})
+	log.Infof("Triggering executionGroup %q with %d jobs in parallel", groupName, len(groupMembers))
+	members := make([]*groupMember, 0, len(groupMembers))
+	for _, member := range groupMembers {
+		watcherManager, watcherStartErrors := setupWatchers(member, restConfig)
+		*errs = slices.Concat(*errs, watcherStartErrors)
+		measurementsInstance := measurementsFactory.NewMeasurements(&member.Job, kubeClientProvider, embedCfg)
```
ygalblum, line 385: You are disregarding `jobExecutor.MetricsAggregate` in this function. Is that intentional? If yes, it should be documented and emit an log message

test/test-k8s.bats
```diff
@@ -274,3 +274,11 @@ teardown_file() {
     check_metrics_not_created_for_job ${job} ${metric}
   done
 }
+
+@test "kube-burner init: execute jobs in group" {
```
ygalblum, line 278: Some questions about the test:

1. What in this test verifies that the jobs were executed together?
2. The test should also include jobs that are not grouped to verify that not all jobs get executed together. Maybe something like, two grouped, one separate and then two more grouped
3. You are not testing anything about the collected metrics. Can running parallel test impact the collection? Are we sure each output includes only the intended ones?

pkg/burner/job.go
vishnuchalla: +1 on a value in the return.

pkg/burner/job.go
```diff
@@ -100,121 +108,33 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
...
 			}
-			watcherStartErrors := watcherManager.Wait()
-			slices.Concat(errs, watcherStartErrors)
-			var waitListNamespaces []string
-			if measurementsInstance == nil {
-				measurementsJobName = jobExecutor.Name
-				measurementsInstance = measurementsFactory.NewMeasurements(&jobExecutor.Job, kubeClientProvider, embedCfg)
-				measurementsInstance.Start()
+			if jobExecutor.ExecutionGroup == "" {
+				runSingleJob(ctx, jobExecutorIdx, jobExecutor, restConfig, measurementsFactory, kubeClientProvider, embedCfg, metricsScraper, globalConfig, &measurementsInstance, &measurementsJobName, &executedJobs, globalWaitMap, executorMap, &msWg, &errs, &innerRC)
```
vishnuchalla, line 117: These are too many details in one single function call. I wonder if we can get some of them into a struct.

Issue #517, opened 2023-11-18
[RFE] Pass nodes information to templates


rsevilla87:
**Is your feature request related to a problem? Please describe.**

Sending node information to the templates can be useful to implement things based on the node specs. i.e: testing a nodeport service

**Describe the solution you'd like**
The second argument of the function `template.Execute` receives an array of interfaces, that allows to send multiple arrays of nodes to the templates (one array per node role)

PR #973, opened 2025-08-22
feat: Add functionality to pass nodes info to templates

niladrix719:
## Type of change

- New feature

## Description

functionality to pass nodes info to templates

Issue #517

### DIFF
--- a/pkg/util/nodes.go
+++ b/pkg/util/nodes.go
@@ -1,0 +1,79 @@
+package util
+
+import (
+	"context"
+	"sort"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// LoadNodesByRole queries cluster nodes and groups them by commonly used roles.
+func LoadNodesByRole(clientSet kubernetes.Interface) map[string]corev1.NodeList {
+	byRole := map[string]corev1.NodeList{"all": {Items: []corev1.Node{}}}
+	if clientSet == nil {
+		return byRole
+	}
+	nodes, err := clientSet.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		log.Debugf("Skipping node info injection: failed to list nodes: %v", err)
+		return byRole
+	}
+	add := func(role string, n corev1.Node) {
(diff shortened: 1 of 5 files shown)


pkg/util/nodes.go
```diff
@@ -0,0 +1,79 @@
+package util
```
ygalblum, line 1: `util` packages are considered an anti-pattern in golang. I know that we have it, but I prefer we avoid adding new content to it. Maybe `nodes`

pkg/util/nodes.go
```diff
@@ -0,0 +1,79 @@
...
+	log "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// LoadNodesByRole queries cluster nodes and groups them by commonly used roles.
+func LoadNodesByRole(clientSet kubernetes.Interface) map[string]corev1.NodeList {
+	byRole := map[string]corev1.NodeList{"all": {Items: []corev1.Node{}}}
+	if clientSet == nil {
```
ygalblum, line 17: `clientSet` should not be `nil`. If it is, it is a coding error. No need to check this

pkg/util/nodes.go
```diff
@@ -0,0 +1,79 @@
...
+		if len(roles) == 0 {
+			add("unknown", n)
+		} else {
+			for _, r := range roles {
+				add(r, n)
+			}
+		}
+		add("all", n)
+	}
+	for role, nl := range byRole {
```
ygalblum, line 45: Why is sorting needed?
Also, did we consider making `Items` a `map[string]corev1.Node` instead of an array? @rsevilla87 WDYT?

pkg/util/nodes.go
```diff
@@ -0,0 +1,79 @@
...
+			} else if r != "" {
+				rolesSet[r] = struct{}{}
+			}
+		}
+	}
+	out := make([]string, 0, len(rolesSet))
+	for r := range rolesSet {
+		out = append(out, r)
+	}
+	sort.Strings(out)
```
ygalblum, line 77: Why is sorting needed?

vishnuchalla:
@rsevilla87 @niladrix719 I don't think I fully understood need for this feature. Can you elaborate on any use cases that you have in mind?

rsevilla87:
As discussed in slack, let's close this RFE for now, we don't have a clear use case for it and it's better to defer it.

PR #974, opened 2025-08-24
feat: report measurements even when job times out

niladrix719:
## Type of change

- Refactor
- New feature

## Description

Instead of using fatal logs in wait methods, errors are now returned to the caller

Issue #742

### DIFF
--- a/pkg/burner/create.go
+++ b/pkg/burner/create.go
@@ -87,7 +87,7 @@ func (ex *JobExecutor) setupCreateJob(mapper meta.RESTMapper) {
 }
 
 // RunCreateJob executes a creation job
-func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterationEnd int, waitListNamespaces *[]string) {
+func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterationEnd int, waitListNamespaces *[]string) []error {
 	nsAnnotations := make(map[string]string)
 	nsLabels := map[string]string{
 		"kube-burner-job":   ex.Name,
@@ -97,6 +97,7 @@ func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterati
 	var wg sync.WaitGroup
 	var ns string
 	var err error
+	var waitErrors []error
 	maps.Copy(nsLabels, ex.NamespaceLabels)
 	maps.Copy(nsAnnotations, ex.NamespaceAnnotations)
 	if ex.nsRequired && !ex.NamespacedIterations {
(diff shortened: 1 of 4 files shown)


pkg/burner/create.go
```diff
@@ -169,46 +169,8 @@ func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterati
...
-		wg.Wait()
-		close(errChan)
-
-		select {
-		case err := <-errChan:
-			if err != nil && waitErrors == nil {
-				waitErrors = append(waitErrors, err...)
-			}
-		default:
+		if errs := ex.waitForCompletion(iterationStart, iterationEnd, ns, namespacesWaited); len(errs) > 0 && waitErrors == nil {
```
ygalblum, line 180: I know you copied it from the current code. But, if `waitErrors == nil` why do you need to use `append`. Can you just assign `waitErrors == errs`? Or maybe you should not check if it's nil?

pkg/burner/job.go
```diff
@@ -217,7 +226,10 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
 			}
 		}
 		if globalConfig.WaitWhenFinished {
-			runWaitList(globalWaitMap, executorMap)
+			if waitErrs := runWaitList(globalWaitMap, executorMap); len(waitErrs) > 0 {
```
ygalblum, line 220: Now that the executor runner returns the error, does it make sense to wait for the objets if it failed? Won't it wait forever?

ygalblum (commented):
Thanks for the PR. In addition to the comments, we need a way to test this change

pkg/burner/create.go
```diff
@@ -244,6 +230,51 @@ func (ex *JobExecutor) replicaHandler(ctx context.Context, labels map[string]str
 	wg.Wait()
 }
 
+// waitForCompletion waits for objects to be ready across the relevant namespaces
+func (ex *JobExecutor) waitForCompletion(iterationStart, iterationEnd int, ns string, namespacesWaited map[string]bool) []error {
+	log.Infof("Waiting up to %s for actions to be completed", ex.MaxWaitTimeout)
+	// This semaphore limits the maximum number of concurrent goroutines
+	sem := make(chan int, int(ex.restConfig.QPS))
+	errChan := make(chan []error, 1)
```
vishnuchalla, line 246: Do we only want to capture 1 error here?

pkg/burner/job.go
```diff
@@ -406,13 +419,27 @@ func runWaitList(globalWaitMap map[string][]string, executorMap map[string]JobEx
...
+					errChan <- err
+				}
 			}(ns)
 		}
 		wg.Wait()
 	}
+	close(errChan)
+
+	var errs []error
+	for err := range errChan {
```
vishnuchalla, line 411: In contrast to above here we seems to be capturing all the errors. Not sure why?

vishnuchalla:
Closing in favor of this work in progress draft: https://github.com/kube-burner/kube-burner/pull/1044

PR #976, opened 2025-08-26
Handle file read errors

rsevilla87:
## Type of change

- Optimization

## Description

The current logic does not handle file read errors correctly.
The PR also simplifies the overall function logic to receive directory names rather than loose files.

- Related Issue https://github.com/kube-burner/kube-burner-ocp/issues/300
- Closes #

### DIFF
--- a/pkg/workloads/helpers.go
+++ b/pkg/workloads/helpers.go
@@ -16,6 +16,8 @@ package workloads
 
 import (
 	"embed"
+	"io/fs"
+	"os"
 	"path"
 
 	"github.com/kube-burner/kube-burner/pkg/burner"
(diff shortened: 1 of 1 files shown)


pkg/workloads/helpers.go
```diff
@@ -105,27 +104,26 @@ func (wh *WorkloadHelper) RunWithAdditionalVars(configFile string, additionalVar
...
+		dirContent, err := embedConfig.ReadDir(path.Join(embedFSDir, folder))
 		if err != nil {
 			return err
 		}
-	}
-	for _, f := range rootCfg {
-		fileContent, _ := embedConfig.ReadFile(path.Join(embedFSDir, f))
-		if err = util.CreateFile(f, fileContent); err != nil {
-			return err
+		for _, f := range dirContent {
```
ygalblum: Can `f` be a directory of its own? If yes, what would happen?

pkg/workloads/helpers.go
```diff
@@ -105,27 +104,26 @@ func (wh *WorkloadHelper) RunWithAdditionalVars(configFile string, additionalVar
...
+			fileContent, err = embedConfig.ReadFile(path.Join(embedFSDir, folder, f.Name()))
+			if err != nil {
+				return err
+			}
+			if util.CreateFile(f.Name(), fileContent) != nil {
```
ygalblum: All files are stored in the same local directory. What is the same filename exists in different paths?

pkg/workloads/helpers.go
rsevilla87: Our current use cases (only kube-burner-ocp makes use of it) does not use nested directories, we can improve that logic in a follow-up PR (it wasn't supported in the original logic either). The PR goal is about to improve error handling and simplify overall function structure.

pkg/workloads/helpers.go
rsevilla87: What do you mean exactly? having for example a metric profile with a name similar to a object template?

If so, that's a corner case that shouldn't happen in regular conditions, I can add some log message to make the user know that a file existed in the path and was overwritten.

pkg/workloads/helpers.go
ygalblum: My worry is that it will fail because of a directory. So, maybe explicitly check and add a log message?

pkg/workloads/helpers.go
ygalblum: Maybe I'm looking at it in a more generic way and the actual use cases do not have this issue. But, users can organize their files within directories and state it in the path. So, you could have two `vm.yml` files. Having said that, the copy will probably also break the paths anyhow. So, not sure what to do with it

rsevilla87:
@ygalblum I've improved the logic to also extract nested directories

ygalblum (approved):
LGTM.

ygalblum:
@rsevilla87 I've restarted the failed test. Though it's not related to this change

PR #977, opened 2025-08-29
fail on HTTP errors when collecting pprofs

mohit-sheth:
- Added `-f` flag to curl commands to avoid writing HTTP error responses (like 404) to .pprof files.
- Ensures pprof collection fails fast per pod while keeping remaining pods unaffected.
- Prevents invalid profiles from being saved and logged.


afcollins:
Will it change the exit code of the overall kube-burner run ?

mohit-sheth:
nope, just log an error and move on
failing pprof collection is not critical

rsevilla87 (approved):
lgtm

PR #985, opened 2025-09-08
Local single kind

ygalblum:
## Type of change

- Documentation
- CI

## Description

Add script to start Kind locally once and use it as an external cluster


vishnuchalla (approved):
lgtm

PR #991, opened 2025-09-24
Decrease info logging

rsevilla87:
## Type of change

- Optimization

## Description

The current INFO logging level is too verbose, and makes the output difficult to read. Decreasing the logging level of some messages helps on that. Also adding job name prefixes to some log messages improve their readability.

- Related Issue #
- Closes #

### DIFF
--- a/pkg/burner/create.go
+++ b/pkg/burner/create.go
@@ -81,7 +81,7 @@ func (ex *JobExecutor) setupCreateJob(mapper meta.RESTMapper) {
 		if obj.namespaced && obj.namespace == "" {
 			ex.nsRequired = true
 		}
-		log.Infof("Job %s: %d iterations with %d %s replicas", ex.Name, ex.JobIterations, obj.Replicas, gvk.Kind)
+		log.Debugf("Job %s: %d iterations with %d %s replicas", ex.Name, ex.JobIterations, obj.Replicas, gvk.Kind)
 		ex.objects = append(ex.objects, obj)
 	}
 }
@@ -116,7 +116,7 @@ func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterati
 			return
 		}
 		if i == iterationStart+iterationProgress*percent {
-			log.Infof("%v/%v iterations completed", i-iterationStart, iterationEnd-iterationStart)
+			log.Debugf("%v/%v iterations completed", i-iterationStart, iterationEnd-iterationStart)
 			percent++
 		}
 		log.Debugf("Creating object replicas from iteration %d", i)
--- a/pkg/burner/job.go
+++ b/pkg/burner/job.go
@@ -188,7 +188,7 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider, 
 			}
 			if !globalConfig.WaitWhenFinished {
 				elapsedTime := jobEnd.Sub(executedJobs[len(executedJobs)-1].Start).Round(time.Second)
-				log.Infof("Job %s took %v", jobExecutor.Name, elapsedTime)
+				log.Debugf("Job %s took %v", jobExecutor.Name, elapsedTime)
 			}
 			if !jobExecutor.MetricsAggregate {
 				// We stop and index measurements per job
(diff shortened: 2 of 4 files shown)


pkg/burner/create.go
```diff
@@ -116,7 +116,7 @@ func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterati
 			return
 		}
 		if i == iterationStart+iterationProgress*percent {
-			log.Infof("%v/%v iterations completed", i-iterationStart, iterationEnd-iterationStart)
+			log.Debugf("%v/%v iterations completed", i-iterationStart, iterationEnd-iterationStart)
```
rsevilla87: This one can be the more controversial, since kube-burner won't show the job progress anymore with info log-level.

However I think we're in a good situation to remove this message which is useless in the 99% of situations.

pkg/util/namespaces.go
```diff
@@ -38,10 +38,9 @@ func CreateNamespace(clientSet kubernetes.Interface, name string, nsLabels map[s
 		_, err = clientSet.CoreV1().Namespaces().Create(context.TODO(), &ns, metav1.CreateOptions{})
 		if errors.IsForbidden(err) {
 			log.Fatalf("authorization error creating namespace %s: %s", ns.Name, err)
-			return false, err
```
rsevilla87, line 41: Not needed, Fatal already aborts the execution

pkg/burner/create.go
ygalblum: I disagree. I have tests running 10s of iterations with each iteration taking a few minutes as they wait on a condition. So, getting these progress messages (which already do not print on every cycle) is helpful in seeing that the test progresses and not stuck on a waiter

pkg/burner/job.go
```diff
@@ -188,7 +188,7 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
 			}
 			if !globalConfig.WaitWhenFinished {
 				elapsedTime := jobEnd.Sub(executedJobs[len(executedJobs)-1].Start).Round(time.Second)
-				log.Infof("Job %s took %v", jobExecutor.Name, elapsedTime)
+				log.Debugf("Job %s took %v", jobExecutor.Name, elapsedTime)
```
ygalblum, line 191: Why wouldn't you want this in the info log?

pkg/burner/job.go
```diff
@@ -358,24 +358,23 @@ func indexMetrics(uuid string, executedJobs []prometheus.Job, returnMap map[stri
...
 }
 
 func verifyQPSBurst(job *config.Job) {
 	if job.QPS <= 0 {
-		log.Warnf("Invalid QPS (%v); using default: %v", job.QPS, rest.DefaultQPS)
+		log.Warnf("%s: Invalid QPS (%v); using default: %v", job.Name, job.QPS, rest.DefaultQPS)
 		job.QPS = rest.DefaultQPS
 	} else {
-		log.Infof("QPS: %v", job.QPS)
+		log.Infof("%s: QPS: %v", job.Name, job.QPS)
```
ygalblum: This for example can be changed to `Debug`. It just prints a value the user passed

pkg/burner/job.go
```diff
@@ -358,24 +358,23 @@ func indexMetrics(uuid string, executedJobs []prometheus.Job, returnMap map[stri
...
 	}
-
 	if job.Burst <= 0 {
-		log.Warnf("Invalid Burst (%v); using default: %v", job.Burst, rest.DefaultBurst)
+		log.Warnf("%s: Invalid Burst (%v); using default: %v", job.Name, job.Burst, rest.DefaultBurst)
 		job.Burst = rest.DefaultBurst
 	} else {
-		log.Infof("Burst: %v", job.Burst)
+		log.Infof("%s: Burst: %v", job.Name, job.Burst)
```
ygalblum: Same here

pkg/burner/utils.go
```diff
@@ -116,7 +116,7 @@ func (ex *JobExecutor) Verify() bool {
 	var objList *unstructured.UnstructuredList
 	var replicas int
 	success := true
-	log.Info("Verifying created objects")
+	log.Debug("Verifying created objects")
```
ygalblum: This print shows the progress in the test. I think it can stay in `Info`

ygalblum (commented):
Some minor comments. Generally, I think `Debug` message should be used when debugging the test itself. So, prints that describe the progress of the test should be in `Info`

pkg/burner/job.go
rsevilla87: I think is message is not very useful in Info level, it's value is already part of the `jobSummary` metric, and the user can also use the kube-burner log timestamps to have an idea of this value. wdyt?

pkg/burner/create.go
rsevilla87: In order to decrease verbosity in single iteration jobs. What do you think about updating the conditional to:

```golang
if ex.JobIterations > 1 && i == iterationStart+iterationProgress*percent {
	log.Infof("%v/%v iterations completed", i-iterationStart, iterationEnd-iterationStart)
	percent++
}
```

pkg/burner/job.go
ygalblum: OK, makes sense

pkg/burner/create.go
ygalblum: Sure, makes sense

PR #1002, opened 2025-10-09
Add quick install script

rsevilla87:
## Type of change

- New feature
- Documentation

## Description

Adding quick installation script

- Related Issue #
- Closes #

### DIFF
--- a/hack/install.sh
+++ b/hack/install.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+# shellcheck disable=SC2086 
 # Quick install script for kube-burner
 # Downloads the latest release version based on system architecture and OS
 
@@ -7,7 +8,6 @@ set -euo pipefail
 # Configuration
 REPO="kube-burner/kube-burner"
 INSTALL_DIR="${INSTALL_DIR:-${HOME}/.local/bin/}"
-TEMP_DIR=$(mktemp -d)
 
 # Detect OS
 detect_os() {
@@ -67,7 +67,6 @@ download_and_extract() {
 # Verify installation
 verify_installation() {
   if command -v kube-burner &> /dev/null; then
-    local installed_version
     echo "kube-burner is now available in your PATH, installed at ${INSTALL_DIR}"
   else
     echo "kube-burner installed to ${INSTALL_DIR}, but not found in PATH"


README.md
```diff
@@ -17,21 +17,25 @@ Kube-burner is a binary application written in golang that makes extensive usage
...
-## Code of Conduct
-
-This project is for everyone. We ask that our users and contributors take a few minutes to review our [Code of Conduct](./code-of-conduct.md).
 
 ## Documentation
 
 Documentation is [available here](https://kube-burner.github.io/kube-burner/)
 
-## Downloading Kube-burner
+## Quick start
```
ygalblum, line 25: I don't think you should remove the links to the released binaries

hack/install.sh
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env bash
+# Quick install script for kube-burner
+# Downloads the latest release version based on system architecture and OS
+
+set -euo pipefail
+
+# Configuration
+REPO="kube-burner/kube-burner"
+INSTALL_DIR="${INSTALL_DIR:-${HOME}/.local/bin/}"
```
ygalblum, line 10: We need to document the `INSTALL_DIR`:
1. Let users know they can set it
2. Tell users to make sure it's in the `PATH` (not after the installation is deemed as failed)

hack/install.sh
```diff
@@ -0,0 +1,136 @@
...
+      ;;
+    *)
+      echo "Unsupported operating system: ${os}"
+      exit 1
+      ;;
+  esac
+}
+
+# Detect architecture
+detect_arch() {
```
ygalblum: Taken from the Makefile:
`uname -m | sed s/aarch64/arm64/ | sed s/x86_64/amd64/`
Also, from what I can tell, we don't have an `armv7` binary

hack/install.sh
rsevilla87: good call, fixed

README.md
rsevilla87: mm, I probably deleted it accidentally

hack/install.sh
```diff
@@ -0,0 +1,106 @@
...
+  mkdir -p ${INSTALL_DIR}
+  echo "Downloading kube-burner ${version} for ${os}/${arch}..."
+  echo "URL: ${download_url}"
+  curl -sL -f "${download_url}" | tar xz -C ${INSTALL_DIR} kube-burner
+}
+
+# Verify installation
+verify_installation() {
+  if command -v kube-burner &> /dev/null; then
+    local installed_version
```
ygalblum: `installed_version` not used

PR #1003, opened 2025-10-13
Filter not compatible measurements

rsevilla87:
## Type of change

- Optimization

## Description

Adding the necessary logic to skip measurements when they are not compatible with a specific job type.

- Related Issue #
- Closes #

### DIFF
--- a/pkg/measurements/datavolume_latency.go
+++ b/pkg/measurements/datavolume_latency.go
@@ -305,8 +305,6 @@ func (dv *dvLatency) getLatency(normLatency any) map[string]float64 {
 }
 
 func (dv *dvLatency) IsCompatible() bool {
-	if _, exists := supportedDvLatencyJobTypes[dv.JobConfig.JobType]; exists {
-		return true
-	}
-	return false
+	_, exists := supportedDvLatencyJobTypes[dv.JobConfig.JobType]
+	return exists
 }
--- a/pkg/measurements/factory.go
+++ b/pkg/measurements/factory.go
@@ -126,7 +126,7 @@ func (msf *MeasurementsFactory) NewMeasurements(jobConfig *config.Job, kubeClien
 			log.Fatalf("Failed to create measurement [%s]: %v", name, err)
 		}
 		msInstance := mf.NewMeasurement(jobConfig, clientSet, restConfig, embedCfg)
-		if !msInstance.IsCompatible() {
+		if !jobConfig.MetricsAggregate && !msInstance.IsCompatible() {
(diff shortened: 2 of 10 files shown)


rsevilla87:
> This test will break existing tests. With the introduction of `metricsAggregate`, users can use a "non supported" job type (e.g. read) as a marker for starting the measurement. See: https://github.com/kube-burner/kube-burner-ocp/blob/a084051c0836715072c9c18866ab4e280b9ecd57/cmd/config/virt-ephemeral-restart/virt-ephemeral-restart.yml#L165 as an example I'm setting "Request Changes" to make sure it doesn't accidentally go in before clearing this discussion

oh, right, Im not very used to that functionality, we can add an extra conditional to skip the measurement to ensure `metricsAggregate` is not set in the job.

ygalblum:
> we can add an extra conditional to skip the measurement to ensure metricsAggregate is not set in the job.

SGTM

pkg/measurements/factory.go
```diff
@@ -124,10 +125,14 @@ func (msf *MeasurementsFactory) NewMeasurements(jobConfig *config.Job, kubeClien
 		if err != nil {
 			log.Fatalf("Failed to create measurement [%s]: %v", name, err)
 		}
-		ms.MeasurementsMap[name] = mf.NewMeasurement(jobConfig, clientSet, restConfig, embedCfg)
+		msInstance := mf.NewMeasurement(jobConfig, clientSet, restConfig, embedCfg)
+		if !msInstance.IsCompatible() {
```
ygalblum: You already have the `jobConfig`, why not check `MetricsAggregate` before calling `IsCompatible`?

pkg/measurements/datavolume_latency.go
```diff
@@ -297,3 +303,10 @@ func (dv *dvLatency) getLatency(normLatency any) map[string]float64 {
 		string(cdiv1beta1.DataVolumeReady):   float64(dataVolumeMetric.DVReadyLatency),
 	}
 }
+
+func (dv *dvLatency) IsCompatible() bool {
+	if _, exists := supportedDvLatencyJobTypes[dv.JobConfig.JobType]; dv.JobConfig.MetricsAggregate && exists {
+		return true
+	}
+	return false
```
ygalblum: Assuming you move checking `MetricsAggregate` to the caller, this can be:
```suggestion
	_, exists := supportedDvLatencyJobTypes[dv.JobConfig.JobType]
	return exists
```

pkg/measurements/factory.go
rsevilla87: you're right!, I missed it, too much nesting lol

ygalblum (approved):
LGTM

PR #1007, opened 2025-10-20
Set info level to job duration message

rsevilla87:
## Type of change

- Optimization

## Description

This message is useful for some users.


jeniferh:
Thank you!! LGTM, this line is helpful in past Virt testing scenarios.

PR #1022, opened 2025-11-07
Standarize kube-burner labels

rsevilla87:
## Type of change

- Refactor
- Optimization

## Description

The labels kube-burner-job, kube-burner-uuid... have been standardized to kube-burner.io/job, kube-burner.io/uuid, etc.

Also optimize the syntax to stop using unnecessary `labels.SelectorFromSet`.

- Related Issue #
- Closes #

### DIFF
--- a/pkg/burner/create.go
+++ b/pkg/burner/create.go
@@ -341,9 +341,9 @@ func (ex *JobExecutor) churnNamespaces(ctx context.Context) {
 	// Create timer for the churn duration
 	timer := time.After(ex.ChurnConfig.Duration)
 	nsLabels := labels.Set{
-		"kube-burner-job":   ex.Name,
-		"kube-burner-uuid":  ex.uuid,
-		"kube-burner-runid": ex.runid,
+		config.KubeBurnerLabelJob:   ex.Name,
+		config.KubeBurnerLabelUUID:  ex.uuid,
+		config.KubeBurnerLabelRunID: ex.runid,
 	}
 	// List namespaces to churn
 	jobNamespaces, err := ex.clientSet.CoreV1().Namespaces().List(ctx, metav1.ListOptions{LabelSelector: labels.SelectorFromSet(nsLabels).String()})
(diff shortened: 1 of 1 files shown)


pkg/burner/create.go
```diff
@@ -133,10 +133,9 @@ func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterati
 				}
 			}
 			labels := map[string]string{
-				"kube-burner-uuid":                 ex.uuid,
-				"kube-burner-job":                  ex.Name,
-				"kube-burner-index":                strconv.Itoa(objectIndex),
```
ygalblum, line 147: Why did you remove `kube-burner-index` without a replacement?

pkg/burner/namespaces.go
```diff
@@ -16,36 +16,42 @@ package burner
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 // Cleanup resources specific to kube-burner for a given iteration range
 func CleanupIterations(ctx context.Context, ex JobExecutor, iterationStart, iterationEnd int, namespace string) {
 	for i := iterationStart; i < iterationEnd; i++ {
-		labelSelector := fmt.Sprintf("kube-burner-job=%s,%s=%d", ex.Name, config.KubeBurnerLabelJobIteration, i)
+		labelSelector := labels.Set{
```
ygalblum: If it's only used as a sting why not store it as a string once as use it as is? Same for `CleanupNamespacesUsingGVR`

ygalblum (commented):
While I agree with this change and think that standardizing and using `kube-burner.io/` prefix is the right thing to do, this is a breaking change. All usages that rely on these labels will break.
Are we OK with that? Does this change require a `v2`?

pkg/burner/namespaces.go
rsevilla87: Im not sure about the balance on when when to use labels.Set or a labelSelector string tbh 🤔

ygalblum: I think you use the set to accumulate the different labels and make it a sting before using it. For example, if you can pass the set between functions that update it. Then it is turned into a string when it is used. In this case, I noticed that the set itself is not really used, only to make a sting. Furthermore, you make a string twice. So, I don't see the need to keep the set

rsevilla87: Noted, I've updated the logic to prevent using labels.Set in not necessary cases

vishnuchalla:
@rsevilla87 as @ygalblum suggested, I think it's better to take it as a V2. Immediate things that I can think might be broken are cpt-dashboard, orion, grafana dashboards we have and others that rely on kube-burner queries (.i.e other AI related work).

rsevilla87:
alright, but that's out of scope of this PR, #1033 addresses it

pkg/burner/create.go
rsevilla87: fixed

PR #1025, opened 2025-11-07
Stop all watchers and wait for cache sync before collecting the measurements

rsevilla87:
## Type of change

- Bug fix

## Description

Thanks to a flacky test that can be observed at https://github.com/kube-burner/kube-burner/actions/runs/19163907877/job/54780056514?pr=1024#step:5:137, its been detected that there're situations where no documents are indexed by the podLatency measurement.

When stopping the pod latency measurement it's possible that the pod watcher hasn't received the pod yet, for that reason we should normalize the metrics after stopping the watcher.

- Related Issue #
- Closes #

### DIFF
--- a/test/k8s/kube-burner-measurements.yml
+++ b/test/k8s/kube-burner-measurements.yml
@@ -14,7 +14,7 @@ jobs:
   - name: precedence-measurements
     measurements:
       - name: podLatency
-    jobIterations: 5
+    jobIterations: 1
     namespace: precedence-measurements
     namespacedIterations: true
     preLoadImages: false
@@ -32,7 +32,7 @@ jobs:
     measurements:
       - name: serviceLatency
         svcTimeout: 5s
-    jobIterations: 5
+    jobIterations: 1
     namespace: merge-measurements
     namespacedIterations: true
     preLoadImages: false
--- a/pkg/measurements/base_measurement.go
+++ b/pkg/measurements/base_measurement.go
@@ -96,7 +96,7 @@ func (bm *BaseMeasurement) stopWatchers() {
 
 func (bm *BaseMeasurement) StopMeasurement(normalizeMetrics func() float64, getLatency func(any) map[string]float64) error {
 	var err error
-	defer bm.stopWatchers()
+	bm.stopWatchers()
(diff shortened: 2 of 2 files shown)


ygalblum:
I think the issue is here:
https://github.com/kube-burner/kube-burner/blob/e59e46e1bf25b18d4bbe1277acb7de09cfb64700/pkg/measurements/base_measurement.go#L99
IIUC stopping the watchers is deferred, so all calculations are already complete by the time we reach `cache.WaitForCacheSync()`.
Makes sense?

rsevilla87:
> I think the issue is here:
> 
> https://github.com/kube-burner/kube-burner/blob/e59e46e1bf25b18d4bbe1277acb7de09cfb64700/pkg/measurements/base_measurement.go#L99
> 
> IIUC stopping the watchers is deferred, so all calculations are already complete by the time we reach `cache.WaitForCacheSync()`.
> Makes sense?



ohh, that's a good point!, didn't notice that, let me fix it!

test/k8s/kube-burner-measurements.yml
```diff
@@ -17,6 +17,7 @@ jobs:
     jobIterations: 1
     namespace: precedence-measurements
     namespacedIterations: true
+    preLoadImages: false
```
ygalblum, line 20: Is this change still needed?

ygalblum (commented):
Please also update the PR title and description so that the squashed commit message will reflect the actual change

test/k8s/kube-burner-measurements.yml
rsevilla87: not strictly necessary for the purpose of this PR, optimizes the test duration though

ygalblum:
Regarding the PR name. The flaky test was not "fixed" by this change, the test actually raised the issue. The fix is: 
```
Stop all watchers and wait for cache sync before collecting the measurements
```

rsevilla87:
> Regarding the PR name. The flaky test was not "fixed" by this change, the test actually raised the issue. The fix is:
> 
> ```
> Stop all watchers and wait for cache sync before collecting the measurements
> ```

Thanks for the suggestions, I've updated the title along with the PR description

PR #1027, opened 2025-11-12
Remove gcTimeout parameter

rsevilla87:
## Type of change

- Optimization

## Description

This config param is not needed anymore as a context is now being passed to the garbage collection method (it was introduced to prevent garbage collection to run forever)

- Related Issue #
- Closes #


rsevilla87:
> Just to understand, what times out the GC now? The global test timeout?

correct

PR #1028, opened 2025-11-12
Add map to jobExecutor struct to track namespaces creation

rsevilla87:
## Type of change

- Optimization

## Description

By adding a map field to the jobExecutor struct, we can reduce the number of arguments in multiple functions, making the code clearer and more intuitive.

- Related Issue #
- Closes #

### DIFF
--- a/pkg/burner/executor.go
+++ b/pkg/burner/executor.go
@@ -61,6 +61,7 @@ type JobExecutor struct {
 func newExecutor(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider, job config.Job, embedCfg *fileutils.EmbedConfiguration) JobExecutor {
 	ex := JobExecutor{
 		Job:               job,
+		createdNamespaces: make(map[string]bool),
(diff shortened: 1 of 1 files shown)


ygalblum:
How big do we expect `createdNamespaces` to be? `slices.Contains` is an O(N) operation while using a map gives us an O(1). Why not define `createdNamespaces` as a map?

vishnuchalla:
> How big do we expect `createdNamespaces` to be? `slices.Contains` is an O(N) operation while using a map gives us an O(1). Why not define `createdNamespaces` as a map?

I have the same question too.

rsevilla87:
Good point @ygalblum @vishnuchalla , I've updated the logic with maps[string]bool, which is even more simple.

ygalblum:
@rsevilla87 the linter is complaining. Also rebase on main to get the CI test fix

rsevilla87:
> @rsevilla87 the linter is complaining. Also rebase on main to get the CI test fix

fixed and rebased

ygalblum (approved):
LGTM

PR #1030, opened 2025-11-13
Add unit to measurements report

rsevilla87:
## Type of change

- Optimization
  
## Description

Units were missing in the message.

- Related Issue #
- Closes #

### DIFF
--- a/pkg/measurements/factory.go
+++ b/pkg/measurements/factory.go
@@ -176,12 +176,3 @@ func (ms *Measurements) Index(jobName string, indexerList map[string]indexers.In
 		measurement.Index(jobName, indexerList)
 	}
 }
-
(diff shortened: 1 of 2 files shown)


pkg/measurements/factory.go
```diff
@@ -176,12 +176,3 @@ func (ms *Measurements) Index(jobName string, indexerList map[string]indexers.In
 		measurement.Index(jobName, indexerList)
 	}
 }
-
```
ygalblum, line 179: How is this change related to this PR?

rsevilla87: not actually related :(, saw that function is not used anywhere

ygalblum: OK

ygalblum (approved):
LGTM

PR #1032, opened 2025-11-14
Fix flacky test by checking command rc rather than value 

rsevilla87:
## Type of change

- Bug fix
- CI

## Description

It's possible for reported latencies to be 0 (k8s API rounds down timestamps), for that reason we shouldn't compare the value of the field with 0.
Checking the command rc (passing the -e flag to jq) is a better approach.

- Related Issue #
- Closes #


vishnuchalla (approved):
lgtm

PR #1039, opened 2025-11-25
Parallel CI testing

rsevilla87:
## Type of change

- CI

## Description

Update logic to allow parallel execution of CI jobs

### DIFF
--- a/test/test-k8s.bats
+++ b/test/test-k8s.bats
@@ -48,7 +48,7 @@ setup() {
   export TIMESERIES_INDEXER=""
   export CRD=""
   export SVC_LATENCY=""
-  export JOB_NAME=ci-${BATS_TEST_NUMBER}
+  export JOB_NAME=kube-burner-ci-${BATS_TEST_NUMBER}
 }
 
 teardown() {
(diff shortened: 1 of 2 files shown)


test/k8s/kube-burner.yml
```diff
@@ -40,7 +42,7 @@ metricsEndpoints:
 {{ end }}
 
 jobs:
-  - name: namespaced
+  - name: namespaced-{{ .BATS_TEST_NUMBER }}
```
ygalblum: Use an environment variable to get the complete job name instead of having to define it twice (here and in in test script)

rsevilla87: fixed

test/helpers.bash
```diff
@@ -174,7 +174,7 @@ check_running_pods() {
 }
 
 check_running_pods_in_ns() {
-  running_pods=$(kubectl get pod -n "${1}" -l kube-burner.io/job=namespaced --field-selector=status.phase==Running -o json | jq '.items | length')
+  running_pods=$(kubectl get pod -n "${1}" -l kube-burner.io/job=${JOB_NAME} --field-selector=status.phase==Running -o json | jq '.items | length')
```
ygalblum: I think it's cleaner that the function gets the job name as a parameter

test/test-k8s.bats
```diff
@@ -45,10 +41,15 @@ setup() {
   export LOCAL_INDEXING=""
   export ALERTING=""
   export TIMESERIES_INDEXER=""
+  export SVC_LATENCY=""
+  export GC=true
+  export PRELOAD_IMAGES=false
+  export CHURN_CYCLES=0
+  export CHURN_MODE=""
```
ygalblum: Should we zero `JOB_NAME` to be on the safe side? Or since when it is set, it's set to the same value, should we just set it here?

test/test-k8s.bats
```diff
@@ -64,108 +65,119 @@ teardown_file() {
...
+  export JOB_NAME=ci-${BATS_TEST_NUMBER}
+  run_cmd ${KUBE_BURNER} init -c kube-burner.yml --uuid=${UUID} --log-level=debug -e metrics-endpoints.yaml
+  check_file_list ${METRICS_FOLDER}/jobSummary.json ${METRICS_FOLDER}/podLatencyQuantilesMeasurement-${JOB_NAME}.json
+  check_destroyed_ns kube-burner.io/uuid=${UUID}
+  check_destroyed_pods default kube-burner.io/uuid=${UUID}
 }
 
 @test "kube-burner index: local-indexing=true; tarball=true" {
-  run_cmd ${KUBE_BURNER} index --uuid="${UUID}" -u http://localhost:9090 -m "metrics-profile.yaml,metrics-profile.yaml" --tarball-name=metrics.tgz --start="$(date -d "-2 minutes" +%s)"
+  sleep 1m
```
ygalblum: Why is this needed?

test/helpers.bash
rsevilla87: This function is going away with https://github.com/kube-burner/kube-burner/pull/1040, for that reason I didn't touch it

ygalblum: OK, np

test/test-k8s.bats
rsevilla87: As tests start now in parallel in random order, I've seen situations where no metrics are returned from Prometheus b/c this job is one of the first ones initiated after spinning up the kind cluster.

ygalblum: Ok, so, please add a comment about it so that it doesn't get removed in the future without considering this issue

PR #1040, opened 2025-11-25
Shift to verify_object_count function

rsevilla87:
## Type of change

- Optimization
- CI

## Description

We can remove lot of code by using the common verify_object_count() function in helpers.bash 

- Related Issue #
- Closes #

### DIFF
--- a/pkg/measurements/service_latency.go
+++ b/pkg/measurements/service_latency.go
@@ -117,8 +117,9 @@ func (s *serviceLatency) handleCreateSvc(obj any) {
 		if err := s.waitForEndpointSlices(svc); err != nil {
 			log.Fatal(err)
 		}
-		// This conditionals prevents a nil pointer dereference panic because
-		// the measurement was stopped and this goroutine is still ongoing as it performs polling
+		// This conditionals prevents a nil pointer dereference panic that can happen when the
+		// measurement is stopped by end of job J, and then the benchmark continues with job J+1,
+		// so this goroutine can still be runnning in the background as it performs polling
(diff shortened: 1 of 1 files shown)


test/test-k8s.bats
```diff
@@ -132,12 +116,9 @@ teardown_file() {
...
 
-@test "kube-burner init: crd" {
-  kubectl apply -f objectTemplates/burnerTest-crd.yml
-  sleep 5
-  run_cmd ${KUBE_BURNER} init -c kube-burner-crd.yml --uuid="${UUID}"
-  kubectl delete -f objectTemplates/storageclass.yml
-  kubectl delete -f objectTemplates/burnerTest-crd.yml
+@test "kube-burner init: delete=true" {
+  run_cmd ${KUBE_BURNER} init -c kube-burner-delete.yml --uuid "${UUID}" --log-level=debug
+  verify_object_count namespace 0 "" kube-burner-uuid="${UUID}
```
ygalblum: The next test runs the same file. Can we move the `verify_object_count` there and remove this test?

test/test-k8s.bats
```diff
@@ -178,17 +158,17 @@ teardown_file() {
 }
 
 @test "kube-burner log file output" {
```
ygalblum: This check is already covered in a different test. Maybe just add a comment stating that and remove this test

rsevilla87: correct, I moved the logic to another test but I forgot to remove it. :)

ygalblum (approved):
LGTM

PR #1044, opened 2025-11-26
Capture errors to act in case of timeout

### LINKED ISSUE #742: [RFE] Report measurements also when job times out

vishnuchalla:
## Type of change

- Refactor
- New feature

## Description

- Instead of using fatal logs in wait methods, errors are now returned to the caller
- Fixed Lint error by refactoring `RunCreateJob` to reduce function complexity

- Closes #742 

## Testing
CI tests and verified in local.

### DIFF
--- a/cmd/kube-burner/kube-burner.go
+++ b/cmd/kube-burner/kube-burner.go
@@ -83,6 +83,7 @@ func initCmd() *cobra.Command {
 	var userDataFile string
 	var allowMissingKeys bool
 	var rc int
+	var setValues []string
 	cmd := &cobra.Command{
 		Use:   "init",
 		Short: "Launch benchmark",
@@ -108,6 +109,10 @@ func initCmd() *cobra.Command {
 			if !skipLogFile {
 				util.SetupFileLogging(uuid)
 			}
+			setVars, err := config.ParseSetValues(setValues)
+			if err != nil {
+				log.Fatal(err.Error())
+			}
 			kubeClientProvider := config.NewKubeClientProvider(kubeConfig, kubeContext)
 			clientSet, _ = kubeClientProvider.DefaultClientSet()
 			configFileReader, err := fileutils.GetWorkloadReader(configFile, nil)
@@ -121,7 +126,7 @@ func initCmd() *cobra.Command {
 					log.Fatalf("Error reading user data file %s: %s", userDataFile, err)
 				}
 			}
-			configSpec, err := config.ParseWithUserdata(uuid, timeout, configFileReader, userDataFileReader, allowMissingKeys, nil)
+			configSpec, err := config.ParseWithUserdata(uuid, timeout, configFileReader, userDataFileReader, allowMissingKeys, nil, setVars)
 			if err != nil {
 				log.Error("Config error")
 				fmt.Printf("%s", err.Error())
@@ -159,6 +164,7 @@ func initCmd() *cobra.Command {
 	cmd.Flags().StringVar(&userDataFile, "user-data", "", "User provided data file for rendering the configuration file, in JSON or YAML format")
 	cmd.Flags().BoolVar(&allowMissingKeys, "allow-missing", false, "Do not fail on missing values in the config file")
 	cmd.Flags().BoolVar(&skipLogFile, "skip-log-file", false, "Skip writing to a log file")
+	cmd.Flags().StringSliceVar(&setValues, "set", []string{}, "Set arbitrary key=value pairs to override values in the config file")
 	cmd.Flags().SortFlags = false
 	cmd.MarkFlagsMutuallyExclusive("config", "configmap")
 	return cmd
--- a/docs/cli/index.md
+++ b/docs/cli/index.md
@@ -48,6 +48,7 @@ This is the main subcommand; it triggers a new kube-burner benchmark and it supp
 - `user-metadata`: YAML file path containing custom user-metadata to be indexed along with the `jobSummary` document.
 - `user-data`: YAML or JSON file path containing input variables for rendering the configuration file.
 - `allow-missing`: Allow missing keys in the config file. Needed when using the [`default`](https://masterminds.github.io/sprig/defaults.html) template function
+- `set`: Set arbitrary `key=value` pairs to override values in the configuration file. Similar to Helm's `--set` flag, this allows you to override base YAML values directly from the command line. Multiple values can be specified by separating them with commas or by using multiple `--set` flags. Nested keys are supported using dot notation, and array indices can be used with numeric keys (e.g., `jobs.0.name=test`).
 
 !!! Note "Prometheus authentication"
     Both basic and token authentication methods need permissions able to query the given Prometheus endpoint.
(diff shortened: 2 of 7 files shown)


rsevilla87:
Needs rebase @vishnuchalla

pkg/burner/job.go
```diff
@@ -81,6 +82,8 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
 	res := make(chan int, 1)
 	uuid := configSpec.GlobalConfig.UUID
 	globalConfig := configSpec.GlobalConfig
+	globalWaitMap := make(map[string][]string)
+	executorMap := make(map[string]JobExecutor)
```
rsevilla87: We don't need these variables, the logic is simpler without them.

pkg/burner/job.go
```diff
@@ -370,23 +389,39 @@ func newExecutorList(configSpec config.Spec, kubeClientProvider *config.KubeClie
...
+	close(errChan)
+
+	var errs []error
+	for err := range errChan {
+		errs = append(errs, err...)
+	}
+	if len(errs) > 0 {
+		return errs
+	}
+	return nil
```
rsevilla87: The zero value of an empty array is nil:

```suggestion
  return errs
```

pkg/burner/job.go
```diff
@@ -370,23 +389,39 @@ func newExecutorList(configSpec config.Spec, kubeClientProvider *config.KubeClie
...
+				if err := executor.waitForObjects(ns); err != nil {
+					errChan <- err
+				}
 			}(ns)
 		}
 		wg.Wait()
 	}
+	close(errChan)
+
+	var errs []error
```
rsevilla87: Please, create this variable at the beginning of the method

pkg/burner/create.go
```diff
@@ -162,7 +163,9 @@ func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterati
 			if !ex.NamespacedIterations || !namespacesWaited[ns] {
 				log.Infof("Waiting up to %s for actions to be completed in namespace %s", ex.MaxWaitTimeout, ns)
 				wg.Wait()
-				ex.waitForObjects(ns)
+				if errs := ex.waitForObjects(ns); len(errs) > 0 {
```
rsevilla87: As the zero value of an empty array is nil, I think is cleaner to compare with nil, as any other error

pkg/burner/job.go
rsevilla87: In fact, they were recently removed at https://github.com/kube-burner/kube-burner/pull/1028

vishnuchalla: Looked the main stream and updated accordingly.

pkg/burner/job.go
vishnuchalla: Had to rewrite this logic with new changes.

pkg/burner/job.go
vishnuchalla: Correct, Updated in all the code changes invovled.

pkg/burner/create.go
```diff
@@ -98,7 +98,7 @@ func (ex *JobExecutor) setupCreateJob() {
 }
 
 // RunCreateJob executes a creation job
-func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterationEnd int) {
+func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterationEnd int, waitListNamespaces *[]string) []error {
```
rsevilla87: waitListNamespaces is not needed anymore

vishnuchalla: Interesting, it was an unused variable and linter did not complain about it.

pkg/burner/create.go
```diff
@@ -252,6 +235,51 @@ func (ex *JobExecutor) replicaHandler(ctx context.Context, labels map[string]str
 	wg.Wait()
 }
 
+// waitForCompletion waits for objects to be ready across the relevant namespaces
+func (ex *JobExecutor) waitForCompletion(iterationStart, iterationEnd int, ns string, namespacesWaited map[string]bool) []error {
+	log.Infof("Waiting up to %s for actions to be completed", ex.MaxWaitTimeout)
+	// This semaphore limits the maximum number of concurrent goroutines
+	sem := make(chan int, int(ex.restConfig.QPS))
+	errChan := make(chan []error, 1)
```
rsevilla87, line 243: What if there're errors waiting in more than 1 namespace? i.e: simultaneous context timeout in multiple go routines

vishnuchalla: Good question, I had the same thought about this when I looked into previous draft from Niladri. I think one error must be sufficient to take action in this case as these are errors related to wait for objects in namespace which turn out to be for a same reason most of the times. So instead of having the same set of errors pile up creating noise, just having 1 error from this category should be fine. Let me know if you have any alternate suggestions here.

rsevilla87: Just wonder if the code would hang in that case,

pkg/burner/create.go
rsevilla87: yeah, I think that linter doesn't complain about unused arguments

pkg/burner/create.go
rsevilla87: Just checked that the goroutines get stuck until having room in the channel and panics when the channel is closed https://go.dev/play/p/AIZ0NTpm0wn

vishnuchalla: I do not see any possibility of hanging because of `errChan`. Because we have `sem` to regulate the concurrency and once the channel is full simply the other errors are discarded. And about the snippet you shared above, its different from what we have because close is done early in the routine which is different from our case which regulates concurrency using `sem` and wait for all parallel routines and closes the channel. 

If at all there is a change for us to hang, It should be because of  `sem := make(chan int, int(ex.restConfig.QPS))` if we set QPS to 0 in our config which also seem to be the default. Though there is a chance of that vulnerability client-go is handling it for us by setting a non-zero value as default [here](https://github.com/kubernetes/client-go/blob/46360b527ebcd4cb7e99ae1770ad9a613bb0f128/rest/config.go#L122)

rsevilla87: good point, if the channel is full, the select statement will fallback to the default branch, and finish the goroutine smoothly

Issue #1019, opened 2025-11-05
[RFE] Node processes profiling


rsevilla87:
**Is your feature request related to a problem? Please describe.**

At the moment kube-burner supports pprof collection of golang processes via https://github.com/kube-burner/kube-burner/blob/main/docs/measurements/index.md#pprof-collection, this collection is performed by connecting to the label matching pods and execute a curl command to the specified endpoint.

With the previous implementation is not possible to collect golang profiling information from node processes like kubelet or the container runtime.

**Describe the solution you'd like**

Implement the necessary logic to provide kube-burner the ability to collect node processes profiling information. 

Adding an extra optional field to the measurment

```yaml
  measurements:
  - name: pprof
    pprofInterval: 30m
    pprofDirectory: pprof-data
    nodeAffinity:
      node-role.kubernetes.io/worker: ""
    pprofTargets:
    - name: kubelet
      url: https://localhost:6443/debug/pprof/heap
```

```mermaid
flowchart TD
    A([Start kube-burner job])
    B([Start pprof measurement])
    C{nodeAffinity field set?}
    D([Spin up DaemonSet with nodeAffinity<br/])
    E([Collect profiling information])
    F([Stop pprof measurement and remove DaemonSet if required])
    G([End job])

    A --> B --> C
    C -->|Yes| D --> E
    C -->|No| E
    E --> F --> G
```

 ---

- By default kubelet listens in the port 10250 and requires client certificate authentication, this certificate can be provided  to the daemonSet pods by a serviceAccount with the required permissions (TBD, cluster-admin works)
- On the other hand cri-o listens in the `/var/run/crio/crio.sock` unix socket file, so the pods will need access to this path.

wolf-06:
Hey @rsevilla87 
Can I work on it ?

sanjay7178:
Hi @rsevilla87,

I would like to work on this feature. I’ve reviewed the description and the flowchart, and I understand the goal is to extend the pprof collection logic to support node processes (like `kubelet` and `crio`) by deploying a DaemonSet with the necessary `nodeAffinity` and permissions.

If the design is finalized, could you please assign this issue to me? I'll start by exploring how the current measurements are configured and how to integrate the DaemonSet deployment.

rsevilla87:
> Hi [@rsevilla87](https://github.com/rsevilla87),
> 
[...]

PR #1045, opened 2025-11-27
Implement pprof DaemonSet deployment and management for profiling to collect from Node Processes : Kubelet and CRIO

sanjay7178:
…collect from Node Processes : Kubelet and CRIO

## Type of change

- New feature

## Description

- Added optional yaml node affinity entry in Measurement struct  at types.go 
-  Add pkg/measurements/deployment_set.go to deploy daemon set pods on nodes based on prior node label  
- deployment_set.go has all helper functions to create daemon set pods in specific node based on pprofTargets such as kubelet and cri-o . These helper functions  ensures to create a deployment set with cluster role with kubelet permissions , service account ,  cluster role binding . The volume mounts are dynamic based on pprofTargets 
- Refactored  pprof.go to reduce cyclomatic complexity , wrote buildPProfRequest() to simplify it . 
- As mentioned in #1019 , until the daemon set deployment is ready , it waits for to start the pprof 
-  After the pprof starts the getPods will use the daemoset pods for pprof command exec 
- Added additional verbose logs for better debugging

> I used k0s with CRI-O runtime for this PR on a pair of debian vm's running on qemu , added additional `examples/workloads/kubelet-density/kubelet-density-crio.yml` workflow to simulate the CRI-O and kubelet pprof tests .

> I tried using running these pprof tests under a kind cluster , but failed to run for kubelet . As kind clusters runs of docker runtime and it throws a  403 error .

> Also CRI-O doesn't support CPU profiling as per some go limitations  

- Closes https://github.com/kube-burner/kube-burner/issues/1019

### DIFF
--- a/pkg/measurements/pprof.go
+++ b/pkg/measurements/pprof.go
@@ -29,6 +29,7 @@ import (
 	"github.com/cloud-bulldozer/go-commons/v2/indexers"
 	"github.com/kube-burner/kube-burner/v2/pkg/config"
 	"github.com/kube-burner/kube-burner/v2/pkg/measurements/types"
+	"github.com/kube-burner/kube-burner/v2/pkg/util"
 	"github.com/kube-burner/kube-burner/v2/pkg/util/fileutils"
 	log "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
@@ -43,7 +44,6 @@ import (
 
 type pprof struct {
 	BaseMeasurement
-
 	stopChannel chan bool
 }
 
@@ -161,13 +161,12 @@ func (p *pprof) getPProf(wg *sync.WaitGroup, first bool) {
 			wg.Add(1)
 			go func(target types.PProftarget, pod corev1.Pod) {
 				defer wg.Done()
-
-				// Determine identifier for filename
-				identifier := pod.Name
-
-				pprofFile := fmt.Sprintf("%s-%s-%d.pprof", target.Name, identifier, time.Now().Unix())
-				f, err := os.Create(path.Join(p.Config.PProfDirectory, pprofFile))
+				pprofFile := fmt.Sprintf("%s-%s-%d.pprof", target.Name, pod.Name, time.Now().Unix())
+				if target.LabelSelector == nil {
+					pprofFile = fmt.Sprintf("%s-%s-%d.pprof", target.Name, pod.Spec.NodeName, time.Now().Unix())
+				}
 				var stderr bytes.Buffer
+				f, err := os.Create(path.Join(p.Config.PProfDirectory, pprofFile))
 				if err != nil {
 					log.Errorf("Error creating pprof file %s: %s", pprofFile, err)
 					return
@@ -219,42 +218,15 @@ func (p *pprof) buildPProfRequest(target types.PProftarget, pod corev1.Pod) ([]s
 	// Build command based on target type
 	if p.getPprofNodeTargets(target) != nil {
 		// Node-level collection via DaemonSet
-		if strings.HasPrefix(target.URL, "unix://") {
-			// Unix socket (CRI-O)
-			socketPath := strings.TrimPrefix(target.URL, "unix://")
-
-			// Extract the pprof path from target.URL or use default
-			pprofPath := "/debug/pprof/profile"
-			if strings.Contains(target.Name, "heap") {
-				pprofPath = "/debug/pprof/heap"
-			} else if strings.Contains(target.Name, "goroutine") {
-				pprofPath = "/debug/pprof/goroutine"
-			}
-
-			// Convert duration to seconds for CPU profiling
-			seconds := int(p.Config.PProfInterval.Seconds())
-			if seconds > 300 {
-				seconds = 30 // Cap at 30 seconds for CPU profiling
-			}
-
+		if target.UnixSocketPath != "" {
 			command = []string{"curl", "-fsSLk",
-				"--unix-socket", socketPath,
-				fmt.Sprintf("http://localhost%s?seconds=%d", pprofPath, seconds),
-			}
-			log.Debugf("Using unix socket: %s for %s", socketPath, pprofPath)
+				"--unix-socket", path.Join("/mnt", target.UnixSocketPath), target.URL}
 		} else {
 			// HTTPS endpoint (kubelet) - use insecure since kubelet typically uses self-signed certs
 			command = []string{"sh", "-c",
-				fmt.Sprintf("curl -fsSLk -H \"Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)\" %s", target.URL),
+				fmt.Sprintf("curl -fsSLkH \"Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)\" %s", target.URL),
 			}
 		}
-		pprofReq = p.ClientSet.CoreV1().
-			RESTClient().
-			Post().
-			Resource("pods").
-			Name(pod.Name).
-			Namespace(types.PprofNamespace).
-			SubResource("exec")
 	} else {
 		// Pod-level collection (etcd, api-server, etc.)
 		if target.BearerToken != "" {
@@ -264,14 +236,14 @@ func (p *pprof) buildPProfRequest(target types.PProftarget, pod corev1.Pod) ([]s
 		} else {
 			command = []string{"curl", "-fsSLk", target.URL}
 		}
-		pprofReq = p.ClientSet.CoreV1().
-			RESTClient().
-			Post().
-			Resource("pods").
-			Name(pod.Name).
-			Namespace(pod.Namespace).
-			SubResource("exec")
 	}
+	pprofReq = p.ClientSet.CoreV1().
+		RESTClient().
+		Post().
+		Resource("pods").
+		Name(pod.Name).
+		Namespace(pod.Namespace).
+		SubResource("exec")
 	return command, pprofReq
 }
 
@@ -281,6 +253,20 @@ func (p *pprof) Collect(measurementWg *sync.WaitGroup) {
 
 func (p *pprof) Stop() error {
 	p.stopChannel <- true
+	if p.needsDaemonSet() {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+		defer cancel()
+		err := p.ClientSet.RbacV1().ClusterRoles().Delete(ctx, types.PprofRole, metav1.DeleteOptions{})
+		if err != nil {
+			log.Errorf("Error deleting cluster role %s: %s", types.PprofRole, err)
+			return err
+		}
+		err = util.CleanupNamespaces(ctx, p.ClientSet, fmt.Sprintf("kubernetes.io/metadata.name=%s", types.PprofNamespace))
+		if err != nil {
+			log.Errorf("Error cleaning up namespaces %s: %s", types.PprofNamespace, err)
+			return err
+		}
+	}
 	return nil
 }
 
--- a/pkg/measurements/pprof_ds.go
+++ b/pkg/measurements/pprof_ds.go
@@ -12,7 +12,7 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/utils/ptr"
 )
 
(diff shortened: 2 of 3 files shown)


sanjay7178:
Here are some pprof flamegraphs I've 
- kubelet heap : https://pprof.me/713a6120c3a734ee9535e823b9c49dd4/?profileType=profile%3Aalloc_objects%3Acount%3Aspace%3Abytes
- cri-o heap : https://pprof.me/4145c33843fdd4b60df0fe92c74d29c3/?profileType=profile%3Aalloc_objects%3Acount%3Aspace%3Abytes
- cri-o allocs : https://pprof.me/c547981e75ace775a7ff3f6968a3130d/?profileType=profile%3Aalloc_objects%3Acount%3Aspace%3Abytes
- cri-o goroutines : https://pprof.me/b7f8f06b2bde81aa5e41865931974189/

pkg/measurements/deployment_set.go
```diff
@@ -0,0 +1,330 @@
...
+	// Create namespace
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: types.PprofNamespace,
+		},
+	}
+	_, err := p.ClientSet.CoreV1().Namespaces().Create(ctx, ns, metav1.CreateOptions{})
+	if err != nil && !strings.Contains(err.Error(), "already exists") {
+		return fmt.Errorf("failed to create namespace: %v", err)
+	}
```
rsevilla87:
```suggestion
	if err != nil && !errors.IsAlreadyExists(err) { {
		return fmt.Errorf("failed to create namespace: %v", err)
	}
```

pkg/measurements/deployment_set.go
```diff
@@ -0,0 +1,330 @@
...
+
+	// Create ServiceAccount
+	sa := &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      types.PprofSA,
+			Namespace: types.PprofNamespace,
+		},
+	}
+	_, err = p.ClientSet.CoreV1().ServiceAccounts(types.PprofNamespace).Create(ctx, sa, metav1.CreateOptions{})
+	if err != nil && !strings.Contains(err.Error(), "already exists") {
```
rsevilla87: Same as above

pkg/measurements/types/types.go
```diff
@@ -92,3 +94,11 @@ const (
...
 	SvcLatencyCheckerName = "svc-checker"
 )
+
+const (
+	PprofNamespace   = "pprof-collector"
+	PprofDaemonSet   = "pprof-collector"
+	PprofSA          = "pprof-collector"
+	PprofRole        = "pprof-collector"
+	PprofRoleBinding = "pprof-collector"
+)
```
rsevilla87: Can you add the prefix `kube-burner` to the objects name?

pkg/measurements/deployment_set.go
```diff
@@ -0,0 +1,330 @@
...
+			},
+			{
+				APIGroups: []string{""},
+				Resources: []string{"nodes", "nodes/proxy", "nodes/stats", "nodes/metrics"},
+				Verbs:     []string{"get", "list"},
+			},
+		},
+	}
+	_, err = p.ClientSet.RbacV1().ClusterRoles().Create(ctx, clusterRole, metav1.CreateOptions{})
+	if err != nil && !strings.Contains(err.Error(), "already exists") {
```
rsevilla87: Same as above

pkg/measurements/deployment_set.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	// Create DaemonSet
+	ds := p.buildDaemonSet()
+	_, err = p.ClientSet.AppsV1().DaemonSets(types.PprofNamespace).Create(ctx, ds, metav1.CreateOptions{})
+	if err != nil && !strings.Contains(err.Error(), "already exists") {
+		return fmt.Errorf("failed to create daemonset: %v", err)
+	}
+
+	log.Infof("DaemonSet %s deployed in namespace %s", types.PprofDaemonSet, types.PprofNamespace)
```
rsevilla87:
```suggestion
	log.Infof("pprof-collector: DaemonSet %s deployed in namespace %s", types.PprofDaemonSet, types.PprofNamespace)
```

pkg/measurements/pprof.go
```diff
@@ -99,6 +111,18 @@ func (p *pprof) Start(measurementWg *sync.WaitGroup) error {
 }
 
 func (p *pprof) getPods(target types.PProftarget) []corev1.Pod {
+	// If DaemonSet is deployed and no explicit label selector is provided, use DaemonSet pods
```
rsevilla87, line 112: We should be able to combine previous label selector logic and the new one using node affinities, so the returned  podList should be the combination of both.

rsevilla87: So it's possible that a new field like `pprofNodeTargets` would be needed to distinguish between node  and pods targets

pkg/measurements/deployment_set.go
sanjay7178: the `errors` package from go's standard lib , doesn't have IsAlreadyExists() . do I need to implement from `k8s.io/apimachinery/pkg/api/errors` ?

pkg/measurements/pprof.go
sanjay7178: > We should be able to combine previous label selector logic and the new one using node affinities, so the returned podList should be the combination of both.

As from [this](https://github.com/kube-burner/kube-burner/blob/11d415706c615d96108f8bc8741f0d3f9a1b3e96/pkg/measurements/pprof.go#L113) line  , the podList varies because , it is iterating over the list of pprofTargets . In my implementation i did differentiated/distinguish node affinity based daemonset pods and normal pod based on labelSelector on pprofTarget . So the podList every time would be individual to each pprofTarget .  ie .. merging  node affinity based pods and normal pods to single pod list wouldn't be a better idea .  
https://github.com/kube-burner/kube-burner/blob/11d415706c615d96108f8bc8741f0d3f9a1b3e96/pkg/measurements/pprof.go#L113

rsevilla87: oh, that's right, somehow I forgot this funcion was iterating over pprofTargets

pkg/measurements/pprof.go
```diff
@@ -44,7 +44,8 @@ import (
 type pprof struct {
 	BaseMeasurement
 
-	stopChannel chan bool
+	stopChannel       chan bool
+	daemonSetDeployed bool
```
rsevilla87: This variable is not strictly needed, when the daemonset deployment fails, the measurement gets aborted, so we can use  `p.needsDaemonSet()` rathern than p.daemonSetDeployed

pkg/measurements/pprof.go
```diff
@@ -140,7 +164,17 @@ func (p *pprof) getPProf(wg *sync.WaitGroup, first bool) {
 			wg.Add(1)
 			go func(target types.PProftarget, pod corev1.Pod) {
 				defer wg.Done()
-				pprofFile := fmt.Sprintf("%s-%s-%d.pprof", target.Name, pod.Name, time.Now().Unix())
+
+				// Determine identifier for filename
+				identifier := pod.Name
+				if p.daemonSetDeployed && len(target.LabelSelector) == 0 {
```
rsevilla87: Just ignore `target.LabelSelector`

```suggestion
				if p.daemonSetDeployed {
```

if labelSelector is set accidentally we should use the nodeName identifier as well

pkg/measurements/pprof.go
```diff
@@ -140,7 +164,17 @@ func (p *pprof) getPProf(wg *sync.WaitGroup, first bool) {
...
+					if nodeName := pod.Spec.NodeName; nodeName != "" {
+						identifier = nodeName
+						log.Infof("Collecting pprof from pod %s on node %s", pod.Name, nodeName)
```
rsevilla87: Remove this log line, it's too verbose, we can log it later

pkg/measurements/pprof.go
```diff
@@ -140,7 +164,17 @@ func (p *pprof) getPProf(wg *sync.WaitGroup, first bool) {
 			wg.Add(1)
 			go func(target types.PProftarget, pod corev1.Pod) {
 				defer wg.Done()
-				pprofFile := fmt.Sprintf("%s-%s-%d.pprof", target.Name, pod.Name, time.Now().Unix())
+
+				// Determine identifier for filename
+				identifier := pod.Name
+				if p.daemonSetDeployed && len(target.LabelSelector) == 0 {
+					if nodeName := pod.Spec.NodeName; nodeName != "" {
```
rsevilla87: Under which conditions nodeName is null?

pkg/measurements/pprof.go
```diff
@@ -154,30 +188,19 @@ func (p *pprof) getPProf(wg *sync.WaitGroup, first bool) {
...
-					Resource("pods").
-					Name(pod.Name).
-					Namespace(pod.Namespace).
-					SubResource("exec")
-				log.Debugf("Collecting pprof using URL: %s", req.URL())
-				req.VersionedParams(&corev1.PodExecOptions{
+
+				command, pprofReq := p.buildPProfRequest(target, pod)
+
+				log.Debugf("Collecting pprof using URL: %s", pprofReq.URL())
```
rsevilla87:
```suggestion
				log.Tracef("Collecting pprof using URL: %s", pprofReq.URL())
```

pkg/measurements/pprof.go
```diff
@@ -189,13 +212,78 @@ func (p *pprof) getPProf(wg *sync.WaitGroup, first bool) {
 				if err != nil {
 					log.Errorf("Failed to get pprof from %s: %s", pod.Name, stderr.String())
 					os.Remove(f.Name())
+				} else {
+					log.Infof("Successfully collected pprof data: %s", pprofFile)
```
rsevilla87:
```suggestion
					log.Debugf("Successfully collected pprof data: %s", pprofFile)
```

pkg/measurements/pprof.go
```diff
@@ -189,13 +212,78 @@ func (p *pprof) getPProf(wg *sync.WaitGroup, first bool) {
...
 	}
 	wg.Wait()
 }
 
+func (p *pprof) buildPProfRequest(target types.PProftarget, pod corev1.Pod) ([]string, *rest.Request) {
+	var pprofReq *rest.Request
+	var command []string
+
+	// Build command based on target type
+	if p.daemonSetDeployed && len(target.LabelSelector) == 0 {
```
rsevilla87: As suggested previously, we don't need to check for LabelSelector length

```suggestion
	if p.daemonSetDeployed {
```

pkg/measurements/deployment_set.go
```diff
@@ -0,0 +1,339 @@
...
+	log.Infof("pprof-collector: DaemonSet %s deployed in namespace %s", types.PprofDaemonSet, types.PprofNamespace)
+	return nil
+}
+
+func (p *pprof) buildDaemonSet() *appsv1.DaemonSet {
+	privileged := true
+	hostNetwork := true
+
+	affinity := &corev1.Affinity{}
+	if len(p.Config.NodeAffinity) > 0 {
```
rsevilla87: Not needed, this condition is checked previously

pkg/measurements/deployment_set.go
```diff
@@ -0,0 +1,339 @@
...
+		return fmt.Errorf("failed to create daemonset: %v", err)
+	}
+
+	log.Infof("pprof-collector: DaemonSet %s deployed in namespace %s", types.PprofDaemonSet, types.PprofNamespace)
+	return nil
+}
+
+func (p *pprof) buildDaemonSet() *appsv1.DaemonSet {
+	privileged := true
+	hostNetwork := true
```
rsevilla87: These 2 variables are only used once

pkg/measurements/pprof_ds.go
```diff
@@ -0,0 +1,339 @@
...
+				corev1.NodeSelectorRequirement{
+					Key:      key,
+					Operator: corev1.NodeSelectorOpIn,
+					Values:   []string{value},
+				},
+			)
+		}
+	}
+	hostPathDirectoryOrCreate := corev1.HostPathDirectoryOrCreate
+	hostPath := p.Config.PProfDirectory
```
rsevilla87: The variable PProfDirectory is used locally where kube-burner is running

rsevilla87: We should always mount node's `/var/run/` in `/mnt` for example

pkg/measurements/deployment_set.go
```diff
@@ -0,0 +1,339 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+func (p *pprof) needsDaemonSet() bool {
+	return len(p.Config.NodeAffinity) > 0
+}
+
+// needsCRIOSocket checks if any pprofTarget uses a unix socket URL (CRI-O)
+func (p *pprof) needsCRIOSocket() bool {
```
rsevilla87: There's no need to use this conditional, let's always mount /var/run in the pod

pkg/measurements/pprof_ds.go
```diff
@@ -0,0 +1,339 @@
...
+			Name: "pprof-data-directory",
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: hostPath,
+					Type: &hostPathDirectoryOrCreate,
+				},
+			},
+		},
+		{
+			Name: "kubelet-ca",
```
rsevilla87: What is this volume for?

pkg/measurements/pprof.go
```diff
@@ -109,10 +133,10 @@ func (p *pprof) getPods(target types.PProftarget) []corev1.Pod {
 
 func (p *pprof) getPProf(wg *sync.WaitGroup, first bool) {
 	var err error
-	var command []string
 	for pos, target := range p.Config.PProfTargets {
 		log.Infof("Collecting %s pprof", target.Name)
-		podList := p.getPods(target)
+		pprofNodeTarget, isNodeTarget := p.getPprofNodeTargets(target)
```
rsevilla87: We don't need to call this function because we can check that with p.needsDaemonSet() in p.getPods(), and obtain the value of `pprofNodeTarget` there as well

pkg/measurements/deployment_set.go
sanjay7178: only privileged variable is needed , because the security context config  of this daemon set pod needed bool var with pointer .
`SecurityContext : &corev1.SecurityContext{Privileged = *privileged}`

pkg/measurements/pprof.go
sanjay7178: I mean I left  the nodeName as if any case pods based pprof tests are happened .

pkg/measurements/pprof_ds.go
sanjay7178: Sure

pkg/measurements/pprof_ds.go
sanjay7178: this requires because the running daemonset pod which was created using service account , the mount gives the access to get the pprof for kubelet heap . this mount would give svc secret's access for curl exec in the daemonset pod

pkg/measurements/deployment_set.go
sanjay7178: here are the changes https://github.com/kube-burner/kube-burner/pull/1045/files#diff-9e5d693c2374e500ac52a572511600ff12b8a452c21c62f46c17df312c466d76R158

sanjay7178:
Hi @rsevilla87  , as you mentioned in few code reviews , to differentiate between node affinity based pods and normal labelled pods , I was using  p.getPprofNodeTargets(target) from each pprofTargets . As per current changes from this commit https://github.com/kube-burner/kube-burner/pull/1045/commits/a48d433d204578320595f08c395f846ec0744c6e , pod exec and pprof retrieval is working fine

pkg/measurements/deployment_set.go
```diff
@@ -0,0 +1,324 @@
...
+	hostPathDirectoryOrCreate := corev1.HostPathDirectoryOrCreate
+	hostPath := p.Config.PProfDirectory
+	if !strings.HasPrefix(hostPath, "/") {
+		hostPath = "/mnt/" + hostPath
+		log.Warnf("PProfDirectory %s is not an absolute path, using %s", p.Config.PProfDirectory, hostPath)
+	}
+
+	// Build volume mounts and volumes based on configuration
+	volumeMounts := p.buildVolumeMounts(hostPath)
+	volumes := p.buildVolumes(hostPath, hostPathDirectoryOrCreate)
```
rsevilla87: hostPath is the host's directory we want to mount in the pod, it should be `/var/run` and only used in buildVolumes()

```suggestion
	volumeMounts := p.buildVolumeMounts()
	volumes := p.buildVolumes(hostPathDirectoryOrCreate)
```

pkg/measurements/pprof_ds.go
rsevilla87: The service account gest automatically mounted in all pods using it, we don't need to manually mount it

pkg/measurements/deployment_set.go
```diff
@@ -0,0 +1,324 @@
...
+		{
+			Name:      "pprof-data-directory",
+			MountPath: hostPath,
+		},
+		{
+			Name:      "kubelet-ca",
+			MountPath: "/var/run/secrets/kubernetes.io/serviceaccount",
+			ReadOnly:  true,
+		},
+	}
```
rsevilla87: Not needed as previously stated, service accounts are automatically mounted in pods

pkg/measurements/pprof_ds.go
```diff
@@ -0,0 +1,324 @@
...
+
+	return ds
+}
+
+// buildVolumeMounts creates volume mounts for pprof data, kubelet CA, and optionally CRI-O socket
+func (p *pprof) buildVolumeMounts(hostPath string) []corev1.VolumeMount {
+	mounts := []corev1.VolumeMount{
+		{
+			Name:      "pprof-data-directory",
+			MountPath: hostPath,
```
rsevilla87:
```suggestion
			MountPath: "/mnt",
```

pkg/measurements/pprof_ds.go
rsevilla87, line 1: Can you rename this file to something more related with the feature? , e.g. pprof_daemonset.go

pkg/measurements/deployment_set.go
rsevilla87: The variable hostPathDirectoryOrCreate is not actually needed, you can replace it for ` ptr.To(corev1.HostPathDirectoryOrCreate)` where ptr is imported from ` k8s.io/utils/ptr`

pkg/measurements/pprof_ds.go
sanjay7178: If I move the file to `measurement/utils` I was getting cyclic dependency issue

sanjay7178: image

rsevilla87 (approved):
lgtm

PR #1048, opened 2025-12-01
Bats test tags logic

rsevilla87:
- CI

## Description

 Add intelligent test filtering for Bats tests based on changed code paths to optimize CI runtime.

This PR introduces a smart test filtering system that automatically runs only relevant Bats tests when specific code areas are modified. The implementation includes:

  - Mapping Configuration (hack/bats_test_mappings.yml): Maps subsystem labels to specific code paths, enabling targeted test execution
  - Label Detection Script (hack/get_changed_labels.py): Python script that analyzes git changes against the main branch and returns appropriate test filter tags
  - Makefile Integration: Automatically applies generated filter tags to Bats test execution
  - Test Tagging: Added subsystem tags to all existing Bats tests for proper categorization

If one of the modified files is not listed in `bats_test_mappings.yml`, the script won't return anything,  meaning that all tests will be executed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ MANIFEST_ARCHS ?= amd64 arm64 ppc64le s390x
 
 # Bats
 JOBS ?= $(shell nproc)
-FILTER_TAGS = $(shell hack/get_changed_labels.py hack/bats_test_mappings.yml)
+FILTER_TAGS ?= $(shell hack/get_changed_labels.py hack/bats_test_mappings.yml)
 
 all: lint build images push
 
--- a/hack/get_changed_labels.py
+++ b/hack/get_changed_labels.py
@@ -93,19 +93,17 @@ def main():
     # Process each label/paths entry
     labels_to_output = []
     for entry in data:
-        if "label" in entry and "paths" in entry:
-            if not isinstance(entry["paths"], list):
-                print(f"Invalid YAML structure, paths should be a list", file=sys.stderr)
-                sys.exit(1)
-            
-            label = entry["label"]
-            paths = entry["paths"]
-
-            if check_paths_changed(paths, changed_files):
-                labels_to_output.append(label)
-        else:
+        if "label" not in entry or "paths" not in entry:
             print(f"Invalid YAML structure, either label or paths are missing", file=sys.stderr)
             sys.exit(1)
+        if not isinstance(entry["paths"], list):
+            print(f"Invalid YAML structure, paths should be a list", file=sys.stderr)
+            sys.exit(1)
+        if check_paths_changed(entry["paths"], changed_files):
+            labels_to_output.append(entry["label"])
+        # If no paths match, exit loop and return empty list to run all tests
+        else:
+            sys.exit(0)
     
     # Output unique labels (one per line)
     if labels_to_output:


Makefile
```diff
@@ -32,6 +32,7 @@ MANIFEST_ARCHS ?= amd64 arm64 ppc64le s390x
 
 # Bats
 JOBS ?= $(shell nproc)
+FILTER_TAGS = $(shell hack/get_changed_labels.py hack/bats_test_mappings.yml)
```
ygalblum: I think we should allow users to set it for local usage. WDYT?

hack/get_changed_labels.py
```diff
@@ -0,0 +1,119 @@
...
+        if "label" in entry and "paths" in entry:
+            if not isinstance(entry["paths"], list):
+                print(f"Invalid YAML structure, paths should be a list", file=sys.stderr)
+                sys.exit(1)
+            
+            label = entry["label"]
+            paths = entry["paths"]
+
+            if check_paths_changed(paths, changed_files):
+                labels_to_output.append(label)
```
ygalblum: According to the description:
> If one of the modified files is not listed in bats_test_mappings.yml, the script won't return anything, meaning that all tests will be executed.

My understanding is that this would translate into an `else` clause here returning without printing anything. How is the requirement expressed in the description come into play here?

Makefile
ygalblum: Should we check if the call succeeded? `FILTER_TAGS` will be assigned the printed string regardless to whether it fails to succeeds

hack/get_changed_labels.py
```diff
@@ -0,0 +1,119 @@
...
+        else:
+            print(f"Invalid YAML structure, either label or paths are missing", file=sys.stderr)
+            sys.exit(1)
+    
+    # Output unique labels (one per line)
+    if labels_to_output:
+        unique_labels = sorted(set(labels_to_output))
```
vishnuchalla, line 110: For these classified labels, do we need a script to have them applied to GH? I guess might be coming up in a follow up PR.

hack/get_changed_labels.py
rsevilla87: If `labels_to_output` is empty (no labels detected), nothing is printed.

ygalblum: Maybe I'm missing something. But, consider the following case:

- Two files were modified
- The first returned `true` on `check_paths_changed`, adding content to `labels_to_output`
- The second returns `false`
- The loop finishes and `labels_to_output` is not empty.
- `unique_labels` is set 
- The filters are printed
- Only parts of the test are executed

Did I miss something here?

rsevilla87: good point, I've made some small changes to address it, now, when check_paths_changed returns False, the scripts exits with rc 0 so all tests are executed.

Makefile
rsevilla87: done

rsevilla87: I haven't found any good way to evaluate the command output in Makefile, do you have any suggestion?

ygalblum: I think that because all error prints go to `stderr`, `FILTER_TAGS` will be empty. If that's the case, then we're good because worst case, no filter is applied

PR #1051, opened 2025-12-01
Reduce GVR logging

afcollins:
## Type of change

- Refactor

## Description

Decreases log level of resource-label delete to Debug, which is quite noisy in a workload like node-density-cni. Instead, this adds a log message for deleting a namespace using GVR.


afcollins:
I am a little confused trying to test this. I observed the logging with v1.18.1, but pushing this change to main might not make sense, as I don't seem to be able to build k-b-ocp since it hasn't been updated to v2 yet.  But it works as expected cherry-picking to v1.18.1 tag.

PR #1052, opened 2025-12-01
Support multiple YAML documents in base object/k8s manifest templates for kube-burner

### LINKED ISSUE #989: [RFE] Support for multiple YAML documents in the same file

It's very common to see files containing multiple YAML documents. Having support for that in kube-burner can reduce and simplify the configuration file.

For example, a single file like, can be used to deploy istio in OpenShift

```yaml
kind: Istio
apiVersion: sailoperator.io/v1
metadata:
  name: default
spec:
  namespace: istio-system
  updateStrategy:
    inactiveRevisionDeletionGracePeriodSeconds: 30
    type: InPlace
  version: v1.26.3
---
kind: IstioCNI
apiVersion: sailoperator.io/v1
metadata:
  name: default
spec:
  namespace: istio-cni
  version: v1.26.3
```

Having something like this

```yaml
objects:
  - objectTemplate: istio-assets.yml
    replicas: 1
```

Rather than:
```yaml
objects:
  - objectTemplate: istio.yml
    replicas: 1
  - objectTemplate: istio-cni.yml
    replicas: 1
```

sanjay7178:
…rk policy handling

## Type of change

- New feature

## Description

- Added `yamlToUnstructuredMultiple()` functions for decoding multi document yaml manifest and returns slice of unstructured object and group version kind . Which uses `k8s.io/apimachinery/pkg/util/yaml` to decode yaml . 
- Refactored `getObjectTypes()` func in `pkg/measurements/netpol_latency.go` for support multi yaml document 
- Added `hasNetworkPolicy()`  helper func to checking network policy is present in a objects
- Renamed and refactored `getNetworkPolicy()` with `getNetworkPolicies()` to support multi yaml manifest which returns slice of policies 
- Refactored `prepareConnections()` to support slice of network policies to prepare multiple connections 
- Added `extractImagesFromObject()` as a helper func for `getJobImages()`
- Added 	`documentIndex` in object struct to support for other functions 
- Refacotred `setupCreateJob()`  to support multiple gvk's  

Usage : 
```bash
kube-burner init -c ./kubelet-density.yml --log-level=debug 
```
get istio virtual service and gateway : 

```bash
kubectl get gateway -n kubelet-density && echo "---" && kubectl get virtualservice -n kubelet-density
```

kubelet-density.yml
```yaml
---
global:
  gc: false
  measurements:
    - name: podLatency
jobs:
  - name: kubelet-density
    jobIterations: 2
    qps: 2
    burst: 2
    preLoadImages: false
    namespacedIterations: false
    namespace: kubelet-density
    waitWhenFinished: true
    podWait: false
    objects:
      - objectTemplate: ./istio-combined.yml
        replicas: 1
```

istio-combined.yaml : 
```yaml
apiVersion: networking.istio.io/v1
kind: Gateway
metadata:
  name: kubelet-density-gateway-{{.Iteration}}
spec:
  selector:
    istio: ingressgateway
  servers:
    - port:
        number: 80
        name: http
        protocol: HTTP
      hosts:
        - "*"
---
apiVersion: networking.istio.io/v1
kind: VirtualService
metadata:
  name: kubelet-density-vs-{{.Iteration}}
spec:
  hosts:
    - "*"
  gateways:
    - kubelet-density-gateway-{{.Iteration}}
  http:
    - route:
        - destination:
            host: kubelet-density-svc
            port:
              number: 80
```

- Closes #989

### DIFF
--- a/test/test-k8s.bats
+++ b/test/test-k8s.bats
@@ -260,6 +260,6 @@ teardown_file() {
   # 6 frontend + 6 backend
   verify_object_count pod 6 "" app=frontend,kube-burner.io/uuid=${UUID}
   verify_object_count pod 6 "" app=backend,kube-burner.io/uuid=${UUID}
-  ${KUBE_BURNER} destroy --uuid ${UUID}
+  ${KUBE_BURNER} destroy -c kube-burner-multi-doc.yml
   verify_object_count namespace 0 "" kube-burner.io/uuid=${UUID}
 }


pkg/burner/create.go
```diff
@@ -68,32 +68,39 @@ func (ex *JobExecutor) setupCreateJob() {
...
+				namespace:     unsList[i].GetNamespace(),
+				gvk:           gvk,
+				documentIndex: i,
+			}
+			if err == nil {
+				obj.gvr = mapping.Resource
+				obj.namespaced = mapping.Scope.Name() == meta.RESTScopeNameNamespace
+			} else {
+				obj.gvr = schema.GroupVersionResource{}
+				obj.gvk = gvk
```
ygalblum: You're already setting `gvk` in the instance creation

pkg/burner/create.go
```diff
@@ -224,29 +231,42 @@ func (ex *JobExecutor) replicaHandler(ctx context.Context, labels map[string]str
...
-			renderedObj := ex.renderTemplateForObject(obj, iteration, r, false)
-			// Re-decode rendered object
-			yamlToUnstructured(obj.ObjectTemplate, renderedObj, newObject)
+			newObjects, gvks := ex.renderTemplateForObjectMultiple(obj, iteration, r)
 
-			maps.Copy(copiedLabels, newObject.GetLabels())
-			newObject.SetLabels(copiedLabels)
-			updateChildLabels(newObject, copiedLabels)
+			// Only process the document that corresponds to this object's documentIndex
+			if obj.documentIndex >= len(newObjects) {
```
ygalblum: Why would this happen? I'm not saying that you shouldn't check, but this seems like something that went really wrong with the flow. Keep in mind that as far as the test flow goes, this is a silent failure (while you print an error, the error is not returned to the caller)

pkg/burner/create.go
```diff
@@ -224,29 +231,42 @@ func (ex *JobExecutor) replicaHandler(ctx context.Context, labels map[string]str
...
+			}
 			// replicaWg is necessary because we want to wait for all replicas
 			// to be created before running any other action such as verify objects,
 			// wait for ready, etc. Without this wait group, running for example,
 			// verify objects can lead into a race condition when some objects
 			// hasn't been created yet
 			replicaWg.Add(1)
-			go func(n string) {
-				if !obj.namespaced {
+			go func(gvr schema.GroupVersionResource, newObj *unstructured.Unstructured, n string, namespaced bool) {
```
ygalblum: Why do you need to change the signature? The advantage of using inline functions is that you don't need to add parameters. The reason `namespace` is a parameter is because it can be changed within the function

pkg/burner/pre_load.go
```diff
@@ -94,47 +94,57 @@ func preLoadImages(job JobExecutor, clientSet kubernetes.Interface) error {
...
 		if err != nil {
 			return imageList, err
 		}
-		yamlToUnstructured(object.ObjectTemplate, renderedObj, &unstructuredObject)
-		switch unstructuredObject.GetKind() {
-		case Deployment, DaemonSet, ReplicaSet, Job, StatefulSet:
-			var pod NestedPod
-			runtime.DefaultUnstructuredConverter.FromUnstructured(unstructuredObject.UnstructuredContent(), &pod)
-			for _, i := range pod.Spec.Template.Containers {
+		// yamlToUnstructured(object.ObjectTemplate, renderedObj, &unstructuredObject)
```
ygalblum: remove commented line

pkg/burner/utils.go
```diff
@@ -106,12 +110,40 @@ func updateChildLabels(obj *unstructured.Unstructured, labels map[string]string)
 
 func yamlToUnstructured(fileName string, y []byte, uns *unstructured.Unstructured) (runtime.Object, *schema.GroupVersionKind) {
 	o, gvk, err := scheme.Codecs.UniversalDeserializer().Decode(y, nil, uns)
+	log.Debugf("Decoded object GVK: %v", gvk)
```
ygalblum: This method is called many times. Do we really need this print?

pkg/burner/utils.go
```diff
@@ -106,12 +110,40 @@ func updateChildLabels(obj *unstructured.Unstructured, labels map[string]string)
...
+			if err != io.EOF {
+				break
+			}
+			break
+		}
+		if len(uns.Object) == 0 {
+			break
+		}
+		gvk := uns.GroupVersionKind()
+		log.Debugf("Decoded object GVK: %v", gvk)
```
ygalblum: Same for the print

pkg/burner/utils.go
```diff
@@ -106,12 +110,40 @@ func updateChildLabels(obj *unstructured.Unstructured, labels map[string]string)
...
 
+func yamlToUnstructuredMultiple(fileName string, y []byte) ([]*unstructured.Unstructured, []*schema.GroupVersionKind) {
+	decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(y), 4096)
+	var gvks []*schema.GroupVersionKind
+	var objects []*unstructured.Unstructured
+	for {
+		uns := &unstructured.Unstructured{}
+		err := decoder.Decode(uns)
+		if err != nil {
+			if err != io.EOF {
```
ygalblum: You're breaking regardless to the type of the error

pkg/measurements/netpol_latency.go
```diff
@@ -125,13 +124,13 @@ func (nplmf netpolLatencyMeasurementFactory) NewMeasurement(jobConfig *config.Jo
...
 }
 
-func yamlToUnstructured(fileName string, y []byte, uns *unstructured.Unstructured) (runtime.Object, *schema.GroupVersionKind) {
-	o, gvk, err := scheme.Codecs.UniversalDeserializer().Decode(y, nil, uns)
-	if err != nil {
-		log.Fatalf("Error decoding YAML (%s): %s", fileName, err)
-	}
-	return o, gvk
-}
+// func yamlToUnstructured(fileName string, y []byte, uns *unstructured.Unstructured) (runtime.Object, *schema.GroupVersionKind) {
```
ygalblum: Remove commented code

pkg/measurements/netpol_latency.go
```diff
@@ -221,16 +220,37 @@ func (n *netpolLatency) getNetworkPolicy(iteration int, replica int, obj config.
...
+	decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(renderedObj), 4096)
+
+	for {
+		var rawObj map[string]interface{}
+		err := decoder.Decode(&rawObj)
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			log.Warnf("Error decoding YAML document: %v", err)
```
ygalblum: Why is failure to marshal the object or to decode the policy considered as fatal while failure to decode the YAML file is just a warning and failure to cast doesn't even get a print

pkg/measurements/netpol_latency.go
```diff
@@ -245,77 +265,94 @@ func (n *netpolLatency) prepareConnections() {
...
-						if from.NamespaceSelector == nil && from.PodSelector == nil {
-							continue
-						}
-						for _, port := range ingress.Ports {
-							if *port.Protocol != corev1.ProtocolTCP {
-								log.Fatalf("Only TCP ports supported in testing")
+
+				for _, networkPolicy := range n.getNetworkPolicies(i, r, obj, cleanTemplate) {
+
+					// networkPolicy := n.getNetworkPolicy(i, r, obj, cleanTemplate)
```
ygalblum: remove commented code

pkg/burner/create.go
sanjay7178: that's the edge case I found out to have . I mean replicaHandler() doesn't have return types  . Do  I need to consider log.Fatalf() instead of log.Errorf() ?

pkg/burner/create.go
sanjay7178: Sorry I didn't looked for the goroutine's signatures to have multiple .

pkg/burner/utils.go
sanjay7178: Nope

sanjay7178: I made it for debugging purpose in dev perspective

pkg/measurements/netpol_latency.go
sanjay7178: Sorry , I couldn't maintain the flow . It should be a fatal .

pkg/burner/create.go
sanjay7178: Sorry I'm unsure about this , I'm following the previous approach https://github.com/kube-burner/kube-burner/pull/1052/files#diff-76e5989cd670c9a6f994b653f5f41d8eacf7d309dc9074b48f9754bbac133018L88 , as iterating the list of gvks

pkg/burner/create.go
ygalblum: > that's the edge case I found out to have

Do you know why this happens? Can there be an error you're missing in a preceding step?

pkg/burner/create.go
ygalblum: In the original code `gvk` was not assigned during the instantiation of `obj`. It was assigned only here. But, now it does. So, you're setting it twice

pkg/burner/create.go
```diff
@@ -224,28 +231,41 @@ func (ex *JobExecutor) replicaHandler(ctx context.Context, labels map[string]str
...
+			newObject := newObjects[obj.documentIndex]
+			gvk := gvks[obj.documentIndex]
 
+			objectLabels := make(map[string]string)
+			maps.Copy(objectLabels, copiedLabels)
+			maps.Copy(objectLabels, newObject.GetLabels())
+			newObject.SetLabels(objectLabels)
+			updateChildLabels(newObject, objectLabels)
+
+			_, err := ex.mapper.RESTMapping(gvk.GroupKind())
```
ygalblum, line 232: Why is this needed?

pkg/burner/pre_load.go
```diff
@@ -94,47 +94,56 @@ func preLoadImages(job JobExecutor, clientSet kubernetes.Interface) error {
...
-		case Deployment, DaemonSet, ReplicaSet, Job, StatefulSet:
-			var pod NestedPod
-			runtime.DefaultUnstructuredConverter.FromUnstructured(unstructuredObject.UnstructuredContent(), &pod)
-			for _, i := range pod.Spec.Template.Containers {
+		unsList, _ := yamlToUnstructuredMultiple(object.ObjectTemplate, renderedObj)
+		for _, uns := range unsList {
+			images := extractImagesFromObject(uns, renderedObj)
+			imageList = append(imageList, images...)
+		}
+
```
ygalblum: Remove redundant line

pkg/burner/utils.go
ygalblum: We're trying to reduce logging load even in debug. See: https://github.com/kube-burner/kube-burner/pull/1053/files. Is this print really needed?

pkg/burner/utils.go
```diff
@@ -106,12 +110,36 @@ func updateChildLabels(obj *unstructured.Unstructured, labels map[string]string)
...
 }
 
+func yamlToUnstructuredMultiple(fileName string, y []byte) ([]*unstructured.Unstructured, []*schema.GroupVersionKind) {
+	decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(y), 4096)
+	var gvks []*schema.GroupVersionKind
+	var objects []*unstructured.Unstructured
+	for {
+		uns := &unstructured.Unstructured{}
+		err := decoder.Decode(uns)
+		if err != nil || err == io.EOF {
```
ygalblum: You'll never get to the `err == io.EOF` part

pkg/burner/create.go
sanjay7178: Sorry I got it , I was confused , the gvk is actually setting in objects map , previously in `yamlToUnstructuredMultiple()`

pkg/burner/create.go
sanjay7178: Before attempting to create an object, this error confirms the REST mapping exists. If the mapping fails, the function returns early, preventing a futile create attempt. The object's GVK might not have been resolvable during setupCreateJob() perhaps corresponding CRD might not installed or kube-api server isn't reachable at the moment .

pkg/burner/utils.go
sanjay7178: No needed

pkg/burner/utils.go
sanjay7178: Sure, I'm removing

pkg/burner/create.go
sanjay7178: Nope I didn't get any error while testing , I found out to have that in the flow . I guess the `if` loop might not catch most of the time

sanjay7178: So I'm removing it

pkg/burner/create.go
```diff
@@ -68,32 +68,38 @@ func (ex *JobExecutor) setupCreateJob() {
...
+				gvk:           gvk,
+				documentIndex: i,
+			}
+			if err == nil {
+				obj.gvr = mapping.Resource
+				obj.namespaced = mapping.Scope.Name() == meta.RESTScopeNameNamespace
+			} else {
+				obj.gvr = schema.GroupVersionResource{}
+			}
+			obj.Kind = gvk.Kind
```
ygalblum: sorry for not seeing this earlier and I see that this line was already there. But, can't setting the `Kind` be part of the `obj` instanciation?

pkg/burner/create.go
ygalblum: Please add a comment about it. The comment is needed because the error states that there was an error getting the mapping, but a mapping is not stored

pkg/burner/create.go
sanjay7178: object struct{} embeds the `config.Object`  , so the kind field can be instantiated &object{} var init .

pkg/burner/create.go
sanjay7178: changes at https://github.com/kube-burner/kube-burner/pull/1052/commits/04f6d0c25da60f215f8b15d069e21ec7ddaaae63

pkg/burner/create.go
sanjay7178: changes at https://github.com/kube-burner/kube-burner/pull/1052/commits/04f6d0c25da60f215f8b15d069e21ec7ddaaae63

ygalblum (approved):
LGTM

test/test-k8s.bats
```diff
@@ -253,3 +253,14 @@ teardown_file() {
...
 }
+
+@test "kube-burner-multi-doc.yml: multi-document YAML templates" {
+  export GC=false
+  run_cmd ${KUBE_BURNER} init -c kube-burner-multi-doc.yml --uuid=${UUID} --log-level=debug
+  # 3 iterations * 2 replicas * 2 pods per template = 12 pods total
+  # 6 frontend + 6 backend
+  verify_object_count pod 6 "" app=frontend,kube-burner.io/uuid=${UUID}
+  verify_object_count pod 6 "" app=backend,kube-burner.io/uuid=${UUID}
+  ${KUBE_BURNER} destroy --uuid ${UUID}
```
ygalblum: Please rebase on top of main. This will require changing the call to `destroy`

ygalblum (commented):
Thanks for adding the docs and test. Please rebase on top of main and address the change in the `destroy` command. See https://github.com/kube-burner/kube-burner/pull/1057 for details

sanjay7178:
@ygalblum I seen the PR you've mentioned what's between using `${KUBE_BURNER} destroy -c kube-burner-multi-doc.yml` and `${KUBE_BURNER} destroy --uuid ${UUID}`  . I mean what can't be used uuid ?

ygalblum:
> I mean what can't be used uuid ?

Yes. `uuid` is no longer a supported argument. Users need to pass the orignal configuration file and environment variables used to run `init`

ygalblum (approved):
LGTM

PR #1053, opened 2025-12-01
Print delete log messages only when something is going to be deleted

rsevilla87:
## Type of change

- Bug fix
- Optimization
 
## Description

Some cleanup messages are printed unconditionally, we  should print them when something is going to be deleted to avoid confusion.

- Related Issue https://github.com/kube-burner/kube-burner-ocp/issues/341
- Closes #

### DIFF
--- a/cmd/kube-burner/kube-burner.go
+++ b/cmd/kube-burner/kube-burner.go
@@ -83,6 +83,7 @@ func initCmd() *cobra.Command {
 	var userDataFile string
 	var allowMissingKeys bool
 	var rc int
+	var setValues []string
 	cmd := &cobra.Command{
 		Use:   "init",
 		Short: "Launch benchmark",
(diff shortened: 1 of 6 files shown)


pkg/util/namespaces.go
```diff
@@ -126,13 +126,11 @@ func CleanupNonNamespacedResources(ctx context.Context, clientSet kubernetes.Int
...
 func DeleteNonNamespacedResources(ctx context.Context, resources *unstructured.UnstructuredList, resourceInterface dynamic.NamespaceableResourceInterface) {
-	if len(resources.Items) > 0 {
-		for _, item := range resources.Items {
-			log.Debugf("Deleting non-namespaced resource: %s", item.GetName())
-			err := resourceInterface.Delete(ctx, item.GetName(), metav1.DeleteOptions{})
-			if err != nil {
-				log.Errorf("Error deleting %v/%v: %v", item.GetKind(), item.GetName(), err)
-			}
+	for _, item := range resources.Items {
+		log.Debugf("Deleting non-namespaced resource: %s", item.GetName())
```
ygalblum: In the other file, we're printing once, but here we print for each item. Do we want that?

rsevilla87: That's right, printing for each item can be overkill, deleting this entry

ygalblum: Don't we want a general one (like you added in the other cases)?

rsevilla87: Fixed now, thanks for the suggestions

PR #1054, opened 2025-12-02
feature: jobs.preLoadImage added for custom preLoadImage

### LINKED ISSUE #1036: [RFE] Customisable PreLoad Image
Currently the preload image is hardcoded to "registry.k8s.io/pause:3.1". Since we have a disconnected openshift cluster and we are not able to access these images at run time. We would like to be able to customize this field so that the right image may be pulled from our internal registry.
https://github.com/kube-burner/kube-burner/blob/c74094300d4c30eed237929b4e0a9c0f54d587f0/pkg/burner/pre_load.go#L174

May be a preLoadImage field in the Job config so that the image can be customised as needed.

1. Rebuild the sources with expected image.
    This may be easiest solution for us.
2. Set up some kind of mirror in our cluster so that "registry.k8s.io" transparently fetches from our registry instead.
    This is a much more involved solution for us, I think.
3. Create the preLoad DaemonSet manually as part of our tests, and set preLoadImages to false in our kube burner job configurations.

**Additional context**

Mujib-Ahasan:
## Type of change

- New feature

## Description
A `preLoadImage` field in the job config is added so that the image can be customized as needed 

- Closes #1036

### DIFF
--- a/pkg/burner/pre_load.go
+++ b/pkg/burner/pre_load.go
@@ -79,7 +79,7 @@ func preLoadImages(job JobExecutor, clientSet kubernetes.Interface) error {
 		log.Infof("No images found to pre-load, continuing")
 		return nil
 	}
-	err = createDSs(clientSet, imageList, job.NamespaceLabels, job.NamespaceAnnotations, job.PreLoadNodeLabels)
+	err = createDSs(clientSet, imageList, job.NamespaceLabels, job.NamespaceAnnotations, job.PreLoadNodeLabels, job.PreLoadImage)
 	if err != nil {
 		return fmt.Errorf("pre-load: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


ygalblum:
The change on its own seems OK. But, I think the name is misleading. One can interpret it as an image to preload while in fact it is the image used to run the DS that triggers the preloading of images.

rsevilla87:
We shouldn't merge this PR, context https://github.com/kube-burner/kube-burner/issues/1036#issuecomment-3611518390

vishnuchalla:
Closing this, as it has been stale for a while. Also based on the conversation here, this change isn't needed for now: https://github.com/kube-burner/kube-burner/issues/1036

PR #1055, opened 2025-12-11
Object deletion fixes

rsevilla87:
## Type of change

- Bug fix
- Optimization

## Description

Not critical fixes, since they just help to avoid printing ugly error messages due to conflicts or missing objects.

- Added a new property nsChurning to the jobExecutor struct, to prevent kube-burner to try re-creating non namespaced resources with when "namespaces churning" is enabled.
-  obj.namespaced conditional added to the CleanupNamespacesUsingGVR() method to prevent kube-burner to try deleting non namespaced objects.
- Renamed some methods to better understanding their aim

- Related Issue #
- Closes #

### DIFF
--- a/pkg/util/namespaces.go
+++ b/pkg/util/namespaces.go
@@ -118,7 +118,7 @@ func CleanupNonNamespacedResourcesByLabel(ctx context.Context, clientSet kuberne
 					log.Debugf("Unable to list resource %s: %v", resource.Name, err)
 					continue
 				}
-				if len(resources.Items) > 0 {
+				if len(resources.Items) == 0 {
 					log.Infof("Deleting %d %ss labeled with %s", len(resources.Items), resource.Kind, labelSelector)
 					DeleteNonNamespacedResources(ctx, resources, resourceInterface)
 				}


pkg/burner/pre_load.go
```diff
@@ -88,7 +88,7 @@ func preLoadImages(job JobExecutor, clientSet kubernetes.Interface) error {
 	// 5 minutes should be more than enough to cleanup this namespace
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
 	defer cancel()
-	util.CleanupNamespaces(ctx, clientSet, "kube-burner-preload=true")
+	util.CleanupNamespacesByLabel(ctx, clientSet, "kube-burner-preload=true")
```
ygalblum, line 91: I know it's not related to this change. But, wasn't this label aligned with the `kube-burner.io` labels change?

pkg/util/namespaces.go
```diff
@@ -118,7 +118,7 @@ func CleanupNonNamespacedResources(ctx context.Context, clientSet kubernetes.Int
 					log.Debugf("Unable to list resource %s: %v", resource.Name, err)
 					continue
 				}
-				if len(resources.Items) == 0 {
+				if len(resources.Items) > 0 {
```
ygalblum: According to the PR description this is a minor change mostly to avoid false negative prints. But, this change seems more substantial. Or am I misunderstanding it?

rsevilla87: That's right, this is a bug fix, not mentioned in the PR description, I think it deserves its own PR. let me file it.

pkg/burner/pre_load.go
rsevilla87: I probably forgot to update it, I'll try to update it in another PR

ygalblum (approved):
LGTM

PR #1056, opened 2025-12-11
feat: namespaceDelay

luckylittle:
## Type of change

- New feature - the `namespaceDelay` flag to kube-burner.

### Add:

- `NamespaceDelay time.Duration` to the `Job` struct with YAML and JSON tags
- Implemented the delay logic (`pkg/burner/create.go`)

### Changed:

- Updated the documentation** (`docs/reference/configuration.md`)
   - Added `namespaceDelay` to the Jobs section table
   - Placed it right after `iterationsPerNamespace` since they're related
   - Included a description explaining it helps reduce OVS saturation

The implementation:
- Applies the delay only when transitioning to a new namespace (not on the first namespace)
- Works with any `iterationsPerNamespace` value
- Logs when the delay is applied
- Respects the context cancellation

The delay is applied before creating a new namespace when `iterationsPerNamespace` causes a namespace transition, helping reduce OVS saturation when many pods are created at once.

- https://issues.redhat.com/browse/OCPBUGS-65910
- https://github.com/kube-burner/kube-burner-ocp/pull/361


ygalblum:
I'm missing the "why is this needed?". The user knows if the object they are creating requires a namespace and whether or not `namespacedIterations` is set. So, why can't they use `jobIterationDelay`. Can you please elaborate on how is using `namespaceDelay` different?

luckylittle:
> I'm missing the "why is this needed?". The user knows if the object they are creating requires a namespace and whether or not `namespacedIterations` is set. So, why can't they use `jobIterationDelay`. Can you please elaborate on how is using `namespaceDelay` different?

`jobIterationDelay` is a delay between each pod creations. Even with `1s` delay, this makes the test running for hours on the testing cluster. Adding delay between creation of large (~100) pods avoids hitting the OVS saturation bugs:
* OCPBUGS-65910 ovs performance degradation when deleting and creating large amount of Pods.
* FDP-1778 Improve robustness of OVS against RTNL contention
* RHEL-77265 Improve RTNL locking in mlx5e driver
* RHEL-76544 OVS errors and potential resource limitations during SR-IOV VF
* RHEL-68117 CNB96: arp: Random clean up and RCU conversion for ioctl(SIOCGARP)
* RHELMISC-22716 Reducing the scope of the RTNL to reduce contention
* PRFRSR-244 Monitor OVS control path performance with SR-IOV VF interfaces

QPS/Burst prevent API overload, while `NamespaceDelay` will help manage infrastructure/OVS load when many pods are created/deleted simultaneously. We have tested the approach of adding 30s between each 100 pods and it works nicely.

pkg/burner/create.go
```diff
@@ -126,6 +127,13 @@ func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterati
...
 		}
 		if ex.nsRequired && ex.NamespacedIterations {
+			currentNsIndex := i / ex.IterationsPerNamespace
+			// Apply delay when moving to a new namespace (and it's not the first namespace)
+			if previousNsIndex >= 0 && currentNsIndex != previousNsIndex && ex.NamespaceDelay > 0 {
+				log.Infof("Namespace delay: sleeping for %v before creating new namespace", ex.NamespaceDelay)
+				time.Sleep(ex.NamespaceDelay)
+			}
+			previousNsIndex = currentNsIndex
 			ns = ex.createNamespace(ex.generateNamespace(i), nsLabels, nsAnnotations)
```
ygalblum: Instead of calculating an index, you can just use the generate namespace name:
```suggestion
			currentNsName := ex.generateNamespace(i)
			// Apply delay when moving to a new namespace (and it's not the first namespace)
			if currentNsName != previousNsName {
			    if ex.NamespaceDelay > 0 {
				    log.Infof("Namespace delay: sleeping for %v before creating new namespace", ex.NamespaceDelay)
				    time.Sleep(ex.NamespaceDelay)
				}
			    previousNsIndex = currentNsIndex
                 ns = ex.createNamespace(currentNsName, nsLabels, nsAnnotations)
             }
```

You'd also need to change the instantiation of `previousNsIndex` to:
```
previousNsName := ""
```

ygalblum:
I missed the fact that iterations may share a namespace when `iterationsPerNamespace` is set. This is the missing part. Otherwise, each iteration has it's own namespace and the delay would be the same.

I've added a comment about the code (removing the need for another index calculate)

Having said that, I am still now convinced this is the correct way to address your issue. My reason is that now it seems that you are using the namespaces to mark the need for a delay. Is that the case? Do you really need the separation into namespaces as you define it, or are you only using it as a marker? Maybe a better solution is some sort of a marker for the delay. That is, delay for X seconds every Y iterations regardless to namespaces.

luckylittle:
> I missed the fact that iterations may share a namespace when `iterationsPerNamespace` is set. This is the missing part. Otherwise, each iteration has it's own namespace and the delay would be the same.
> 
> I've added a comment about the code (removing the need for another index calculate)
> 
Thanks, I have commited your suggestion.

> Having said that, I am still now convinced this is the correct way to address your issue. My reason is that now it seems that you are using the namespaces to mark the need for a delay. Is that the case? Do you really need the separation into namespaces as you define it, or are you only using it as a marker? Maybe a better solution is some sort of a marker for the delay. That is, delay for X seconds every Y iterations regardless to namespaces.

Yes, it seemed like a good place to introduce the delay and provide gradual increase, withouth killing the cluster due to OVS saturation bugs. Indeed, adding delay for X seconds every Y iterations regardless of namespaces would be an alternative solution! I am happy to discuss internally etc.
I just want to re-iterate that we will see more of these issues coming up as the hardware is becoming cheaper. This issue is easily reproducable with:
* Large amount of cores (e.g. 512 core AMD nodes) - the size of the nodes is directly related to the size of the problem.
* Intel E810 and 100GBit Mellanox NICs (possibly others might be affected too).
* OCP 4.19
* No SRIOV just veths.
The issues happens when we delete ~700 or more PODs using kube-burner and/or creating ~700 pods in one go.

ygalblum:
> I just want to re-iterate that we will see more of these issues coming up as the hardware is becoming cheaper. This issue is easily reproducable with

I am not saying this is not an issue. I'm still not convinced that this is the correct solution. 

I'm still asking: "Do you really need the separation into namespaces as you define it, or are you only using it as a marker?" Moreover, think about future scenarios. This change compels them to split the resources into namespaces in order to add the needed delay. What if namespacing breaks the use case?

PR #1057, opened 2025-12-11
Refactor destroy cmd

rsevilla87:
## Type of change

- Refactor
- Optimization

## Description

Refactor of the destroy subcommand to take advantage of the available garbage collection methods rather than deleting by label

- Related Issue #
- Closes #

### DIFF
--- a/cmd/kube-burner/kube-burner.go
+++ b/cmd/kube-burner/kube-burner.go
@@ -200,6 +200,7 @@ func destroyCmd() *cobra.Command {
 	var timeout time.Duration
 	var kubeConfig, kubeContext, configFile string
 	var skipLogFile bool
+	var userDataFile string
 	uuid := uid.NewString()
 	cmd := &cobra.Command{
 		Use:   "destroy",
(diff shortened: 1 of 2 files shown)


cmd/kube-burner/kube-burner.go
```diff
@@ -200,37 +199,51 @@ func healthCheck() *cobra.Command {
...
 	var timeout time.Duration
-	var kubeConfig, kubeContext string
+	var kubeConfig, kubeContext, configFile string
 	var skipLogFile bool
 	var rc int
 	cmd := &cobra.Command{
 		Use:   "destroy",
-		Short: "Destroy old namespaces labeled with the given UUID.",
+		Short: "Destroy benchmark assets",
 		PostRun: func(cmd *cobra.Command, args []string) {
```
ygalblum: Why is this needed? On errors the code either runs `log.Fatal` which calls `os.Exit`, or called the latter directly. Not sure we ever reach here with `rc` that is not `0`

cmd/kube-burner/kube-burner.go
```diff
@@ -200,37 +199,51 @@ func healthCheck() *cobra.Command {
...
-			log.Info("👋 Exiting kube-burner ", uuid)
+			log.Info("👋 Exiting kube-burner")
 			os.Exit(rc)
 		},
 		Args: cobra.NoArgs,
+		PreRun: func(cmd *cobra.Command, args []string) {
+			if uuid == "" {
```
ygalblum: IIUC the `uuid` was removed from the parameters list, so it's always empty, no? Also, does assigning it has to be in the `PreRun`? Can't it just go into the `Run`

ygalblum:
I think there might be an issue with this approach, but I could be wrong. What if the config file has a dependency on environment variables? Calling destroy not only requires the user to set the variables, but also know the values used in the test. No?

cmd/kube-burner/kube-burner.go
rsevilla87: correct, deleting it

cmd/kube-burner/kube-burner.go
rsevilla87: good point, removing this useless conditional

rsevilla87:
> I think there might be an issue with this approach, but I could be wrong. What if the config file has a dependency on environment variables? Calling destroy not only requires the user to set the variables, but also know the values used in the test. No?

Yeah, that's right, it's up to the user to use the very same configuration file and required env vars. Although the current approach is not perfect it's better than the previous one.

ygalblum:
> Although the current approach is not perfect it's better than the previous one

OK I guess

> Yeah, that's right, it's up to the user to use the very same configuration file and required env vars

The documentation should be updated accordingly

PR #1061, opened 2025-12-16
Fix namespaces churning logic

rsevilla87:
## Type of change

- Bug fix

## Description

The logic of namespaces churning is broken. The order of namespaces returned from List() is not guaranteed, so it's possible for kube-burner to mistakenly try to re-create a namespace that has not been deleted.

Rolling back to the previous logic that dynamically generates the namespace name to churn.

It's possible that the issue wasn't detected by CI b/c the number namespaces to churn was very low (20% of 5 iterations)

- Related Issue #
- Closes #

### DIFF
--- a/pkg/burner/create.go
+++ b/pkg/burner/create.go
@@ -414,7 +414,7 @@ func (ex *JobExecutor) churnNamespaces(ctx context.Context) []error {
 			numJ, _ := strconv.Atoi(nsList[j].Name[strings.LastIndex(nsList[j].Name, "-")+1:])
 			return numI < numJ
 		})
-		// delete numToChurn namespaces starting at randStart
+		// Add label to namespaces to use label selector for deletion
 		for _, ns := range nsList[randStart : numToChurn+randStart] {
 			_, err = ex.clientSet.CoreV1().Namespaces().
 				Patch(ctx, ns.Name, types.StrategicMergePatchType, []byte(delPatch), metav1.PatchOptions{})


pkg/burner/create.go
```diff
@@ -400,20 +391,21 @@ func (ex *JobExecutor) churnNamespaces(ctx context.Context) []error {
 		}
 
 		// Max amount of churn is 100% of namespaces
-		var randStart int
-		if len(jobNamespaces.Items)-numToChurn+1 > 0 {
-			randStart = rand.Intn(len(jobNamespaces.Items) - numToChurn + 1)
+		if ex.JobIterations-numToChurn+1 > 0 {
+			randStart = rand.Intn(ex.JobIterations - numToChurn + 1)
 		}
 		// delete numToChurn namespaces starting at randStart
```
ygalblum: I know it was there already. But, please change the comment to something like `mark the namespaces for deletion`. It took me time to understand that this loop only marks them with a label and the deletion happens only later

rsevilla87:
> I don't know the history of the changes here. But, you wrote that part of the issue was that the order of the list is not guarantied. Would sorting them alphabetically have solved the issue?

Sorting them didn't work because alphabetical is not 

> I don't know the history of the changes here. But, you wrote that part of the issue was that the order of the list is not guarantied. Would sorting them alphabetically have solved the issue?

fixed, initially tried to perform an alphabetical sorting, but realized that I needed a natural sorting, so for example, namespace-10 is placed after namespace-9

ygalblum (approved):
LGTM

PR #1068, opened 2025-12-28
Add hook support for job execution stages and background processing

### LINKED ISSUE #1041: [RFE]: Fork a process from kube-burner targetting combined workload execution
Kube-burner [rds-core workload](https://github.com/kube-burner/kube-burner-ocp?tab=readme-ov-file#core-rds-workloads) is executed with data-plane workload running in parallel While kube-burner-ocp rds-core churns namespaces, we try to observe if the data-plane and metrics collected by kube-burner-ocp are -severely- affected by these operations.

The control-plane, kube-burner-ocp rds-core workload, takes X amount of time to be deployed and in "running state". The X amount of time highly depends on the cluster size (number of worker nodes). For instance, in my typical bare-metal environments with +100 workers, it takes up to +7min to have all pods deployed and verified (in running state), before churn can start. (Note that, for rds-core, `--iterations` == # worker nodes, and each iteration corresponds to 1 namespace with 52 pods. In total, kube-burner-ocp is deploying +5000 pods before it starts with churning).

The data-plane, crucible workload, takes Y amount of time to be deployed and in "running state". The Y amount of time highly depends on the number of pod pairs deployed, e.g., typically 2 or 8 pairs (client and server) for uni-/bidirectional traffic and, on the cluster size (number of worker nodes). For instance, for the same environments mentioned above, crucible took +8min to have all pods deployed and "started", i.e. sending traffic.

If I have to measure the impact of control-plane (kube-burner-ocp rds churning operations) on the data-plane (crucible client-server pods), ideally, the "running state" of both tools should be synched in time, i.e., not overlapping -too much- with their deployment phase, e.g., creating pods, downloading or checking images, etc.

Let kube-burner-ocp fork a crucible process, and let a control/supervisor process(?) suspend crucible's execution via e.g. sigpause(?), before it enters its "running state" (in the example below, stop by `client-start-begin`), and let crucible continue only when kube-burner-ocp reaches the point of start with churning (in the example below, by `Verifying created objects`). To my knowledge there are some sync primitives that can be explored, but it would have to be added in both tools or another option would work via a parent/control process. 

**Additional context**
Below, I post the runtime output of both tools, which I used to synch them up, with help of a separate script/process:

Crucible:
```
echo 'Roadblock: Mon Jul 14 07:55:37 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:endpoint-deploy-begin'
#Roadblock: Mon Jul 14 07:58:25 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:endpoint-deploy-end
#Roadblock: Mon Jul 14 07:58:28 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:engine-init-begin
#Roadblock: Mon Jul 14 07:58:32 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:engine-init-end
#Roadblock: Mon Jul 14 07:58:37 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:get-data-begin
#Roadblock: Mon Jul 14 07:58:42 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:get-data-end
#Roadblock: Mon Jul 14 07:58:53 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:collect-sysinfo-begin
echo 'Roadblock: Mon Jul 14 07:58:57 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:collect-sysinfo-end'
#Roadblock: Mon Jul 14 07:59:01 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:start-tools-begin
#Roadblock: Mon Jul 14 07:59:05 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:start-tools-end
#Roadblock: Mon Jul 14 07:59:21 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:setup-bench-begin
#Roadblock: Mon Jul 14 07:59:26 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:setup-bench-end
echo 'Starting iteration 1 sample 1 (test 1 of 1) attempt number 1 of 1'
echo 'Roadblock: Mon Jul 14 07:59:31 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:1-1-1:infra-start-begin'
#Roadblock: Mon Jul 14 07:59:35 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:1-1-1:infra-start-end
#Roadblock: Mon Jul 14 07:59:41 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:1-1-1:server-start-begin
#Roadblock: Mon Jul 14 07:59:46 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:1-1-1:server-start-end
echo 'Roadblock: Mon Jul 14 08:00:01 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:1-1-1:endpoint-start-begin'
#Roadblock: Mon Jul 14 08:00:05 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:1-1-1:endpoint-start-end
#Roadblock: Mon Jul 14 08:01:06 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:1-1-1:client-start-begin
echo 'Found new client-start-end timeout value: 780'
echo 'Roadblock: Mon Jul 14 08:01:11 UTC 2025 role: leader uuid: 861c58fe-283a-4045-905f-e243f1bc857d:1-1-1:client-start-end'
```
Only at this point in time, crucible starts sending data, after `client-start-end`. Note by the timestamps how long this can take (there is more to the runtime prior to the pod deployment omitted for brevity).

Kube-burner:

```
#time="2025-07-14 08:39:37" level=info msg="❤️ Checking for Cluster Health" file="cluster_health.go:46"
#time="2025-07-14 08:39:37" level=info msg="Cluster is Healthy" file="cluster_health.go:54"
#time="2025-07-14 08:39:37" level=info msg="📁 Creating opensearch indexer: indexer-0" file="metrics.go:67"
#time="2025-07-14 08:39:38" level=info msg="👽 Initializing prometheus client with URL: https://xxxx-k8s-xxx-monitoring.apps.vlan607.xxx.xxxx.redhat.com" file="prometheus.go:45"
#time="2025-07-14 08:39:38" level=info msg="🔔 Initializing alert manager for prometheus: https://xxxx-k8s-xxxx-monitoring.apps.vlan607.xxx.xxx.xxx.com" file="alert_manager.go:89"
echo 'time="2025-07-14 08:39:38" level=info msg="🔥 Starting kube-burner (main@78dac190a5d0b2d80d0c75bbe993ba1e219621ba) with UUID b8c1d06e-5953-47ee-a73e-fdd1e3354c70" file="job.go:84"'
#time="2025-07-14 08:39:38" level=info msg="📈 Registered measurement: podLatency" file="factory.go:106"
#time="2025-07-14 08:39:38" level=info msg="📈 Registered measurement: serviceLatency" file="factory.go:106"
#time="2025-07-14 08:39:38" level=info msg="QPS: 20" file="job.go:331"
#time="2025-07-14 08:39:38" level=info msg="Burst: 20" file="job.go:332"
#time="2025-07-14 08:39:38" level=info msg="Job bgp-setup: 1 iterations with 1 IPAddressPool replicas" file="create.go:78"
#time="2025-07-14 08:39:38" level=info msg="Job bgp-setup: 1 iterations with 1 BGPAdvertisement replicas" file="create.go:78"
#time="2025-07-14 08:39:38" level=info msg="Job bgp-setup: 1 iterations with 1 BGPPeer replicas" file="create.go:78"
#time="2025-07-14 08:39:38" level=info msg="QPS: 20" file="job.go:331"
#time="2025-07-14 08:39:38" level=info msg="Burst: 20" file="job.go:332"
#time="2025-07-14 08:39:38" level=info msg="Job rds: 115 iterations with 42 Secret replicas" file="create.go:78"
#time="2025-07-14 08:39:38" level=info msg="Job rds: 115 iterations with 30 ConfigMap replicas" file="create.go:78"
#time="2025-07-14 08:39:38" level=info msg="Job rds: 115 iterations with 1 NetworkPolicy replicas" file="create.go:78"
#time="2025-07-14 08:39:38" level=info msg="Job rds: 115 iterations with 1 NetworkPolicy replicas" file="create.go:78"
#time="2025-07-14 08:39:38" level=info msg="Job rds: 115 iterations with 1 NetworkPolicy replicas" file="create.go:78"
#time="2025-07-14 08:39:38" level=info msg="Job rds: 115 iterations with 2 SriovNetwork replicas" file="create.go:78"
#time="2025-07-14 08:39:38" level=info msg="Job rds: 115 iterations with 20 Service replicas" file="create.go:78"
#time="2025-07-14 08:39:38" level=info msg="Job rds: 115 iterations with 1 Service replicas" file="create.go:78"
#time="2025-07-14 08:39:38" level=info msg="Job rds: 115 iterations with 2 Route replicas" file="create.go:78"
#time="2025-07-14 08:39:38" level=info msg="Job rds: 115 iterations with 15 Deployment replicas" file="create.go:78"
#time="2025-07-14 08:39:38" level=info msg="Job rds: 115 iterations with 10 Deployment replicas" file="create.go:78"
#time="2025-07-14 08:39:38" level=info msg="Job rds: 115 iterations with 2 Deployment replicas" file="create.go:78"
#time="2025-07-14 08:39:38" level=info msg="Pre-load: images from job bgp-setup" file="pre_load.go:71"
#time="2025-07-14 08:39:38" level=info msg="No images found to pre-load, continuing" file="pre_load.go:77"
#time="2025-07-14 08:39:38" level=info msg="Pre-load: images from job rds" file="pre_load.go:71"
#time="2025-07-14 08:39:38" level=info msg="Pre-load: Creating DaemonSet using images [quay.io/xxx-xxx/nginx:latest quay.io/xxx-xxxx/curl:xxxx ghcr.io/xxxxx/fedora-stress-ng:master] in namespace preload-kube-burner" file="pre_load.go:197"
#time="2025-07-14 08:39:38" level=info msg="Pre-load: Sleeping for 15s" file="pre_load.go:84"
#time="2025-07-14 08:39:53" level=info msg="Deleting 1 namespaces with label: kube-burner-preload=true" file="namespaces.go:67"
#time="2025-07-14 08:40:05" level=info msg="Creating Pod latency watcher for bgp-setup" file="pod_latency.go:163"
#time="2025-07-14 08:40:05" level=info msg="Namespace kube-burner-service-latency already exists" file="namespaces.go:44"
#time="2025-07-14 08:40:05" level=warning msg="pods \"svc-checker\" already exists" file="common.go:203"
#time="2025-07-14 08:40:05" level=info msg="Creating service latency watcher for bgp-setup" file="service_latency.go:176"
#time="2025-07-14 08:40:06" level=info msg="Triggering job: bgp-setup" file="job.go:106"
#time="2025-07-14 08:40:06" level=info msg="Deleting IPAddressPools labeled with kube-burner-job=bgp-setup in metallb-system" file="namespaces.go:55"
#time="2025-07-14 08:40:06" level=info msg="Deleting BGPAdvertisements labeled with kube-burner-job=bgp-setup in metallb-system" file="namespaces.go:55"
#time="2025-07-14 08:40:06" level=info msg="Deleting BGPPeers labeled with kube-burner-job=bgp-setup in metallb-system" file="namespaces.go:55"
#time="2025-07-14 08:40:06" level=info msg="0/1 iterations completed" file="create.go:117"
#time="2025-07-14 08:40:06" level=info msg="Waiting up to 4h0m0s for actions to be completed" file="create.go:167"
#time="2025-07-14 08:40:06" level=info msg="Actions completed" file="waiters.go:70"
#time="2025-07-14 08:40:06" level=info msg="Verifying created objects" file="utils.go:119"
#time="2025-07-14 08:40:06" level=info msg="Job bgp-setup took 0s" file="job.go:167"
#time="2025-07-14 08:40:06" level=info msg="Stopping measurement: podLatency" file="factory.go:147"
#time="2025-07-14 08:40:06" level=info msg="Evaluating latency thresholds" file="metrics.go:48"
#time="2025-07-14 08:40:06" level=info msg="Stopping measurement: serviceLatency" file="factory.go:147"
#time="2025-07-14 08:40:06" level=info msg="Deleting 1 namespaces with label: kubernetes.io/metadata.name=kube-burner-service-latency" file="namespaces.go:67"
#time="2025-07-14 08:40:13" level=info msg="Indexing collected data from measurement: serviceLatency" file="factory.go:159"
#time="2025-07-14 08:40:13" level=info msg="Indexing metric svcLatencyMeasurement" file="common.go:105"
#time="2025-07-14 08:40:13" level=info msg="Indexing skipped due to 0 docs" file="common.go:114"
#time="2025-07-14 08:40:13" level=info msg="Indexing metric svcLatencyQuantilesMeasurement" file="common.go:105"
#time="2025-07-14 08:40:13" level=info msg="Indexing skipped due to 0 docs" file="common.go:114"
#time="2025-07-14 08:40:13" level=info msg="Indexing collected data from measurement: podLatency" file="factory.go:159"
#time="2025-07-14 08:40:13" level=info msg="Indexing metric podLatencyMeasurement" file="common.go:105"
#time="2025-07-14 08:40:13" level=info msg="Indexing skipped due to 0 docs" file="common.go:114"
#time="2025-07-14 08:40:13" level=info msg="Indexing metric podLatencyQuantilesMeasurement" file="common.go:105"
#time="2025-07-14 08:40:13" level=info msg="Indexing skipped due to 0 docs" file="common.go:114"
#time="2025-07-14 08:40:13" level=info msg="Creating Pod latency watcher for rds" file="pod_latency.go:163"
#time="2025-07-14 08:40:16" level=info msg="Creating service latency watcher for rds" file="service_latency.go:176"
echo 'time="2025-07-14 08:40:16" level=info msg="Triggering job: rds" file="job.go:106"'
#echo 'time="2025-05-08 08:31:43" level=info msg="Triggering job: node-density-cni" file="job.go:98"'
#time="2025-07-14 08:40:16" level=info msg="Deleting SriovNetworks labeled with kube-burner-job=rds in openshift-sriov-network-operator" file="namespaces.go:55"
#time="2025-07-14 08:40:16" level=info msg="Churning enabled" file="job.go:113"
#time="2025-07-14 08:40:16" level=info msg="Churn cycles: 0" file="job.go:114"
#time="2025-07-14 08:40:16" level=info msg="Churn duration: 1h0m0s" file="job.go:115"
#time="2025-07-14 08:40:16" level=info msg="Churn percent: 10" file="job.go:116"
#time="2025-07-14 08:40:16" level=info msg="Churn delay: 2m0s" file="job.go:117"
#time="2025-07-14 08:40:16" level=info msg="Churn deletion strategy: default" file="job.go:118"
echo 'time="2025-07-14 08:41:25" level=info msg="11/115 iterations completed" file="create.go:117"'
#time="2025-07-14 08:42:35" level=info msg="22/115 iterations completed" file="create.go:117"
#time="2025-07-14 08:43:45" level=info msg="33/115 iterations completed" file="create.go:117"
#time="2025-07-14 08:44:55" level=info msg="44/115 iterations completed" file="create.go:117"
#time="2025-07-14 08:46:05" level=info msg="55/115 iterations completed" file="create.go:117"
#time="2025-07-14 08:47:15" level=info msg="66/115 iterations completed" file="create.go:117"
echo 'time="2025-07-14 08:48:25" level=info msg="77/115 iterations completed" file="create.go:117"'
#time="2025-07-14 08:49:34" level=info msg="88/115 iterations completed" file="create.go:117"
#time="2025-07-14 08:50:44" level=info msg="99/115 iterations completed" file="create.go:117"
echo 'time="2025-07-14 08:51:54" level=info msg="110/115 iterations completed" file="create.go:117"'
echo 'time="2025-07-14 08:52:26" level=info msg="Waiting up to 4h0m0s for actions to be completed" file="create.go:167"'
#time="2025-07-14 08:52:28" level=info msg="Actions in namespace rds-11 completed" file="waiters.go:68"
#time="2025-07-14 08:53:07" level=info msg="Actions in namespace rds-114 completed" file="waiters.go:68"
#time="2025-07-14 08:52:28" level=info msg="Actions in namespace rds-11 completed" file="waiters.go:68"
#time="2025-07-14 08:53:07" level=info msg="Actions in namespace rds-114 completed" file="waiters.go:68"
#time="2025-07-14 08:53:07" level=info msg="Verifying created objects" file="utils.go:119"
#time="2025-07-14 08:53:12" level=info msg="Deleting 11 namespaces with label: churndelete=delete" file="namespaces.go:67"
#time="2025-07-14 08:54:05" level=info msg="Re-creating deleted objects" file="create.go:365"
(...)
echo 'time="2025-07-14 09:44:35" level=info msg="Garbage collection metrics on, waiting for GC" file="job.go:199"'
```
In the example above, rds-core deployes 115 iterations (115 namespaces), and churn only starts after `Verifying created objects`. Note the time it takes by the corresponding timestamps.

CC. @vishnuchalla as requested.

sanjay7178:
## Type of change

- New feature

## Description

As discussed with @rsevilla87 and @sferlin , this is the initial POC to hook support for job execution stages and background processing

> Didn't test yet with crucible/netperf data planes on k8s to verify it 

things i've implemented : 
- Added `pkg/burner/hooks.go` to manage hook process and their process life cycle in foreground and background , with goroutines , wait groups , channels , mutexes for efficient handling
- Added `HookBeforeDeployment` , `HookOnEachIteration` , `HookAfterDeployment` , `HookBeforeChurn` , `HookAfterChurn` hooks execution logic in `pkg/burner/create.go`
- Added additional struct fields to `JobExecutor`  at `pkg/burner/executor.go` for hook handling at different levels
- Added `HookBeforeGC` , `HookAfterGC` hooks execution logic for churn operation in `pkg/burner/job.go` and also Refactored `pkg/burner/job.go`  to manage `HookBeforeCleanup` execution
- Added required yaml fields for hooks at `pkg/config/types.go`

 

usage : 
``` bash
kube-burner init -c ./config-basic-hooks.yaml --log-level=debug --uuid="hook-test-$(date +%s)"
```

files: 

`config-basic-hooks.yaml`
```yaml 
global:
  gc: false
  waitWhenFinished: true
  measurements:
    - name: podLatency

jobs:
  - name: basic-hook-test
    jobType: create
    jobIterations: 5
    qps: 5
    burst: 5
    namespace: hook-test
    namespacedIterations: true
    verifyObjects: true
    
    # Hook configuration
    hooks:
      # Before deployment starts
      - cmd: ["/bin/bash", "test/hooks/simple-hook.sh", "beforeDeployment"]
        when: beforeDeployment
        background: false
      
      # After deployment completes
      - cmd: ["/bin/bash", "test/hooks/simple-hook.sh", "afterDeployment"]
        when: afterDeployment
        background: false
      
      # Before cleanup
      - cmd: ["/bin/bash", "test/hooks/simple-hook.sh", "beforeCleanup"]
        when: beforeCleanup
        background: false
      
      # On each iteration
      - cmd: ["/bin/bash", "test/hooks/simple-hook.sh", "onEachIteration"]
        when: onEachIteration
        background: false
  
    # Churn configuration
    churnConfig:
      duration: 30s
      percent: 20
      delay: 5s
      
    objects:
      - objectTemplate: test/hooks/pod-template.yaml
        replicas: 2
        inputVars:
          containerImage: busybox:latest
```

`simple-hook.sh` 

```bash
set -e

HOOK_LOG_DIR="/tmp/kube-burner-hooks"
mkdir -p "$HOOK_LOG_DIR"

TIMESTAMP=$(date '+%Y-%m-%d %H:%M:%S')
HOOK_PHASE="${1:-unknown}"
LOG_FILE="$HOOK_LOG_DIR/hook-${HOOK_PHASE}.log"

echo "========================================" | tee -a "$LOG_FILE"
echo "Hook Execution: $HOOK_PHASE" | tee -a "$LOG_FILE"
echo "Timestamp: $TIMESTAMP" | tee -a "$LOG_FILE"
echo "PID: $$" | tee -a "$LOG_FILE"
echo "Environment:" | tee -a "$LOG_FILE"
env | grep -E "KUBE_BURNER|JOB_|UUID" | tee -a "$LOG_FILE" || echo "No kube-burner env vars" | tee -a "$LOG_FILE"
echo "========================================" | tee -a "$LOG_FILE"

# Simulate some work
sleep 1

echo "Hook $HOOK_PHASE completed successfully" | tee -a "$LOG_FILE"
exit 0
```

- Closes #1041


sanjay7178:
`pod-template.yaml`


```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-pod-{{.Iteration}}-{{.Replica}}
  labels:
    app: hook-test
    iteration: "{{.Iteration}}"
    replica: "{{.Replica}}"
spec:
  containers:
  - name: pause
    image: {{.containerImage}}
    resources:
      requests:
        memory: "32Mi"
        cpu: "10m"
      limits:
        memory: "64Mi"
        cpu: "50m"
  restartPolicy: Never
  terminationGracePeriodSeconds: 0

```

pkg/burner/hooks.go
```diff
@@ -0,0 +1,355 @@
...
+	"os/exec"
+	"sync"
+	"syscall"
+	"time"
+
+	"github.com/kube-burner/kube-burner/v2/pkg/config"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
```
rsevilla87: Can you create a new type, `jobHook` to declare these constants?

pkg/burner/create.go
```diff
@@ -126,7 +126,15 @@ func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterati
 	// We have to sum 1 since the iterations start from 1
 	iterationProgress := (iterationEnd - iterationStart) / 10
 	percent := 1
+
+	if err := ex.executeHooks(HookBeforeDeployment); err != nil {
```
rsevilla87: This hook shouldn't be here, it should be called just before calling RunCreateJob and jobExecutor.Run()

pkg/burner/create.go
```diff
@@ -126,7 +126,15 @@ func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterati
...
+		log.Errorf("Error executing hooks for %s: %v", HookBeforeDeployment, err)
+	}
 	for i := iterationStart; i < iterationEnd; i++ {
+		// Execute onEachIteration hooks
+		if err := ex.executeHooks(HookOnEachIteration); err != nil {
+			log.Errorf("Error executing hooks for %s: %v", HookOnEachIteration, err)
```
rsevilla87:
```suggestion
			log.Errorf("Error executing %s hooks: %v", HookOnEachIteration, err)
```

pkg/burner/create.go
```diff
@@ -190,6 +198,9 @@ func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterati
 			waitErrors = append(waitErrors, errs...)
 		}
 	}
+	if err := ex.executeHooks(HookAfterDeployment); err != nil {
```
rsevilla87: This hook shouldn't be here, it should be called right after calling RunCreateJob() and jobExecutor.Run()

pkg/burner/create.go
```diff
@@ -374,6 +385,10 @@ func (ex *JobExecutor) createNamespace(ns string, nsLabels, nsAnnotations map[st
 func (ex *JobExecutor) RunCreateJobWithChurn(ctx context.Context) []error {
 	// Cleanup namespaces based on the labels we added to the objects
 	log.Infof("Churning mode: %s", ex.ChurnConfig.Mode)
+	// Execute beforeChurn hooks
+	if err := ex.executeHooks(HookBeforeChurn); err != nil {
```
rsevilla87: This hook is not very useful, it overlaps with `HookAfterDeployment`, and can be confusing.

pkg/burner/executor.go
```diff
@@ -72,6 +76,14 @@ func newExecutor(configSpec config.Spec, kubeClientProvider *config.KubeClientPr
 		embedCfg:          embedCfg,
 		deletionStrategy:  configSpec.GlobalConfig.DeletionStrategy,
 		objectOperations:  0,
+		backgroundHooks:   make([]*hookProcess, 0),
+		Hooks:             job.Hooks,
+		hookManager:       NewHookManager(context.Background()),
+	}
+
+	// Validate hooks
+	if err := validateHooks(job.Hooks); err != nil {
```
rsevilla87: This is a config validation, can you do it in config.go?

pkg/burner/job.go
```diff
@@ -159,16 +167,24 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
...
 			}
-			if jobExecutor.BeforeCleanup != "" {
-				log.Infof("Waiting for beforeCleanup command %s to finish", jobExecutor.BeforeCleanup)
-				stdOut, stdErr, err := util.RunShellCmd(jobExecutor.BeforeCleanup, jobExecutor.embedCfg)
-				if err != nil {
-					err = fmt.Errorf("BeforeCleanup failed: %v", err)
-					log.Error(err.Error())
-					errs = append(errs, err)
-					innerRC = 1
+			if jobExecutor.BeforeCleanup != "" || len(jobExecutor.Hooks) > 0 {
```
rsevilla87, line 201: Along with this PR, we should deprecate the redundant BeforeCleanup config field.

pkg/burner/hooks.go
```diff
@@ -0,0 +1,355 @@
...
+}
+
+func validateHooks(hooks []config.Hook) error {
+	validWhen := map[string]bool{
+		HookBeforeDeployment: true,
+		HookAfterDeployment:  true,
+		HookBeforeChurn:      true,
+		HookAfterChurn:       true,
+		HookBeforeCleanup:    true,
+		HookAfterCleanup:     true,
```
rsevilla87: Not used

pkg/burner/job.go
```diff
@@ -220,6 +236,7 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
 		if globalConfig.GC {
 			//nolint:govet
 			for _, jobExecutor := range jobExecutors {
+				// jobExecutor.cleanupBackgroundHooks()
```
rsevilla87: ?

pkg/burner/hooks.go
```diff
@@ -0,0 +1,355 @@
+// Copyright 2024 The Kube-burner Authors.
```
rsevilla87:
```suggestion
// Copyright 2026 The Kube-burner Authors.
```

pkg/burner/hooks.go
```diff
@@ -0,0 +1,355 @@
...
+const (
+	HookBeforeDeployment = "beforeDeployment"
+	HookAfterDeployment  = "afterDeployment"
+	HookBeforeChurn      = "beforeChurn"
+	HookAfterChurn       = "afterChurn"
+	HookBeforeCleanup    = "beforeCleanup"
+	HookAfterCleanup     = "afterCleanup"
+	HookBeforeGC         = "beforeGC"
+	HookAfterGC          = "afterGC"
+	HookOnEachIteration  = "onEachIteration"
```
rsevilla87: The hook names should be aligned with kube-burner terminology

```suggestion
	HookBeforeDeployment = "beforeJobExecution"
	HookAfterDeployment  = "afterJobExecution"
	HookBeforeChurn      = "beforeChurn"
	HookAfterChurn       = "afterChurn"
	HookBeforeCleanup    = "beforeCleanup"
	HookAfterCleanup     = "afterCleanup"
	HookBeforeGC         = "beforeGC"
	HookAfterGC          = "afterGC"
	HookOnEachIteration  = "onEachIteration"
```

pkg/burner/hooks.go
```diff
@@ -0,0 +1,355 @@
...
+	resultChan := make(chan hookResult, len(ex.Hooks))
+
+	// var backgroundProcesses []*hookProcess
+	for _, hook := range ex.Hooks {
+		if hook.When != when {
+			continue
+		}
+
+		if len(hook.CMD) == 0 {
+			log.Warnf("Hook for %s has been empty command, skipping", when)
```
rsevilla87:
```suggestion
			log.Warnf("Empty command for hook %s, skipping it", when)
```

pkg/burner/hooks.go
```diff
@@ -0,0 +1,355 @@
...
+			continue
+		}
+		if hook.Background {
+			// Start in background , store process
+			if err := ex.executeBackgroundHook(hook, when); err != nil {
+				// log.Errorf("Error executing background hook for %s: %v", when, err)
+				errorsMu.Lock()
+				_ = append(errorInternal, fmt.Errorf("failed to start background hook at '%s': %w", when, err))
```
rsevilla87: Why are you using this syntax instead of 

```golang
errInternal = append(errorInternal, fmt.Errorf("failed to run '%s' background hook : %w", when, err))
```

pkg/burner/hooks.go
```diff
@@ -0,0 +1,355 @@
...
+				errorsMu.Unlock()
+				return err
+			}
+		} else {
+			foregroundWg.Add(1)
+			go func(h config.Hook) {
```
rsevilla87: For the sake of simplicity, we should use the more simple implementation as possible for this feature, running the hooks one after another fits better in my opinion, there's no need to run them in parallel.

pkg/burner/hooks.go
sanjay7178: as errorInternal array is been appended along pass through in functions , so I thought it would be more verbose to log the errors on every err nil checks , for prototyping the RFE I did that .

pkg/burner/executor.go
sanjay7178: Sure , I'll make the changes needed

pkg/burner/job.go
sanjay7178: according to the issue https://github.com/kube-burner/kube-burner/issues/1041 I believe to manage hooks efficiently with goroutines , but unsure about whether to implement background hooks goroutines cleanup after job execution , so I commented the cleanup implementation

pkg/burner/job.go
sanjay7178: Sure , I'll have a note of it

pkg/burner/create.go
sanjay7178: Ok sure I will remove it .

sanjay7178: Should I remove  `HookAfterChurn` as well ?

pkg/burner/job.go
sanjay7178: as we deprecate it , should the `HookBeforeCleanup` also to be removed ?

pkg/burner/hooks.go
sanjay7178: `HookAfterCleanup` is not been used so far , I've added it in `job.go` ,

pkg/burner/hooks.go
sanjay7178: Got it

pkg/config/config.go
```diff
@@ -423,6 +426,47 @@ func jobIsDuped() error {
 	return nil
 }
 
+func validateHooks() error {
+	hooks := configSpec.Jobs[0].Hooks
+	validWhen := map[JobHook]bool{
+		HookBeforeDeployment: true,
```
vishnuchalla: Can we rename it to `HookBeforeWorkload`? Deployment might confuse folks with k8s Deployment.

pkg/burner/hooks.go
```diff
@@ -0,0 +1,298 @@
...
+type hookResult struct {
+	hook     config.Hook
+	when     config.JobHook
+	err      error
+	duration time.Duration
+	stdout   string
+	stderr   string
+}
+
+type hookProcess struct {
```
vishnuchalla: There are some fields in this type that are in common with `hookResult`. Can we avoid redundancy here please? and have `hookResult` be a part of `hookProcess` instead?

pkg/config/config.go
sanjay7178: Sure , I'll make the changes

pkg/burner/hooks.go
sanjay7178: Sure , I'll make the changes

pkg/burner/hooks.go
```diff
@@ -0,0 +1,298 @@
...
+
+// NewHookManager creates a new HookManager
+func NewHookManager(ctx context.Context) *HookManager {
+	ctx, cancel := context.WithCancel(ctx)
+
+	return &HookManager{
+		backgroundHooks: make([]*hookProcess, 0),
+		ctx:             ctx,
+		cancel:          cancel,
+		resultChan:      make(chan hookResult, 100), // Buffered channel to avoid blocking
```
vishnuchalla: I think size of this channel should be same as number of hooks. Or just initialize empty and keep on appending stuff later in the runtime.

pkg/burner/hooks.go
vishnuchalla: Yes, how about just letting all the background hooks go in parallel and just have foreground ones run in a sequence? just like a shell command w/ and w/o `&` to it.

pkg/burner/hooks.go
```diff
@@ -0,0 +1,298 @@
...
+			return fmt.Errorf("hook timed out after %v at '%s': %w", timeout, when, err)
+		}
+		return fmt.Errorf("hook failed at '%s' after %v: %w", when, duration, err)
+	}
+
+	log.Infof("Hook completed at '%s' in %v", when, duration)
+	return nil
+}
+
+// WaitForBackgroundHooks waits for all background hooks to complete with timeout
```
vishnuchalla: Where are these two method even being used? Can we just have this logic be incorporated into `monitorBackgroundHook`?

pkg/config/config.go
```diff
@@ -423,6 +426,47 @@ func jobIsDuped() error {
...
+		}
+		if len(hook.Cmd) == 0 {
+			return fmt.Errorf("hook %d has empty command", i)
+		}
+	}
+
+	return nil
+}
+
+func getValidWhenValues() []JobHook {
```
vishnuchalla: We do not need to list the hook types twice to validate them. Can you condense this logic to have the entire list once and use the same for validating and at the same time listing when errored out.

pkg/config/config.go
sanjay7178: made the changes for this at https://github.com/kube-burner/kube-burner/pull/1068/commits/a3c6867a9d106dd3317ea358428ef5b4b9b97dc1

pkg/burner/hooks.go
sanjay7178: made the changes for this at https://github.com/kube-burner/kube-burner/pull/1068/commits/a3c6867a9d106dd3317ea358428ef5b4b9b97dc1

pkg/burner/hooks.go
sanjay7178: I just used un-buffered channel to make it simple instead of making restrict via number of  background hooks in config

pkg/config/config.go
sanjay7178: The map literal provides O(1) validation performance while the slice function enables readable error message ,so for the two distinct data structures I made to optimized for their specific use cases

pkg/burner/hooks.go
sanjay7178: I intentionally kept these separate because I previously had `cleanupBackgroundHooks()` for resource cleanup/cancellation. While I removed it pending discussion with @rsevilla87, keeping the wait/result methods separate maintains the infrastructure for re-adding cleanup logic without refactoring the entire monitoring flow.

sanjay7178: It's a unnecessary method for now , so i will remove it.

pkg/config/config.go
sanjay7178: As par with DRY principle if it's needed to remove duplicated validation checks , please let me know

pkg/burner/hooks.go
vishnuchalla: can we just initialize it as an empty channel and then let it append in the runtime?

pkg/config/config.go
vishnuchalla: We can just have one data structure help us do the both actions IMHO.

pkg/burner/hooks.go
sanjay7178: Sorry If i use a buffered/Nil Channel , if no one is reading at the exact moment of send, the default case executes and the result is lost. Also  with unbuffered channels require both sender and receiver to be ready simultaneously

sanjay7178: I would rather do maintain the channel size based on number of hooks

sanjay7178: Like this 
```go
// NewHookManager creates a new HookManager
func NewHookManager(ctx context.Context, configSpec config.Spec) *HookManager {
	ctx, cancel := context.WithCancel(ctx)
	channelSize := 0
	if len(configSpec.Jobs) > 0 {
		channelSize = func() int {
			j := 0
			for _, job := range configSpec.Jobs {
				j += len(job.Hooks)
			}
			return j
		}()
	}
	return &HookManager{
		backgroundHooks: make([]*hookProcess, 0),
		ctx:             ctx,
		cancel:          cancel,
		resultChan:      make(chan hookResult, channelSize),
	}
}
```

pkg/burner/hooks.go
```diff
@@ -0,0 +1,265 @@
...
+			continue
+		}
+
+		if len(hook.Cmd) == 0 {
+			log.Warnf("Empty command for hook %s, skipping it", when)
+			continue
+		}
+		if hook.Background {
+			// Start in background, store process
+			if err := ex.executeBackgroundHook(hook, when); err != nil {
```
ygalblum: We've already established that `when` is equal to `hook.When`, so no need to pass it

pkg/burner/executor.go
```diff
@@ -72,6 +76,9 @@ func newExecutor(configSpec config.Spec, kubeClientProvider *config.KubeClientPr
 		embedCfg:          embedCfg,
 		deletionStrategy:  configSpec.GlobalConfig.DeletionStrategy,
 		objectOperations:  0,
+		backgroundHooks:   make([]*hookProcess, 0),
+		Hooks:             job.Hooks,
```
ygalblum: `JobExecutor` already extends `Job` so, `job.Hooks` is already available. No need to define it again

pkg/burner/job.go
ygalblum: With the introduction of `HookBeforeCleanup` (as part of the entire hooks mechanism), there is no need for the specific `BeforeCleanup` job parameter. But, the former should still be supported.

@rsevilla87 does this mean we should release this only as a 3.x? It will break existing configurations

pkg/burner/job.go
ygalblum: Either uncomment the code, or remove it. Don't commit commented code

pkg/burner/hooks.go
```diff
@@ -0,0 +1,265 @@
...
+	cancel          context.CancelFunc
+	resultChan      chan hookResult
+}
+
+// NewHookManager creates a new HookManager
+func NewHookManager(ctx context.Context, configSpec config.Spec) *HookManager {
+	ctx, cancel := context.WithCancel(ctx)
+	channelSize := 0
+	if len(configSpec.Jobs) > 0 {
+		channelSize = len(configSpec.Jobs[0].Hooks)
```
ygalblum: How will this work for the rest of the Jobs (e.g. in index 3)?

pkg/burner/hooks.go
ygalblum: We should define the order. That is, will all background hooks be started before all foreground? Or will it be mixed based on the order in the configuration file?

pkg/burner/hooks.go
```diff
@@ -0,0 +1,265 @@
...
+		}
+	}
+	go func() {
+		foregroundWg.Wait()
+		close(resultChan)
+	}()
+
+	for err := range resultChan {
+		errorsMu.Lock()
+		errorInternal = append(errorInternal, err.err)
```
ygalblum: `errorInternal` is a method variable and you update it in the sync part of the method. Are you sure you need a mutex?

pkg/burner/hooks.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	if len(errorInternal) > 0 {
+		return fmt.Errorf("hook execution errorInternal at '%s': %v", when, errorInternal)
+	}
+
+	return nil
+}
+
+func (ex *JobExecutor) executeBackgroundHook(hook config.Hook, when config.JobHook) error {
```
ygalblum: This method should belong to the `HookManager`

pkg/burner/hooks.go
```diff
@@ -0,0 +1,265 @@
...
+	ex.hookManager.mu.Unlock()
+
+	ex.hookManager.wg.Add(1)
+	go ex.monitorBackgroundHook(hp, &stdout, &stderr)
+
+	return nil
+}
+
+// monitorBackgroundHook monitors a background hook with proper error handling
+func (ex *JobExecutor) monitorBackgroundHook(hp *hookProcess, stdout, stderr *bytes.Buffer) {
```
ygalblum: This method should belong to the `HookManager`

pkg/burner/hooks.go
```diff
@@ -0,0 +1,265 @@
...
+				log.Debugf("Hook stdout: %s", result.stdout)
+			}
+		}
+
+	case <-ex.hookManager.ctx.Done():
+		log.Warnf("Hook monitor cancel for '%s'", hp.result.when)
+	}
+}
+
+func (ex *JobExecutor) executeForegroundHook(hook config.Hook, when config.JobHook) error {
```
ygalblum: This method should belong to the `HookManager`

pkg/burner/hooks.go
```diff
@@ -0,0 +1,265 @@
...
+		}
+		return fmt.Errorf("hook failed at '%s' after %v: %w", when, duration, err)
+	}
+
+	log.Infof("Hook completed at '%s' in %v", when, duration)
+	return nil
+}
+
+// GetBackgroundHookResults returns results from background hooks (non-blocking)
+func (ex *JobExecutor) GetBackgroundHookResults() []hookResult {
```
ygalblum: This method should belong to the `HookManager`

pkg/config/config.go
```diff
@@ -423,6 +426,38 @@ func jobIsDuped() error {
 	return nil
 }
 
+func HookBeforeWorkload() error {
+	// No jobs/Hooks defined
+	if len(configSpec.Jobs) == 0 || len(configSpec.Jobs[0].Hooks) == 0 {
```
ygalblum: What about the other jobs?

sanjay7178: I seen few of the config's having single job in present repository , so I made a mistake and did make it limited to 1 job , sure I will change it

pkg/burner/hooks.go
sanjay7178: I seen few of the config's having single job in present repository , so I made a mistake and did make it limited to 1 job , sure I will change it

sanjay7178: duplicate : https://github.com/kube-burner/kube-burner/pull/1068#discussion_r2683517449

pkg/burner/hooks.go
sanjay7178: yes the mutex is unnecessary in this specific implementation . The reason I felt to have to mutex is the Multiple goroutines can safely send to `resultChan`,  the main goroutine safely receives them one at a time and  I thought the `errorInternal` slice is might interfere between the main goroutine .

pkg/burner/hooks.go
vishnuchalla: Maintaining size based on number of hooks, sounds good to me.

pkg/burner/hooks.go
```diff
@@ -59,7 +59,13 @@ func NewHookManager(ctx context.Context, configSpec config.Spec) *HookManager {
 	ctx, cancel := context.WithCancel(ctx)
 	channelSize := 0
 	if len(configSpec.Jobs) > 0 {
-		channelSize = len(configSpec.Jobs[0].Hooks)
+		channelSize = func() int {
```
ygalblum: Why do you need an inline function? Just compute `channelSize` the same way you compute `j`

pkg/burner/hooks.go
```diff
@@ -95,17 +100,17 @@ func (ex JobExecutor) executeHooks(when config.JobHook) error {
...
 		} else {
 			foregroundWg.Add(1)
 			go func(h config.Hook) {
 				defer foregroundWg.Done()
-				if err := ex.executeForegroundHook(h, when); err != nil {
+				if err := ex.hookManager.executeForegroundHook(h); err != nil {
 					resultChan <- hookResult{
 						hook: h,
-						when: when,
+						when: h.When,
```
ygalblum: `when` is already in the `hook`. So whoever gets the result can get it from there (`hook.When`). No need to pass it again

pkg/burner/executor.go
```diff
@@ -58,6 +59,8 @@ type JobExecutor struct {
 	deletionStrategy  string
 	objectOperations  int32
 	nsChurning        bool
+	backgroundHooks   []*hookProcess
```
ygalblum: Where is this field used? I couldn't find it

pkg/burner/hooks.go
```diff
@@ -0,0 +1,251 @@
...
+type hookProcess struct {
+	cmd       *exec.Cmd
+	startTime time.Time
+	mu        sync.RWMutex
+	done      chan struct{}
+	result    hookResult
+}
+
+type HookManager struct {
+	backgroundHooks []*hookProcess
```
ygalblum: Why is `backgroundHooks` needed? I only see it being initialized and appended. What am I missing?

pkg/burner/hooks.go
```diff
@@ -0,0 +1,251 @@
...
+		hp.result.duration = time.Since(hp.startTime)
+		hp.result.stdout = stdout.String()
+		hp.result.stderr = stderr.String()
+		result := hp.result
+
+		hp.mu.Unlock()
+
+		// Send result to channel (non-blocking)
+		select {
+		case hm.resultChan <- result:
```
ygalblum, line 128: Who's listening to this channel?

pkg/burner/executor.go
```diff
@@ -58,6 +59,8 @@ type JobExecutor struct {
...
+	hookManager       *HookManager
```
ygalblum, line 62: Why is the `HookManager` part of the `JobExecutor`? You initialize a new one for each job, but don't store anything job specific on it. So, each `JobExecutor` will hold a new instance that looks kinda the same. 
The `HookManager` should either have job specific details (for example channel in the length of the number of hooks of the specific job) or be one for all jobs

pkg/burner/hooks.go
```diff
@@ -0,0 +1,248 @@
...
+	ctx, cancel := context.WithCancel(ctx)
+	return &HookManager{
+		ctx:        ctx,
+		cancel:     cancel,
+		resultChan: make(chan hookResult, hookCount),
+	}
+}
+
+// executeHooks executes hooks based on the specified timing (when)
+func (ex JobExecutor) executeHooks(when config.JobHook) error {
```
ygalblum: I think this method should also belong to the `HookManager`. This will allow for better encapsulation

pkg/burner/hooks.go
```diff
@@ -0,0 +1,248 @@
...
+				if err := ex.hookManager.executeForegroundHook(h); err != nil {
+					resultChan <- hookResult{
+						hook: h,
+						err:  err,
+					}
+				}
+			}(hook)
+		}
+	}
+	go func() {
```
ygalblum: I could be wrong here. But, why does this need to run in a go func? Don't we want to wait within the sync part?

pkg/burner/hooks.go
```diff
@@ -0,0 +1,248 @@
...
+
+	hm.wg.Add(1)
+	go hm.monitorBackgroundHook(hp, &stdout, &stderr)
+
+	return nil
+}
+
+// monitorBackgroundHook monitors a background hook with proper error handling
+func (hm *HookManager) monitorBackgroundHook(hp *hookProcess, stdout, stderr *bytes.Buffer) {
+	defer hm.wg.Done()
```
ygalblum: Who's waiting on the `WaitGroup`? I only see calls for `Add` and `Done`

pkg/burner/hooks.go
```diff
@@ -0,0 +1,248 @@
...
+	defer close(hp.done)
```
ygalblum: Who's listening on the channel?

pkg/burner/hooks.go
```diff
@@ -0,0 +1,248 @@
...
+
+	// Wait for process with timeout context
+	errChan := make(chan error, 1)
+	go func() {
+		errChan <- hp.cmd.Wait()
+	}()
+
+	select {
+	case err := <-errChan:
+		hp.mu.Lock()
```
ygalblum: Why do you need a mutex here? Each `hp` is monitored by a separate function. Furthermore, I'm not sure why you need to keep the `result` on the `hp` is you're copying it to `result` and sending it to the channel. You can pass the `hook` to this method and just use a local variable. Moreover, it seems that you only use the `When` field of the hook. So, why not just pass it

pkg/burner/hooks.go
```diff
@@ -0,0 +1,248 @@
...
+		return fmt.Errorf("hook failed at '%s' after %v: %w", hook.When, duration, err)
+	}
+
+	log.Infof("Hook completed at '%s' in %v", hook.When, duration)
+	return nil
+}
+
+// GetBackgroundHookResults returns results from background hooks (non-blocking)
+func (hm *HookManager) GetBackgroundHookResults() []hookResult {
+	if hm == nil {
```
ygalblum: You don't check this anywhere else, why here?

pkg/burner/job.go
```diff
@@ -126,10 +134,30 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
...
+							if result.err != nil {
+								log.Errorf("Background hook failed: %v (duration: %v)", result.err, result.duration)
+								errs = append(errs, result.err)
+								innerRC = 1
+							} else {
+								log.Infof("Background hook completed successfully (duration: %v)", result.duration)
+							}
+						}
+					}
+				}
```
ygalblum: - `hookManager` is set as part of the initialization of the `jobExecutor`. You don't check it anywhere else. No need to check it here
- No need to check `len(backgroundResults) > 0`, `range` will do that for you
 
```suggestion
				if len(jobExecutor.Hooks) > 0 && jobExecutor.hookManager != nil {
					backgroundResults := jobExecutor.hookManager.GetBackgroundHookResults()
					for _, result := range backgroundResults {
						if result.err != nil {
							log.Errorf("Background hook failed: %v (duration: %v)", result.err, result.duration)
							errs = append(errs, result.err)
							innerRC = 1
						} else {
							log.Infof("Background hook completed successfully (duration: %v)", result.duration)
						}
					}
				}
```

pkg/config/config.go
```diff
@@ -423,6 +426,46 @@ func jobIsDuped() error {
...
+	numHooks := func() int {
+		j := 0
+		for i := 0; i < len(configSpec.Jobs); i++ {
+			j += len(configSpec.Jobs[i].Hooks)
+		}
+		return j
+	}
+	if len(configSpec.Jobs) == 0 || numHooks() == 0 {
+		return nil
+	}
```
ygalblum: You don't need any of this code. If there are no jobs or no hooks in the jobs then the validation will just pass

pkg/burner/hooks.go
sanjay7178: yes you're correct , I just followed the concurrent pattern given by copilot . This might lead to deadlock state

pkg/burner/hooks.go
sanjay7178: yes , I got to know this goes under single goroutine , no need of mutex . I would make it simple by removing uneccessary `hookProcess` struct , mutex and `hp.Result`

sanjay7178:
The current progress of the PR uses simple goroutines and waitgroup without  mutexes . 

@ygalblum @rsevilla87 @vishnuchalla Thanks for the review , I've done this PR implementation with some portion of copilot help , I appreciate your patience on reviewing this PR on every minute details. I got to know , I was lacking in go concurrency patterns while resolving the PR .

pkg/burner/hooks.go
ygalblum: @rsevilla87 @vishnuchalla let's answer the discussion here so the PR can continue

pkg/burner/hooks.go
```diff
@@ -0,0 +1,221 @@
...
+			duration: duration,
+			stdout:   stdout.String(),
+			stderr:   stderr.String(),
+		}
+
+		// Send result to channel (non-blocking)
+		select {
+		case hm.resultChan <- result:
+		default:
+			log.Warnf("Result channel full, logging directly")
```
ygalblum: Not sure I understand the log message here. The lines below, writing to the logs will run regardless to whether writing to the channel succeeded or not.

pkg/burner/job.go
```diff
@@ -126,10 +134,29 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
...
 				if jobErrs := jobExecutor.RunCreateJob(ctx, 0, jobExecutor.JobIterations); jobErrs != nil {
 					errs = append(errs, jobErrs...)
 					innerRC = 1
 				}
+				if err := jobExecutor.hookManager.executeHooks(jobExecutor.Hooks, config.HookAfterDeployment); err != nil {
+					log.Errorf("Error executing hooks for %s: %v", config.HookAfterDeployment, err)
+				}
+				// Collect background hook results
+				if len(jobExecutor.Hooks) > 0 && jobExecutor.hookManager != nil {
```
ygalblum: `hookManager` is set during the initialization of `jobExecutor`. So, it's guarantied to not be `nil`. Same as when you don't check it before calling `executeHooks`
```suggestion
				if len(jobExecutor.Hooks) > 0  {
```

pkg/burner/job.go
rsevilla87: That's right @ygalblum , let's support both for the time being, I think is too soon to have a 3.x release
cc @sanjay7178

pkg/burner/hooks.go
rsevilla87: An approach that basically follows the order of the hooks in the list is fine to me regardless they run in background or foreground

pkg/config/config.go
```diff
@@ -423,6 +426,34 @@ func jobIsDuped() error {
...
+		HookAfterCleanup:     true,
+		HookBeforeGC:         true,
+		HookAfterGC:          true,
+		HookOnEachIteration:  true,
+	}
+
+	for _, job := range configSpec.Jobs {
+		for i, hook := range job.Hooks {
+			if !validWhen[hook.When] {
+				return fmt.Errorf("hook %d has invalid 'when' value: %s (valid: %v)",
```
rsevilla87:
```suggestion
				return fmt.Errorf("unsupported when value in %s hook %d: %s, (supported: %v)", job.Name, i, hook.When, maps.Keys(validWhen)",
```

pkg/config/config.go
```diff
@@ -423,6 +426,34 @@ func jobIsDuped() error {
...
+					i, hook.When, maps.Keys(validWhen))
+			}
+			if len(hook.Cmd) == 0 {
+				return fmt.Errorf("hook %d has empty command", i)
```
rsevilla87:
```suggestion
				return fmt.Errorf("hook %d in job %s has empty command", i, job.Name)
```

pkg/burner/hooks.go
```diff
@@ -0,0 +1,220 @@
...
+	"sync"
+	"syscall"
+	"time"
+
+	"github.com/kube-burner/kube-burner/v2/pkg/config"
+	log "github.com/sirupsen/logrus"
+)
+
+type hookResult struct {
+	hook     config.Hook
```
ygalblum: Why is the `hook` field needed? I don't see it being used from the `result`. Same for `stderr` and `stdout`

pkg/burner/job.go
ygalblum: Fine by me

ygalblum (commented):
Sorry for not seeing this until now. But this PR must include test code - either a new test or integrate this feature into an existing test

sanjay7178:
> Sorry for not seeing this until now. But this PR must include test code - either a new test or integrate this feature into an existing test

Sure i'll add them

test/k8s/kind.yml
```diff
@@ -2,4 +2,4 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
   - role: control-plane
-  - role: worker
+  - role: worker
```
ygalblum: I guess this is something automatic in your IDE. But, please do not commit this change (the next person will have a different configuration and we'll keep ping-ponging on it)

test/helpers.bash
```diff
@@ -302,3 +302,63 @@ check_metrics_not_created_for_job() {
...
+    esac
+    
+    echo "FAIL: ${hook} executed ${count} times, expected ${operator} ${expected}"
+    return 1
+  }
+
+  # Verify reliable job lifecycle hooks
+  check_hook "beforeJobExecution" 1 || failed=1
+  check_hook "afterJobExecution" 1 || failed=1
+  check_hook "onEachIteration" ${job_iterations} ge || failed=1
```
ygalblum, line 353: Why is this `ge` and not `eq`?

test/test-k8s.bats
```diff
@@ -68,6 +69,29 @@ teardown_file() {
...
+@test "kube-burner hooks execution verification" {
+  export JOB_NAME="basic-hook-test"
+  export BATS_TEST_TIMEOUT=300
+  rm -rf /tmp/kube-burner-hooks
+  mkdir -p /tmp/kube-burner-hooks
+  
+  run_cmd ${KUBE_BURNER} init -c kube-burner-hooks.yml --uuid=${UUID} --log-level=info
+  
+  # Manual cleanup to trigger GC hooks if they haven't fired
+  kubectl delete ns -l kube-burner.io/job=${JOB_NAME} --ignore-not-found --wait=false 2>/dev/null || true
```
ygalblum: This call does not trigger anything from `kube-burner` as you delete the resources directly using `kubectl`. But, first, is it really needed?

pkg/burner/hooks.go
vishnuchalla: @ygalblum sounds good to me. First we can have all the background hooks triggered in parallel and then go with the foreground ones that run sequential, at the end we can wait for all the hook processes to finish and proceed with rest of the workflow.

sanjay7178:
As per mentioned in this discussion https://github.com/kube-burner/kube-burner/pull/1068#discussion_r2671910424 , 

-  Now `executeHooks()` first starts all background hooks, then runs foreground hooks sequentially and in `job.go` with wiatgroup it waits for all background hooks to complete   
- Replaced `GetBackgroundHookResults()` with `WaitBackgroundHooks()` (Non-blocking, returned immediately) in hooks.go , now blocks until all background hooks complete, then drains results and resets channel for reuse .

pkg/burner/hooks.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+
+	case <-hm.ctx.Done():
+		log.Warnf("Hook monitor cancel for '%s'", hook.When)
+	}
+}
+
+func (hm *HookManager) executeForegroundHook(hook config.Hook) error {
+	log.Infof("Executing foreground hook: %v", hook.Cmd)
+	timeout := 5 * time.Minute
```
ygalblum: Why is this needed? What if the user wants to run a blocking operation that can take longer than that? For example, waiting for VMs to complete their boot once they are in `Ready` state may take a long time especially when provisioning a large number of VMs

pkg/burner/hooks.go
```diff
@@ -0,0 +1,207 @@
...
+	ctx, cancel := context.WithTimeout(hm.ctx, timeout)
```
ygalblum: Continuing the previous comment, why not user the `hm.ctx` to allow canceling from the caller side

pkg/burner/job.go
```diff
@@ -126,10 +134,29 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
...
 				if jobErrs := jobExecutor.RunCreateJob(ctx, 0, jobExecutor.JobIterations); jobErrs != nil {
 					errs = append(errs, jobErrs...)
 					innerRC = 1
 				}
+				if err := jobExecutor.hookManager.executeHooks(jobExecutor.Hooks, config.HookAfterDeployment); err != nil {
+					log.Errorf("Error executing hooks for %s: %v", config.HookAfterDeployment, err)
+				}
+				// Collect background hook results
+				if len(jobExecutor.Hooks) > 0 {
+					backgroundResults := jobExecutor.hookManager.WaitBackgroundHooks()
```
ygalblum: Sorry for not noticing this earlier. `WaitBackgroundHooks` is only called here which is only for `create` jobs. This means that other background hooks will not be waited on. For example, `HookBeforeDeployment` and `HookAfterDeployment` executed for other job types and `HookBeforeCleanup` for any type of job

sanjay7178:
> In addition, this change is missing documentation

I've added the docs at configuration.md  , also please have a look in this discussion https://github.com/kube-burner/kube-burner/pull/1068#discussion_r2727579846

pkg/burner/job.go
```diff
@@ -210,6 +214,8 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
 					log.Infof("BeforeCleanup out: %v, err: %v", stdOut.String(), stdErr.String())
 				}
 
+				// Collect background hook results from beforeCleanup
+				errs, innerRC = jobExecutor.CollectAndLogBackgroundHookResults(errs, innerRC)
```
ygalblum: This function is called in five different places in the same function. Can't is be called only once after all hooks have been called?

ygalblum:
@sanjay7178 Thanks for all the hard work

ygalblum:
@rsevilla87 @vishnuchalla can you please review so this task can be completed?

vishnuchalla:
@sanjay7178 mind updating this real quick please?

sanjay7178:
I've resolved the conflicts

pkg/burner/job.go
```diff
@@ -133,10 +149,20 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
 					log.Infof("Churn delete delay: %v", jobExecutor.ChurnConfig.DeleteDelay)
 					log.Infof("Churn type: %v", jobExecutor.ChurnConfig.Mode)
 				}
+				if err := jobExecutor.hookManager.executeHooks(jobExecutor.Hooks, config.HookBeforeDeployment); err != nil {
```
vishnuchalla: If these blocks are present in `if` and `else` can't they be brought out of those blocks?

vishnuchalla:
@sanjay7178 thanks for all the patience. One last update please?

sanjay7178:
After resolving the conflicts , I got to know gc resource cleanup replaced as per PR: https://github.com/kube-burner/kube-burner/pull/1185 recently , so I made the required changes for it. also I did resolved some conflicts regarding `HookBeforeDeployment` &  `HookAfterDeployment` as currently `jobExecutor.RunCreateJob()` is wrapped in `runCreateOrIncremental()` function in `job.go` .

pkg/config/types.go
```diff
@@ -22,6 +22,21 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+// JobHook type of hook command
+type JobHook string
+
+const (
+	HookBeforeDeployment JobHook = "beforeJobExecution"
```
vishnuchalla: Please change this name to `HookBeforeJobExecution` everywhere. Right now its inconsistent.

pkg/burner/hooks.go
```diff
@@ -0,0 +1,203 @@
...
+	// Execute background hooks in parallel
+	for _, hook := range hooks {
+		if hook.When != when {
+			continue
+		}
+
+		if len(hook.Cmd) == 0 {
+			log.Warnf("Empty command for hook %s, skipping it", when)
+			continue
+		}
```
rsevilla87: Remove this conditional, this is already verified in config.go

pkg/burner/hooks.go
```diff
@@ -0,0 +1,203 @@
...
+				return fmt.Errorf("failed to start background hook at '%s': %w", when, err)
+			}
+		}
+	}
+	// Execute foreground hooks sequentially
+	for _, hook := range hooks {
+		if hook.When != when {
+			continue
+		}
+		if len(hook.Cmd) == 0 {
```
rsevilla87: Same as above

pkg/config/types.go
```diff
@@ -22,6 +22,21 @@ import (
...
+	HookAfterDeployment  JobHook = "afterJobExecution"
```
rsevilla87:
```suggestion
	HookAfterJobExecution  JobHook = "afterJobExecution"
```

docs/reference/configuration.md
```diff
@@ -386,6 +387,142 @@ This will create both the Gateway and VirtualService for each iteration, with pr
...
+| `background` | Run hook in background (non-blocking)                   | Boolean  | false   |
+
+#### Supported Hook Stages
+
+The `when` field specifies at which stage the hook should execute:
+
+| Stage                    | Description                                           |
+|--------------------------|-------------------------------------------------------|
+| `beforeJobExecution`     | Before job objects are created                        |
+| `afterJobExecution`      | After job objects are created                         |
```
```suggestion
| `afterJobExecution`      | After job objects are created (before churning stage)                        |
```

PR #1073, opened 2026-01-13
add udn/cudn with l2/l3 as example

mohit-sheth:
## Type of change

- Optimization

## Summary

This PR extends the UDN density workload to support **ClusterUserDefinedNetwork (CUDN)** in addition to **UserDefinedNetwork (UDN)**, while also making **L2/L3 selection configurable at render time** via `--set` flags.

The same workload can now be reused across multiple networking scenarios without duplicating templates or maintaining separate configs.

---

## What’s changed

- Added render-time support for the following flags:
  - `ENABLE_LAYER_3` (default: `true`)
  - `CUDN` (default: `false`)
- Conditional selection between:
  - `udn_l3.yml` / `udn_l2.yml`
  - `cudn_l3.yml` / `cudn_l2.yml`
- Defaults preserve existing behavior:
  - UDN + L3 when no flags are provided
- Job naming and namespaces remain consistent and derived from the selected mode

---

## Why this change

- Enables **CUDN vs UDN** testing using a single workload definition
- Allows **L2 vs L3** selection without modifying templates
- Leverages recent changes that make `--set` variables available during template rendering
- Reduces duplication and simplifies future extensions (e.g., dual-stack, BGP, egress IP)

---

## Usage examples

```bash
# Default behavior (UDN + L3)
--set ENABLE_LAYER_3=true

# UDN + L2
--set ENABLE_LAYER_3=false

# CUDN + L3
--set CUDN=true

# CUDN + L2
--set CUDN=true --set ENABLE_LAYER_3=false

### CODE REVIEW

File: examples/workloads/udn-density/templates/cudn_l2.yml [hunk 1]
```diff
@@ -0,0 +1,10 @@
+
+apiVersion: k8s.ovn.org/v1
+kind: ClusterUserDefinedNetwork
```
jtaleric, line 3: for cudn we need to match on a label.

PR #1075, opened 2026-01-15
Optional pprofInterval

rsevilla87:
## Type of change

- New feature
- Optimization

## Description

With this change, the config parameter pprofInterval is not mandatory when using the pprof measurement, if this parameter is not set, profiling information is collected when the measurment starts and stops, the pprof files are suffixed with -start and -end respectively

This "feature" can be quite interesting for us, we can enable pprof in **every** workload, I think is better to have just a few pprof files than many. wdyt @jtaleric @afcollins ? 

- Related Issue #
- Closes #

### DIFF
--- a/.github/workflows/test-k8s.yml
+++ b/.github/workflows/test-k8s.yml
@@ -38,12 +38,18 @@ jobs:
         k8s-version: ${{ fromJson(needs.fetch-latest-kind-node-tags.outputs.tags) }}
     steps:
 
-    - name: Check out code
+    - name: Checkout PR head
       uses: actions/checkout@v6
       with:
-        fetch-depth: 1
+        # Use the PR head SHA
         ref: ${{ github.event.pull_request.head.sha }}
+        # Fetch only the most recent commit of the PR
+        fetch-depth: 1
         persist-credentials: false
+
+    - name: Fetch base branch for comparison
+      # This tells Git about origin/main (or master) without downloading the whole repo
+      run: git fetch origin ${{ github.event.pull_request.base.ref }} --depth=1
 
     - name: Download kube-burner binary
       uses: actions/download-artifact@v7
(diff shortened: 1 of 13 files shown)


vishnuchalla:
As far as I have seen pprof gives useful information when there is some load/activity going on the target. And having pprofs captured at start/end times without ensuring if there any activity going on the target might not give us any useful data.

rsevilla87:
> As far as I have seen pprof gives useful information when there is some load/activity going on the target. And having pprofs captured at start/end times without ensuring if there any activity going on the target might not give us any useful data.

that's a good point, in my opinion having dozens of pprof files doesn't help either and can overwhelm users, for further observability pprofInterval can be enabled in follow-up runs.
Anyway but these changes aren't breaking the previous functionality but adding a another wayt to run this measurement.

vishnuchalla:
> > As far as I have seen pprof gives useful information when there is some load/activity going on the target. And having pprofs captured at start/end times without ensuring if there any activity going on the target might not give us any useful data.
> 
> that's a good point, in my opinion having dozens of pprof files doesn't help either and can overwhelm users, for further observability pprofInterval can be enabled in follow-up runs. Anyway but these changes aren't breaking the previous functionality but adding a another wayt to run this measurement.

My point is not about capturing multiple pprof files, it's about making sure that we capture pprofs both w/ and w/o load to understand the difference. And capturing them just at the start and end times might be a good addition to the existing logic, but having only start/end captured when there is no `pprofInterval` specified is what a doubt to me because at both the times snaps there isn't much activity on the cluster.

rsevilla87:
> specified is what a doubt to me because at both the times snaps there isn't much activity on the cluster.

Capturing cpu profiling information at the end of the workload execution might not be 100% useful, but on the other hand, this approach can be used to identify memory leaks or excessive memory usage in a golang application more easily than capturing dozens of them.

rsevilla87:
PTAL @ygalblum

rsevilla87:
> I'm wondering, was there no check making sure that interval is not 0 (that needs to be removed now)?

no, there wasn't

PR #1086, opened 2026-01-24
Add PVC resize measurement support

### LINKED ISSUE #1082: [RFE] Add measurement support for PVC resize operations
Currently, the pvcLatency measurement does not support patch, read, or delete job types [1]. In the virt-capacity-benchmark scenario from kube-burner-ocp, one of the steps resizes PVCs, but there is no measurement available to observe or quantify the outcome of these resize operations.

[1] https://kube-burner.github.io/kube-burner/latest/measurements/#pvc-latency

A new or extended measurement that can track and report latency and success metrics for PVC resize operations.

I attempted to use vmiLatency, pvcLatency, and dataVolumeLatency, but none of them are able to capture volume resize events.

**Additional context**
N/A

rishabh998186:
This PR adds support for measuring PVC (PersistentVolumeClaim) resize operations in the pvcLatency measurement.

## Changes ##

* Added resizeLatency and resizedCapacity fields to track resize operations

* Implemented resize detection via PVC status conditions (PersistentVolumeClaimResizing, FileSystemResizePending)

* Enabled patch job type support for pvcLatency measurement

* Updated pvcTransformFunc() to preserve spec.resources and spec.storageClassName

* Added "Resized" quantile to measurement output (P50, P95, P99)

## How it works ##
When a PVC is created, the implementation stores its original capacity. It then monitors PVC status conditions for resize indicators (PersistentVolumeClaimResizing, FileSystemResizePending). When the current capacity changes from the original, it detects completion and calculates the resize latency in milliseconds.

Closes #1082

### DIFF
--- a/test/test-k8s.bats
+++ b/test/test-k8s.bats
@@ -293,3 +293,9 @@ teardown_file() {
   check_file_exists ${METRICS_FOLDER}/pprof/*.pprof
   verify_object_count namespace 0 "" kubernetes.io/metadata.name=kube-burner-pprof-collector
 }
+
+
+@test "rc timeout check" {
+  run ${KUBE_BURNER} init -c /tmp/kube-burner.yml --uuid=${UUID} --timeout=2s
+  [ "$status" -eq 2 ]
+}
--- a/.github/workflows/test-k8s-ppc64le.yml
+++ b/.github/workflows/test-k8s-ppc64le.yml
@@ -26,7 +26,7 @@ jobs:
         cache: false
 
     - name: Install bats
-      uses: bats-core/bats-action@3.0.1
+      uses: bats-core/bats-action@4.0.0
       with:
         bats-version: 1.11.0
         support-install: false
--- a/.github/workflows/test-k8s.yml
+++ b/.github/workflows/test-k8s.yml
@@ -58,7 +58,7 @@ jobs:
         path: /tmp/
 
     - name: Install bats
-      uses: bats-core/bats-action@3.0.1
+      uses: bats-core/bats-action@4.0.0
       with:
         bats-version: 1.10.0
         support-install: false
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -1,0 +1,2 @@
+# External GitHub URLs that get rate-limited (429) during CI checks
+https://github.com/kubevirt/containerized-data-importer/blob/main/doc/datavolumes.md#conditions
--- a/docs/contributing/ai-policy.md
+++ b/docs/contributing/ai-policy.md
@@ -1,0 +1,69 @@
+![kube-burner logo](../media/logos/kube-burner-logo-mini.png)
+
+# kube-burner LLM (AI) Contribution Policy
+
+Large language models (LLMs) such as ChatGPT, Copilot, and Claude can help contributors work faster. They can also introduce risk when output is submitted without careful review.
+
+Kube-burner values clear communication, maintainable code, and high review quality. This policy defines expectations when AI tools are used in kube-burner repositories and community spaces.
+
+## Direct Communication
+
+Do not post raw or copy-pasted LLM output as your own communication.
+
+This applies to:
+
+- GitHub issues and issue comments
+- Pull request descriptions and review comments
+- Commit messages
+- Slack, discussions, and other project communication channels
+- Security and vulnerability reports
+
+Write in your own words and make sure you understand what you submit.
+
+### Exceptions
+
+- Translation help is allowed. If you used an LLM for translation, say so clearly (for example: "Translated with an LLM from <language>").
+- Maintainer-managed automation or bots may suggest changes. These suggestions are not authoritative and must still be reviewed by humans.
+
+Repeated violations may result in closed submissions or removal from project spaces.
+
+## Code Contributions Assisted by LLMs
+
+Using LLMs for coding is allowed, but responsibility remains with the contributor.
+
+### Requirements
+
+- Follow all project contribution guidance in [CONTRIBUTING.md](../../CONTRIBUTING.md)
+- Keep changes focused, minimal, and easy to review
+- Match existing style, formatting, and structure
+- Remove unnecessary noise (unused files, generated metadata, unrelated edits)
+- Ensure your change builds and passes relevant tests before requesting review
+- Test the behavior you changed
+
+### Ownership and Understanding
+
+You must:
+
+- Review generated code before submitting
+- Explain what changed and why in your own words
+- Be able to discuss implementation details during review
+
+Submissions that are clearly not understood by the author may be rejected.
+
+## Responding to Review Feedback
+
+Do not blindly paste review comments into an LLM and submit the response unchanged.
+
+Contributors are expected to:
+
+- Respond thoughtfully in their own words
+- Make targeted updates that address the feedback
+- Understand and validate each requested change
+
+## Maintainer Discretion
+
+Maintainers may decline changes that are difficult to review, overly large, poorly structured, or repeatedly low quality, regardless of whether AI tools were used.
+
+## Golden Rule
+
+Use LLMs as assistants, not as substitutes for understanding, accountability, and craftsmanship.
--- a/docs/contributing/index.md
+++ b/docs/contributing/index.md
@@ -3,6 +3,10 @@
 If you want to contribute to kube-burner, you can do so by submitting a Pull Request, Issue or starting a Discussion.
 You can also reach us out in the `#kube-burner` channel of the [kubernetes slack](https://kubernetes.slack.com/messages/kube-burner).
 
+## Policies
+
+- [LLM (AI) Contribution Policy](ai-policy.md)
+
 ## CI and Linting
 
 For running pre-commit checks on your code before committing code and opening a PR, you can use the `pre-commit run` functionality.  See [CI docs](https://kube-burner.github.io/kube-burner/latest/contributing/pullrequest/#running-local-pre-commit) for more information on running pre-commits.
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -145,7 +145,7 @@ This section contains the list of jobs `kube-burner` will execute. Each job can 
 | `defaultMissingKeysWithZero` | Stops templates from exiting with an error when a missing key is found, meaning users will have to ensure templates hand missing keys | Boolean  | false    |
 | `executionMode`              | Job execution mode. More details at [execution modes](#execution-modes)                                                               | String   | parallel |
 | `objectDelay`                | How long to wait between each object in a job                                                                                         | Duration | 0s       |
-| `objectWait`                 | Wait for each object to complete before processing the next one - not for Create jobs                                                 | Boolean  | 0s       |
+| `objectWait`                 | Wait for each object to complete before processing the next one - not for Create jobs                                                 | Boolean  | false    |
 | `metricsAggregate`           | Aggregate the metrics collected for this job with those of the next one                                                               | Boolean  | false    |
 | `metricsClosing`             | To define when the metrics collection should stop. More details at [MetricsClosing](#MetricsClosing)                                  | String   | afterJobPause |
 
@@ -597,6 +597,7 @@ Churn supports the following options:
 - `percent`: Percentage of the jobIterations to churn each period
 - `duration`: Length of time that the job is churned for
 - `delay`: Length of time to wait between each churn period
+- `deleteDelay`: Length of time to wait after deletion and before recreation within a churn period. Defaults to `0s`
 - `mode`: Churning mode, either `namespaces`, to churn entire namespaces or `objects`, to churn individual objects of the job's namespaces. Defaults to `namespaces`.
 
 !!! note
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -13,6 +13,7 @@ nav:
   - Alerting: observability/alerting.md
 - Contributing:
   - contributing/index.md
+  - AI Policy: contributing/ai-policy.md
   - GitHub Workflows:
     - contributing/pullrequest.md
     - contributing/release.md
--- a/pkg/burner/create.go
+++ b/pkg/burner/create.go
@@ -452,6 +452,10 @@ func (ex *JobExecutor) churnNamespaces(ctx context.Context) []error {
 		// 1 hour timeout to delete namespace
 		cleanupCtx, cancel := context.WithTimeout(context.Background(), time.Hour)
 		util.CleanupNamespacesByLabel(cleanupCtx, ex.clientSet, config.KubeBurnerLabelChurnDelete)
+		if ex.ChurnConfig.DeleteDelay > 0 {
+			log.Infof("Sleeping for %v after deletion", ex.ChurnConfig.DeleteDelay)
+			time.Sleep(ex.ChurnConfig.DeleteDelay)
+		}
 		// Re-create objects that were deleted
 		log.Infof("Re-creating %d deleted namespaces", numToChurn)
 		if jobErrs := ex.RunCreateJob(cleanupCtx, randStart, numToChurn+randStart); jobErrs != nil {
@@ -522,6 +526,10 @@ func (ex *JobExecutor) churnObjects(ctx context.Context) {
 			}
 		}
 		ex.verifyDelete(ctx, deletedObjects)
+		if ex.ChurnConfig.DeleteDelay > 0 {
+			log.Infof("Sleeping for %v after deletion", ex.ChurnConfig.DeleteDelay)
+			time.Sleep(ex.ChurnConfig.DeleteDelay)
+		}
 		ex.reCreateDeletedObjects(ctx, deletedObjects)
 		log.Infof("Sleeping for %v", ex.ChurnConfig.Delay)
 		time.Sleep(ex.ChurnConfig.Delay)
--- a/pkg/burner/job.go
+++ b/pkg/burner/job.go
@@ -91,9 +91,19 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider, 
 		_, restConfig := kubeClientProvider.DefaultClientSet()
 		measurementsFactory := measurements.NewMeasurementsFactory(configSpec, metricsScraper.MetricsMetadata, additionalMeasurementFactoryMap)
 		jobExecutors = newExecutorList(configSpec, kubeClientProvider, embedCfg)
-		handlePreloadImages(ctx, jobExecutors, kubeClientProvider)
-		// Iterate job list
+		if err := handlePreloadImages(ctx, jobExecutors, kubeClientProvider); err != nil {
+			log.Error(err.Error())
+			return
+		}
 
+		// Cleanup previous runs
+		for _, jobExecutor := range jobExecutors {
+			if jobExecutor.JobType == config.CreationJob && jobExecutor.Cleanup {
+				log.Infof("Cleaning up previous runs for job: %s", jobExecutor.Name)
+				jobExecutor.gc(ctx, nil)
+			}
+		}
+		// Run jobs
 		for jobExecutorIdx, jobExecutor := range jobExecutors {
 			executedJobs = append(executedJobs, prometheus.Job{
 				Start:     time.Now().UTC(),
@@ -114,16 +124,13 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider, 
 			}
 			log.Infof("Triggering job: %s", jobExecutor.Name)
 			if jobExecutor.JobType == config.CreationJob {
-				if jobExecutor.Cleanup {
-					log.Info("Cleaning up previous runs")
-					jobExecutor.gc(ctx, nil)
-				}
 				if config.IsChurnEnabled(jobExecutor.Job) {
 					log.Info("Churning enabled")
 					log.Infof("Churn cycles: %v", jobExecutor.ChurnConfig.Cycles)
 					log.Infof("Churn duration: %v", jobExecutor.ChurnConfig.Duration)
 					log.Infof("Churn percent: %v", jobExecutor.ChurnConfig.Percent)
 					log.Infof("Churn delay: %v", jobExecutor.ChurnConfig.Delay)
+					log.Infof("Churn delete delay: %v", jobExecutor.ChurnConfig.DeleteDelay)
 					log.Infof("Churn type: %v", jobExecutor.ChurnConfig.Mode)
 				}
 				if jobErrs := jobExecutor.RunCreateJob(ctx, 0, jobExecutor.JobIterations); jobErrs != nil {
@@ -266,7 +273,9 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider, 
 	case <-time.After(configSpec.GlobalConfig.Timeout):
 		err := fmt.Errorf("%v timeout reached", configSpec.GlobalConfig.Timeout)
 		log.Error(err.Error())
-		executedJobs[len(executedJobs)-1].End = time.Now().UTC()
+		if len(executedJobs) > 0 {
+			executedJobs[len(executedJobs)-1].End = time.Now().UTC()
+		}
 		errs = append(errs, err)
 		rc = rcTimeout
 		if measurementsInstance != nil {
@@ -299,15 +308,16 @@ func Destroy(ctx context.Context, configSpec config.Spec, kubeClientProvider *co
 }
 
 // If requests, preload the images used in the test into the node
-func handlePreloadImages(ctx context.Context, executorList []JobExecutor, kubeClientProvider *config.KubeClientProvider) {
+func handlePreloadImages(ctx context.Context, executorList []JobExecutor, kubeClientProvider *config.KubeClientProvider) error {
 	clientSet, _ := kubeClientProvider.DefaultClientSet()
 	for _, executor := range executorList {
 		if executor.PreLoadImages && executor.JobType == config.CreationJob {
 			if err := preLoadImages(ctx, executor, clientSet); err != nil {
-				log.Fatal(err.Error())
+				return err
 			}
 		}
 	}
+	return nil
 }
 
 // indexMetrics indexes metrics for the executed jobs
--- a/pkg/burner/pre_load.go
+++ b/pkg/burner/pre_load.go
@@ -22,8 +22,6 @@ import (
 	"strings"
 	"time"
 
-	"maps"
-
 	"github.com/kube-burner/kube-burner/v2/pkg/util"
 	log "github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v3"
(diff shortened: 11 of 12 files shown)


rishabh998186:
hey  @rsevilla87  ,
 could you please review this pr  when you get a chance?

sanjay7178:
> hey @sanjay7178 @rsevilla87 , could you please review this pr when you get a chance?

Sorry i'm not a maintainer of this project btw :-)

pkg/measurements/pvc_latency.go
```diff
@@ -102,18 +108,22 @@ func (p *pvcLatency) handleCreatePVC(obj any) {
 	}
 	log.Tracef("handleCreatePVC: %s", pvc.Name)
 	pvcLabels := pvc.GetLabels()
+
+	originalCapacity := pvc.Spec.Resources.Requests.Storage().String()
```
rsevilla87: not necessary, this is already present in Size

pkg/measurements/pvc_latency.go
```diff
@@ -51,23 +52,28 @@ var (
...
+	pending          int64
+	PendingLatency   int `json:"pendingLatency"`
+	bound            int64
+	BindingLatency   int `json:"bindingLatency"`
+	lost             int64
+	LostLatency      int `json:"lostLatency"`
+	resizeStarted    int64
+	ResizeLatency    int `json:"resizeLatency"`
+	originalCapacity string
+	resizedCapacity  string
```
rsevilla87: Remove

pkg/measurements/pvc_latency.go
```diff
@@ -152,13 +162,44 @@ func (p *pvcLatency) handleUpdatePVC(obj any) {
...
+					pm.resizeStarted = time.Now().UTC().UnixMilli()
+					log.Debugf("PVC %s is resizing", pvc.Name)
+				}
+			}
+		}
+
+		// Check if resize completed by comparing capacity
+		if pvc.Status.Capacity != nil {
+			currentCapacity := pvc.Status.Capacity.Storage().String()
+			if pm.originalCapacity != "" && currentCapacity != pm.originalCapacity {
```
rsevilla87: What if the resizing operation is not finished?

pkg/measurements/pvc_latency.go
```diff
@@ -152,13 +162,44 @@ func (p *pvcLatency) handleUpdatePVC(obj any) {
 		} else {
 			log.Tracef("Skipping update for phase [%s] as PVC is already bound or lost", pvc.Status.Phase)
 		}
+		// Track resize operations via conditions
+		for _, condition := range pvc.Status.Conditions {
+			if condition.Type == corev1.PersistentVolumeClaimFileSystemResizePending {
```
ygalblum: Why do both conditions update the same `resizeStarted` field?

ygalblum:
> This approach won't work over already present volumes in the cluster, there're can be scenarios where the user just want to resize PVCs that were already present in a cluster.
> 
> For that, I think is better to create a new measurement pvc_resize_latency rather than updating this one. wdyt @ygalblum ?

I don't see an issue here. All measurements are limited to resources created by the test

rsevilla87:
> > This approach won't work over already present volumes in the cluster, there're can be scenarios where the user just want to resize PVCs that were already present in a cluster.
> > For that, I think is better to create a new measurement pvc_resize_latency rather than updating this one. wdyt @ygalblum ?
> 
> I don't see an issue here. All measurements are limited to resources created by the test

Do you think this would work with virt-capacity?, I think not, the resize operations happen in a different job.

pkg/measurements/pvc_latency.go
rishabh998186: they represent different stages of the same resize operation.  either condition can appear first, so we check both to capture when resize actually started. The `pm.resizeStarted == 0` guard ensures we only record the first timestamp

pkg/measurements/pvc_latency.go
rishabh998186: Currently if resize doesn't complete, the metric won't be recorded (ResizeLatency stays 0).

rishabh998186: Looking at other latency measurements  they all skip incomplete operations and log a warning. I follow the same pattern here if a resize doesn't complete `(resizeStarted > 0 but ResizeLatency == 0)`, we skip recording it and log a warning. This keeps metrics clean and maintains consistency with how kube-burner handles incomplete operations.

ygalblum:
> Do you think this would work with virt-capacity?, I think not, the resize operations happen in a different job.

I guess it's better to check. But, the listener filter is based on the `RUNID` which is the same throughout the run, not specific to a job. So, it should work. No?

rishabh998186:
@rsevilla87 @ygalblum

Could you please review my new changes? i tested the implementation with virt-capacity and its working perfectly now.
 i also activate the new pvcLatency measurement specifically for the resize-volumes job in the virt-capacity-benchmark .
``` 
measurements:
- name: pvcLatency
 ```
This tells kube-burner to track PVC lifecycle + resize timing during VM disk expansion, giving us the missing storage performance data.
If you like this change, I can create a PR on the kube-burner-ocp repo to merge these changes

File: pkg/measurements/pvc_latency_test.go
Comment (rsevilla87) [inline pkg/measurements/pvc_latency_test.go:1]: We don't run unit tests for now, please remove

File: pkg/measurements/pvc_latency.go [hunk 5]
```diff
@@ -152,19 +159,58 @@ func (p *pvcLatency) handleUpdatePVC(obj any) {
...
+
+		p.Metrics.Store(string(pvc.UID), pm)
 	}
 }
 
 // start pvcLatency measurement
 func (p *pvcLatency) Start(measurementWg *sync.WaitGroup) error {
 	defer measurementWg.Done()
-	if p.JobConfig.JobType == config.ReadJob || p.JobConfig.JobType == config.PatchJob || p.JobConfig.JobType == config.DeletionJob {
+	if p.JobConfig.JobType == config.ReadJob {
```
rsevilla87: This conditional is redundant with https://github.com/rishabh998186/kube-burner/blob/a25a8fcf93491bea09bba9f6e5db3f8d4eb21b73/pkg/measurements/pvc_latency.go#L324-L327, remove please

pkg/measurements/pvc_latency.go
```diff
@@ -152,19 +159,58 @@ func (p *pvcLatency) handleUpdatePVC(obj any) {
...
 	}
 	gvr, err := util.ResourceToGVR(p.RestConfig, "PersistentVolumeClaim", "v1")
 	if err != nil {
 		return fmt.Errorf("error getting GVR for %s: %w", "PersistentVolumeClaim", err)
 	}
+
+	// Preload existing PVCs so patch jobs (resize-volumes) have metrics state
+	pvcs, err := p.ClientSet.CoreV1().PersistentVolumeClaims("").List(
+		context.TODO(),
+		metav1.ListOptions{LabelSelector: p.LabelSelector},
```
rsevilla87: I think you shouldn't use any labelSelector to list the current PVCs, that way, if an already present PVC in the cluster (not only the ones that have been created by a kube-burner within the same benchmark), can be tracked by this measurement

pkg/measurements/pvc_latency.go
```diff
@@ -152,19 +159,58 @@ func (p *pvcLatency) handleUpdatePVC(obj any) {
 		} else {
 			log.Tracef("Skipping update for phase [%s] as PVC is already bound or lost", pvc.Status.Phase)
 		}
+
+		// Detect resize start by spec change (works for all provisioners)
+		if pvc.Spec.Resources.Requests.Storage() != nil {
+			requestedSize := pvc.Spec.Resources.Requests.Storage().String()
+			if requestedSize != pm.Size && pm.resizeStarted == 0 {
+				pm.resizeStarted = time.Now().UTC().UnixMilli()
+				log.Debugf("PVC %s resize started: %s -> %s", pvc.Name, pm.Size, requestedSize)
```
rsevilla87:
```suggestion
				log.Debugf("PVC %s at %s resize started: %s -> %s", pvc.Name, pm.Namespace, pm.Size, requestedSize)
```

pkg/measurements/pvc_latency.go
rsevilla87: Remove the labelSelector, we need to track it if the volume has been created outside of kube-burner or in a previous execution with different RUNID

pkg/measurements/pvc_latency.go
```diff
@@ -128,7 +135,7 @@ func (p *pvcLatency) handleUpdatePVC(obj any) {
 	if value, exists := p.Metrics.Load(string(pvc.UID)); exists {
 		pm := value.(pvcMetric)
 		log.Tracef("handleUpdatePVC: PVC: [%s], Version: [%s], Phase: [%s]", pvc.Name, pvc.ResourceVersion, pvc.Status.Phase)
-		if pm.bound == 0 || pm.lost == 0 {
+		if pm.bound == 0 && pm.lost == 0 {
```
rsevilla87: I've noticed that if an already present volume volume is resized, the bound latency is filled. this can be easily be tested with `kube-burner measure`

pkg/measurements/pvc_latency.go
```diff
@@ -152,19 +159,58 @@ func (p *pvcLatency) handleUpdatePVC(obj any) {
 		} else {
 			log.Tracef("Skipping update for phase [%s] as PVC is already bound or lost", pvc.Status.Phase)
 		}
+
+		// Detect resize start by spec change (works for all provisioners)
+		if pvc.Spec.Resources.Requests.Storage() != nil {
```
rsevilla87: This conditional is not necessary, you can't have a PVC with nil storage

pkg/measurements/pvc_latency.go
rsevilla87: I think this can be fixed by nesting this logic in the an else conditional or when `pvc.Spec.Resources.Requests.Storage() == pm.Size`

pkg/measurements/pvc_latency.go
```diff
@@ -102,6 +108,7 @@ func (p *pvcLatency) handleCreatePVC(obj any) {
 	}
```
rsevilla87, line 113: When using the measurement from  `kube-burner measure`, the already present volumes are listed and passed to this function directly as PersistentVolumeClaim objects.

```suggestion
func (p *pvcLatency) handleCreatePVC(obj any) {
	var err error
	// Verify that object type is pvc
	pvc, ok := obj.(*corev1.PersistentVolumeClaim)
	if !ok {
		pvc, err = util.ConvertAnyToTyped[corev1.PersistentVolumeClaim](obj)
		if err != nil {
			log.Errorf("failed to convert to PersistentVolumeClaim: %v", err)
			return
		}
	}
```

pkg/measurements/pvc_latency.go
```diff
@@ -207,9 +253,19 @@ func (p *pvcLatency) Stop() error {
...
 	p.Metrics.Range(func(key, value any) bool {
 		m := value.(pvcMetric)
+
+		// Final check for resize completion at the end of the job
+		if m.resizeStarted > 0 && m.ResizeLatency == 0 {
+			m.ResizeLatency = int(stopTimestamp - m.resizeStarted)
+			if m.ResizedCapacity == "" {
+				m.ResizedCapacity = "unknown (calculated at stop)"
+			}
+			log.Infof("PVC %v resize latency calculated at job end: %dms", m.Name, m.ResizeLatency)
```
rsevilla87: Remove this log message

pkg/measurements/pvc_latency.go
```diff
@@ -255,6 +317,7 @@ func (p *pvcLatency) getLatency(normLatency any) map[string]float64 {
 		string(corev1.ClaimPending): float64(pvcMetric.PendingLatency),
 		string(corev1.ClaimBound):   float64(pvcMetric.BindingLatency),
 		string(corev1.ClaimLost):    float64(pvcMetric.LostLatency),
+		"Resized":                   float64(pvcMetric.ResizeLatency),
```
rsevilla87:
```suggestion
		"Resize":                   float64(pvcMetric.ResizeLatency),
```

pkg/measurements/pvc_latency.go
```diff
@@ -128,7 +140,28 @@ func (p *pvcLatency) handleUpdatePVC(obj any) {
...
+		// Detect resize start by spec change (works for all provisioners)
+		requestedSize := pvc.Spec.Resources.Requests.Storage().String()
+		if requestedSize != pm.Size && pm.resizeStarted == 0 {
+			pm.resizeStarted = time.Now().UTC().UnixMilli()
+			log.Debugf("PVC %s resize started: %s -> %s", pvc.Name, pm.Size, requestedSize)
+		}
+
+		// Check if resize completed by comparing capacity
+		if pm.resizeStarted > 0 && pm.ResizeLatency == 0 {
+			if pvc.Status.Capacity != nil {
```
rsevilla87: This conditional is not necessary, capacity will always have a value

pkg/measurements/pvc_latency.go
```diff
@@ -207,9 +251,18 @@ func (p *pvcLatency) Stop() error {
 func (p *pvcLatency) normalizeMetrics() float64 {
```
rsevilla87, line 252: There're multiple errors in this function, that prevent the metrics to get generated correctly, for example, ResizeLatency is calculated again when the measurement is stopped using the stopTimestamp, that's incorrect.  I've made some modifcations locally using the following code, and now looks it works ok

```golang
func (p *pvcLatency) normalizeMetrics() float64 {
	totalPVCs := 0
	erroredPVCs := 0

	p.Metrics.Range(func(key, value any) bool {
		m := value.(pvcMetric)
		if m.resizeStarted == 0 && (m.bound == 0 && m.lost == 0) {
			log.Tracef("PVC %v latency ignored as it did not reach a stable state", m.Name)
			return true
		}
		errorFlag := 0
		m.PendingLatency = int(m.pending - m.Timestamp.UnixMilli())
		if m.PendingLatency < 0 {
			log.Tracef("PendingLatency for pvc %v falling under negative case. So explicitly setting it to 0", m.Name)
			if m.pending < 0 {
				errorFlag = 1
			}
			m.PendingLatency = 0
		}
		if m.bound > 0 {
			m.BindingLatency = int(m.bound - m.Timestamp.UnixMilli())
			if m.BindingLatency < 0 {
				log.Tracef("BindingLatency for pvc %v falling under negative case. So explicitly setting it to 0", m.Name)
				errorFlag = 1
				m.BindingLatency = 0
			}
		}
		if m.lost > 0 {
			m.LostLatency = int(m.lost - m.Timestamp.UnixMilli())
			if m.LostLatency < 0 {
				log.Tracef("LostLatency for pvc %v falling under negative case. So explicitly setting it to 0", m.Name)
				m.LostLatency = 0
			}
		}

		// ResizeLatency is already calculated in handleUpdatePVC or above, just validate
		if m.ResizeLatency < 0 {
			log.Tracef("ResizeLatency for pvc %v falling under negative case. So explicitly setting it to 0", m.Name)
			m.ResizeLatency = 0
		}

		totalPVCs++
		erroredPVCs += errorFlag
		p.NormLatencies = append(p.NormLatencies, m)
		return true
	})
	if totalPVCs == 0 {
		return 0.0
	}
	return float64(erroredPVCs) / float64(totalPVCs) * 100.0
}
```

rishabh998186:
Hey @rsevilla87,

Thanks for the detailed feedback and the suggested implementation for `normalizemetrics`. i have addressed all your changes. does its look good now?

pkg/measurements/pvc_latency.go
```diff
@@ -128,7 +140,27 @@ func (p *pvcLatency) handleUpdatePVC(obj any) {
...
+		if requestedSize != pm.Size && pm.resizeStarted == 0 {
+			pm.resizeStarted = time.Now().UTC().UnixMilli()
+			log.Debugf("PVC %s resize started: %s -> %s", pvc.Name, pm.Size, requestedSize)
+		}
+
+		// Check if resize completed by comparing capacity
+		if pm.resizeStarted > 0 && pm.ResizeLatency == 0 {
+			currentCapacity := pvc.Status.Capacity.Storage().String()
+			// If capacity has changed from original size, it's done
+			if currentCapacity != pm.Size {
```
rsevilla87: We need to make sure that the current capacity is the same as the requests.storage

```suggestion
			if currentCapacity == pvc.Spec.Resources.Requests.Storage().String()
```

rsevilla87:
> Hey @rsevilla87,
> 
> Thanks for the detailed feedback and the suggested implementation for `normalizemetrics`. i have addressed all your changes. does its look good now?

Almost :), just a last nit

pkg/measurements/pvc_latency.go
```diff
@@ -128,7 +140,26 @@ func (p *pvcLatency) handleUpdatePVC(obj any) {
...
+		if pm.resizeStarted > 0 && pm.ResizeLatency == 0 {
+			currentCapacity := pvc.Status.Capacity.Storage().String()
+			// If capacity has changed from original size or matches requested, it's done
+			if currentCapacity == pvc.Spec.Resources.Requests.Storage().String() {
+				pm.ResizeLatency = int(time.Now().UTC().UnixMilli() - pm.resizeStarted)
+				pm.ResizedCapacity = currentCapacity
+				log.Debugf("PVC %s resize completed (capacity update): %s -> %s in %dms",
+					pvc.Name, pm.Size, currentCapacity, pm.ResizeLatency)
+			}
+		} else if pm.bound == 0 && pm.lost == 0 && requestedSize == pm.Size {
```
ygalblum: why did `pm.bound == 0 || pm.lost == 0` (meaning one of them) change to `pm.bound == 0 && pm.lost == 0` meaning both of them?

pkg/measurements/pvc_latency.go
```diff
@@ -148,23 +179,33 @@ func (p *pvcLatency) handleUpdatePVC(obj any) {
...
 func (p *pvcLatency) Start(measurementWg *sync.WaitGroup) error {
 	defer measurementWg.Done()
-	if p.JobConfig.JobType == config.ReadJob || p.JobConfig.JobType == config.PatchJob || p.JobConfig.JobType == config.DeletionJob {
-		log.Fatalf("Unsupported jobType:%s for pvcLatency metric", p.JobConfig.JobType)
-	}
 	gvr, err := util.ResourceToGVR(p.RestConfig, "PersistentVolumeClaim", "v1")
 	if err != nil {
 		return fmt.Errorf("error getting GVR for %s: %w", "PersistentVolumeClaim", err)
 	}
+	pvcs, err := p.ClientSet.CoreV1().PersistentVolumeClaims("").List(
```
ygalblum, line 202: This doesn't make sense to me. First, you'll get all PVC in the system. This means that many PVCs will not have any updated information once the measurement is normalized. Second, I'm still not even convinced that it in fact needed. Have you tried without it? Does it miss notifications? Does it get notifications but cannot find a match in the set?

rsevilla87: It works @ygalblum , the idea is to track volumes that are already present on the cluster

Nesar976:
Thanks for the clarification.

Understood — the intent here is to keep `pvcLatency` limited to resources created by the test run. My change only fixes missing spec fields (storageClass, size) during PVC updates and doesn’t alter the measurement scope or behavior.

If there’s consensus to split resize tracking into a separate measurement (e.g. `pvc_resize_latency`) going forward, I’m fine keeping this PR strictly as a bug fix for existing PVC latency metrics.

pkg/measurements/pvc_latency.go
rishabh998186: With`` || ``it keeps watching as long as at least one of these events hasn't happened. For example, after the PVC is Bound, it would keep processing updates forever because ```pm.lost``` is still 0. However, with &&, it only watch if the PVC has not yet ``Bound`` or  ``Lost`` . This means that after the first state is reached, it stops completely and locks in the original timestamp."

pkg/measurements/pvc_latency.go
rishabh998186: A standard Kubernetes Watcher only notifies you of New events (Add , Change ,Update). If a PVC was created before  in a previous job run, the watcher will never see an "Add" event for it.

pkg/measurements/pvc_latency.go
rsevilla87: good point, sgtm

ygalblum: You already lock the values because you kept the check inside each case. But, what happens if it's lost after it was bound? This code will miss it.

pkg/measurements/pvc_latency.go
```diff
@@ -210,8 +251,7 @@ func (p *pvcLatency) normalizeMetrics() float64 {
 
 	p.Metrics.Range(func(key, value any) bool {
 		m := value.(pvcMetric)
-		// If a pvc does not reach the stable state, we skip that one
-		if m.bound == 0 && m.lost == 0 {
+		if m.resizeStarted == 0 && (m.bound == 0 && m.lost == 0) {
```
ygalblum:
```suggestion
		if m.resizeStarted == 0 && m.bound == 0 && m.lost == 0 {
```

pkg/measurements/pvc_latency.go
rishabh998186:
```
// handles pvc update
func (p *pvcLatency) handleUpdatePVC(obj any) {
	pvc, err := util.ConvertAnyToTyped[corev1.PersistentVolumeClaim](obj)
	if err != nil {
		log.Errorf("failed to convert to PersistentVolumeClaim: %v", err)
		return
	}
	log.Tracef("handleUpdatePVC: %s", pvc.Name)

	value, exists := p.Metrics.Load(string(pvc.UID))
	if !exists {
		return
	}

	pm := value.(pvcMetric)
	log.Tracef("handleUpdatePVC: PVC: [%s], Version: [%s], Phase: [%s]", pvc.Name, pvc.ResourceVersion, pvc.Status.Phase)
	now := time.Now().UTC().UnixMilli()

	// 1. Phase Tracking (capture timestamps of first entry into each state)
	switch pvc.Status.Phase {
	case corev1.ClaimPending:
		if pm.pending == 0 {
			pm.pending = now
			log.Debugf("PVC %s is pending", pvc.Name)
		}
	case corev1.ClaimBound:
		if pm.bound == 0 {
			pm.bound = now
			log.Debugf("PVC %s is bound", pvc.Name)
		}
	case corev1.ClaimLost:
		if pm.lost == 0 {
			pm.lost = now
			log.Debugf("PVC %s is lost", pvc.Name)
		}
	}

	// 2. Resize Detection
	requestedSize := pvc.Spec.Resources.Requests.Storage().String()

	// Detect resize start by spec change (works for all provisioners)
	if requestedSize != pm.Size && pm.resizeStarted == 0 {
		pm.resizeStarted = now
		log.Debugf("PVC %s resize started: %s -> %s", pvc.Name, pm.Size, requestedSize)
	}

	// Check if resize completed by comparing capacity
	if pm.resizeStarted > 0 && pm.ResizeLatency == 0 {
		currentCapacity := pvc.Status.Capacity.Storage().String()
		// If capacity matches requested, it's done
		if currentCapacity == requestedSize {
			pm.ResizeLatency = int(now - pm.resizeStarted)
			pm.ResizedCapacity = currentCapacity
			log.Debugf("PVC %s resize completed (capacity update): %s -> %s in %dms",
				pvc.Name, pm.Size, currentCapacity, pm.ResizeLatency)
		}
	}

	p.Metrics.Store(string(pvc.UID), pm)
}
```
 removing that outer lock and replacing it with a Switch statement that runs every single time the PVC status changes. Now, the code evaluates the PVC's current state independently during every event. Even if the PVC has already been recorded as `Bound`, the switch statement still checks the status. If the status flips to Lost, the code moves into the `ClaimLost` case. Because each phase now has its own "inner latch" (e.g.,` if pm.lost == 0`), it checks: "Have I seen the Lost state for this specific PVC before?" Since it's the first time it's seeing it, it records the timestamp.

This means we now capture the first occurrence of every phase a PVC goes through, rather than just the very first phase it reaches. The result is a timeline that can accurately show a PVC moving from `Pending -> Bound -> Lost`, instead of getting stuck forever at Bound.
maybe this can resolve your concern, what do you think ?

ygalblum: I think it does, yes

ygalblum (changes requested):
This PR is still missing tests

test/k8s/kube-burner-pvc-resize.yml
```diff
@@ -0,0 +1,43 @@
...
+    namespace: pvc-resize-test
+    executionMode: parallel
+    jobPause: 60s
+    objects:
+      - kind: PersistentVolumeClaim
+        labelSelector: {kube-burner-job: pvc-resize}
+        patchType: "application/merge-patch+json"
+        objectTemplate: objectTemplates/pvc-patch.json
+        inputVars:
+          pvcSize: "4Gi"
```
ygalblum: Use the smallest possible value to avoid hitting storage issues
```suggestion
          pvcSize: "2Gi"
```

test/test-k8s.bats
```diff
@@ -230,6 +230,13 @@ teardown_file() {
   check_quantile_recorded create-snapshot volumeSnapshotLatency Ready
 }
 
+# bats test_tags=subsystem:measurements
+@test "kube-burner-pvc-resize.yml: pvc resize latency" {
+  run_cmd ${KUBE_BURNER} init -c test/k8s/kube-burner-pvc-resize.yml --uuid=${UUID} --log-level=debug
```
ygalblum: The bats script already assumes it's inside `test/k8s`. This is why the CI is failing
```suggestion
  run_cmd ${KUBE_BURNER} init -c kube-burner-pvc-resize.yml --uuid=${UUID} --log-level=debug
```

test/k8s/kube-burner-pvc-resize.yml
ygalblum: This was marked as resolved but the suggestion was not merged with no explanation

danish9039:
dummy

test/k8s/kube-burner-pvc-resize.yml
```diff
@@ -0,0 +1,43 @@
...
+    jobIterations: 1
+    qps: 10
+    burst: 10
+    namespacedIterations: false
+    namespace: pvc-resize-test
+    waitWhenFinished: true # Ensure bound before next job
+    podWait: false
+    maxWaitTimeout: 5m # Increased timeout
+    objects:
+      - objectTemplate: objectTemplates/pvc.yml
```
ygalblum, line 23: This file doesn't seem to exist. This is why the CI is failing

rishabh998186:
hey @ygalblum  @rsevilla87 ,
Could you please review the updated test files when you have a chance?

pkg/measurements/pvc_latency.go
```diff
@@ -42,6 +44,7 @@ var (
 		string(corev1.ClaimPending): {},
 		string(corev1.ClaimBound):   {},
 		string(corev1.ClaimLost):    {},
+		"Resized":                   {},
```
rsevilla87, line 47:
```suggestion
		"Resize":                   {},
```

pkg/measurements/pvc_latency.go
```diff
@@ -42,6 +44,7 @@ var (
 		string(corev1.ClaimPending): {},
 		string(corev1.ClaimBound):   {},
 		string(corev1.ClaimLost):    {},
+		"Resize":                   {},
```
rsevilla87:
```suggestion
		"Resize":                    {},
```

rsevilla87:
Last check here @ygalblum ?

rsevilla87:
> Copilot has some valid comments (especially in `pvc-patch.json`). Also, the CI is still red

CI failures are unrelated to this change,

vishnuchalla:
looks good to me, except for DCO and Co-pilot's comments on reusable parameters.

rishabh998186:
Hi @ygalblum, @rsevilla87, and @vishnuchalla, I addressed the changes suggested by Copilot also all checks have succesfully passed. please let me know if you want me to change anything else

Issue #1059, opened 2025-12-16
[RFE] Incremental load support


rsevilla87:
**Is your feature request related to a problem? Please describe.**
TBD

**Describe the solution you'd like**
TBD

**Describe alternatives you've considered**
TBD

**Additional context**
TBD

jtaleric:
I would like to see us have some user defined parameters to increase load over-time. 

Here is what I am thinking... 
`--min-iterations`  - What is the minimum number of iterations (basically the start of the test)
`--step-size` - How much we want to increment the iterations to
`--max-iterations` - Where do we stop. We could have 0 be keep going until health-check fails.

After each "step" we run the health-check to determine if the cluster is healthy, I think having some mechanism to poll the cluster for 5m (for example) to ensure nothing is flapping -- versus how we do healthy check now as a snapshot in time.

venkataanil:
In case of openshift "oc get co" will show the stability of the cluster. Its better we have user defined code to execute for health check. In this way, user can pass "oc get co" equivalent code if they are running kube burner through  kube-burner-ocp.
Also sleep between (similar to our cdv2 churn) steps will be helpful so that CPU & mem of (OVN, kube-apiserver) components will stabilize from high spikes before taking further load.

rsevilla87:
> pass "oc get co" equivalent code if they are running kube burner through kube-burner-ocp.
> Also sleep between (similar to our cdv2 churn) steps will be helpful so that CPU & mem of (OVN, kube-apiserver) components will stabilize from high spikes before taking further load.

Hey Anil, a custom health check as you suggest is out of scope of this RFE, it can be an interesting idea though.

rsevilla87:
> I would like to see us have some user defined parameters to increase load over-time.
> 
> Here is what I am thinking... `--min-iterations` - What is the minimum number of iterations (basically the start of the test) `--step-size` - How much we want to increment the iterations to `--max-iterations` - Where do we stop. We could have 0 be keep going until health-check fails.
> 
> After each "step" we run the health-check to determine if the cluster is healthy, I think having some mechanism to poll the cluster for 5m (for example) to ensure nothing is flapping -- versus how we do healthy check now as a snapshot in time.

Overall this approach make sense to me, at this point I'm more worried about the implementation details, these high level flags shouldn't be added to "kube-burner", instead they can live within "kube-burner-ocp". The implementation in kube-burner itself should be driven through its configuration files like any other feature.

Something along these lines:

```yaml
jobss:
[...]

PR #1088, opened 2026-01-25
Adding support for incremental load additive version (Revisit Later)

vishnuchalla:
## Type of change

- Refactor
- New feature
- Optimization
- Documentation
- CI

## Description
Adding support for incremental load. 
* For both linear or exponential and relevant options.
* Adding support for custom health check script.
* Metrics logic remains the same as some one can trace them through the timestamps in the logs.
* Runtime measurements are looked in overall.

- Related Issue # https://github.com/kube-burner/kube-burner/issues/1059

## Testing
Tested and verified in a local AWS cluster. Also added a CI test for the same.
```
time="2026-01-25 17:02:23" level=info msg="Running incremental step: iterations [0 → 5)" file="incremental.go:180"
time="2026-01-25 17:02:23" level=info msg="0/5 iterations completed" file="create.go:134"
W0125 17:02:24.089182 1914565 warnings.go:70] would violate PodSecurity "restricted:latest": allowPrivilegeEscalation != false (container "kubelet-density" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "kubelet-density" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "kubelet-density" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "kubelet-density" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
W0125 17:02:24.256644 1914565 warnings.go:70] would violate PodSecurity "restricted:latest": allowPrivilegeEscalation != false (container "kubelet-density" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "kubelet-density" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "kubelet-density" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "kubelet-density" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
W0125 17:02:24.752316 1914565 warnings.go:70] would violate PodSecurity "restricted:latest": allowPrivilegeEscalation != false (container "kubelet-density" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "kubelet-density" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "kubelet-density" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "kubelet-density" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
W0125 17:02:25.250216 1914565 warnings.go:70] would violate PodSecurity "restricted:latest": allowPrivilegeEscalation != false (container "kubelet-density" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "kubelet-density" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "kubelet-density" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "kubelet-density" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
W0125 17:02:25.762777 1914565 warnings.go:70] would violate PodSecurity "restricted:latest": allowPrivilegeEscalation != false (container "kubelet-density" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "kubelet-density" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "kubelet-density" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "kubelet-density" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
time="2026-01-25 17:02:25" level=info msg="Waiting up to 4h0m0s for actions to be completed" file="create.go:260"
time="2026-01-25 17:02:26" level=info msg="Actions in namespace kubelet-density completed" file="waiters.go:77"
time="2026-01-25 17:02:27" level=info msg="Cluster health check passed, proceeding to next step" file="executor.go:155"
time="2026-01-25 17:02:27" level=info msg="Sleeping 2s before next step" file="incremental.go:212"
time="2026-01-25 17:02:29" level=info msg="Running incremental step: iterations [5 → 10)" file="incremental.go:180"
```

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -128,13 +128,9 @@ This section contains the list of jobs `kube-burner` will execute. Each job can 
 | `executionMode`              | Job execution mode. More details at [execution modes](#execution-modes)                                                               | String   | parallel |
 | `objectDelay`                | How long to wait between each object in a job                                                                                         | Duration | 0s       |
 | `objectWait`                 | Wait for each object to complete before processing the next one - not for Create jobs                                                 | Boolean  | 0s       |
-| `metricsAggregate`           | Aggregate the metrics collected for this job with those of the next one                                                               | Boolean  | false    |
+| `metricsAggregate`           | Aggregate the metrics collected for this job with those of the next one                                                               | Boolean  | false |
 | `metricsClosing`             | To define when the metrics collection should stop. More details at [MetricsClosing](#MetricsClosing)                                  | String   | afterJobPause |
-<<<<<<< HEAD
-| `incrementalLoad`            | Enables incremental load behaviour for creation jobs. See [Incremental Load](#incremental-load).                                      | Object   | {}            |
-=======
-| `incrementalLoad`           | Enables incremental load behaviour for creation jobs. See [Incremental Load](#incremental-load).                                      | Object   | {}            |
->>>>>>> incremental_load
+| `incrementalLoad`           | Enables incremental load behaviour for creation jobs. See [Incremental Load](#incremental-load).                                      | Object   | {} |
 
 !!! note
     Both `churnCycles` and `churnDuration` serve as termination conditions, with the churn process halting when either condition is met first. If someone wishes to exclusively utilize `churnDuration` to control churn, they can achieve this by setting `churnCycles` to `0`. Conversely, to prioritize `churnCycles`, one should set a longer `churnDuration` accordingly.
@@ -155,13 +151,8 @@ Examples of valid configuration files can be found in the [examples folder](http
 
 | Option | Description | Type | Default |
 |--------|-------------|------|---------|
-<<<<<<< HEAD
-| `minIterations` | Minimum iterations to start with. If omitted, the job's `jobIterations` is used. | Integer | `jobIterations` |
-| `maxIterations` | Maximum iterations to reach. If omitted, no increase beyond `minIterations` is performed. | Integer | same as `minIterations` |
-=======
 | `startIterations` | Initial number of iterations to start with. If omitted, the job's `jobIterations` is used. | Integer | `jobIterations` |
 | `totalIterations` | Total number of iterations to reach. If omitted, no increase beyond `startIterations` is performed. | Integer | same as `startIterations` |
->>>>>>> incremental_load
 | `stepDelay` | Delay between incremental steps (Go duration, e.g., `30s`). | Duration | `0s` |
 | `pattern.type` | Load pattern: `linear` or `exponential`. | String | `linear` |
 | `pattern.linear.minSteps` | Minimum number of steps for linear pattern. | Integer | `0` |
@@ -169,22 +160,14 @@ Examples of valid configuration files can be found in the [examples folder](http
 | `pattern.exponential.base` | Base of the exponential increase. | Float | `2.0` |
 | `pattern.exponential.maxIncrease` | Maximum tolerable increase (absolute iterations) for an exponential bump. | Integer | `0` |
 | `pattern.exponential.warmupSteps` | Number of linear warmup steps before applying exponential increases. | Integer | `0` |
-<<<<<<< HEAD
-| `healthCheckScript` | Optional shell script path (.i.e. local/remote) executed after each incremental step. If omitted, a built-in API + node readiness check is used. | String | `""` |
-=======
 | `healthCheckScript` | Optional shell script path (.i.e local/remote) executed after each incremental step. If omitted, a built-in API + node readiness check is used. | String | `""` |
->>>>>>> incremental_load
 
 !!! note
     Linear pattern falls back to a single step when `minSteps` is not provided and the implementation guards against division-by-zero when computing ranges. The runner will stop on any health-check error (script non-zero exit or built-in check failure).
 
 #### Incremental load behavior
 
-<<<<<<< HEAD
-The incremental load feature gradually increases the number of iterations from a configured minimum (`minIterations`) to a configured maximum (`maxIterations`) while retaining the resources created in each step. Two growth patterns are supported:
-=======
 The incremental load feature gradually increases the number of iterations from a configured start (`startIterations`) to a configured total (`totalIterations`) while retaining the resources created in each step. Two growth patterns are supported:
->>>>>>> incremental_load
 
 - Linear: iterations increase by a fixed amount each step (configured with `pattern.linear.stepSize`).
 - Exponential: iterations grow multiplicatively using `pattern.exponential.base`. An optional `pattern.exponential.warmupSteps` value can apply a few initial linear increases before exponential growth begins.
@@ -192,13 +175,8 @@ The incremental load feature gradually increases the number of iterations from a
 After each increase the runner performs the configured health check and will stop early on failure. Between successful steps the runner waits the configured `stepDelay` before applying the next increase.
 
 Simple examples:
-<<<<<<< HEAD
-- Linear: `minIterations=10`, `maxIterations=50`, `pattern.linear.stepSize=10` → 10, 20, 30, 40, 50 with constant increments of 10 iterations in each step.
-- Exponential: `minIterations=5`, `maxIterations=100`, `pattern.exponential.base=2` → 5, 10, 20, 40, 80, 100 by simply multiplying the iterations by 2 in each step.
-=======
 - Linear: `startIterations=10`, `totalIterations=50`, `pattern.linear.stepSize=10` → 10, 20, 30, 40, 50 with constant increments of 10 iterations in each step.
 - Exponential: `startIterations=5`, `totalmaxIterations=100`, `pattern.exponential.base=2` → 5, 10, 20, 40, 80, 100 by simply multiplying the iterations by 2 in each step.
->>>>>>> incremental_load
 
 These examples illustrate the typical behavior without requiring mathematical formulas.
 
--- a/pkg/util/cluster_health.go
+++ b/pkg/util/cluster_health.go
@@ -55,7 +55,7 @@ func areNodesHealthy(ctx context.Context, clientset kubernetes.Interface) bool {
 	var isHealthy = true
 	nodes, err := clientset.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
 	if err != nil {
-		log.Errorf("error getting nodes: %v", err)
+		log.Errorf("Error getting nodes: %v", err)
 		return false
 	}
 
--- a/examples/metrics-profiles/alerts.yml
+++ b/examples/metrics-profiles/alerts.yml
@@ -1,3 +1,0 @@
-- expr: up == 1
-  description: Prometheus {{$labels.instance}} is up & running
-  severity: warning
--- a/pkg/config/types.go
+++ b/pkg/config/types.go
@@ -202,7 +202,7 @@ type Job struct {
 	GC bool `yaml:"gc" json:"gc"`
 	// Measurements job-specific measurements to enable
 	Measurements []mtypes.Measurement `yaml:"measurements" json:"measurements,omitempty"`
-	// IncrementalLoad enables incremental load behaviour for creation jobs
+	// IncrementalLoad enables incremental load behavior for creation jobs
 	IncrementalLoad *IncrementalLoad `yaml:"incrementalLoad" json:"incrementalLoad,omitempty"`
 }
 


pkg/burner/executor.go
```diff
@@ -96,9 +101,61 @@ func newExecutor(configSpec config.Spec, kubeClientProvider *config.KubeClientPr
...
+					s := string(data)
+					if !strings.Contains(strings.ToLower(s), "ok") {
+						return fmt.Errorf("apiserver health check failed: %s", s)
+					}
+				}
+			}
+		}
+	}
+	// 2) Check node conditions
+	nodes, err := ex.clientSet.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
```
ygalblum: In the beginning of the function you check `if ex.clientSet != nil` but here you don't. Is the one in the beginning really needed?

pkg/burner/incremental.go
```diff
@@ -0,0 +1,218 @@
...
+	case config.ExponentialPattern:
+		base := 2.0
+		maxInc := cfg.Pattern.Exponential.MaxIncrease
+		warmup := cfg.Pattern.Exponential.WarmupSteps
+		if cfg.Pattern.Exponential != nil && cfg.Pattern.Exponential.Base > 0 {
+			base = cfg.Pattern.Exponential.Base
+		}
+		return &exponentialCalculator{min: minIt, max: maxIt, base: base, maxIncrease: maxInc, warmup: warmup, stepNo: 0}
+	default:
+		// default to linear behavior
```
ygalblum: Can't this code be merged with the Linear code case? (e.g. using `fallthrough`)

pkg/burner/incremental.go
```diff
@@ -0,0 +1,218 @@
...
+	"time"
+
+	"github.com/kube-burner/kube-burner/v2/pkg/config"
+	"github.com/kube-burner/kube-burner/v2/pkg/util"
+	log "github.com/sirupsen/logrus"
+)
+
+// IterationCalculator returns the next iteration window for incremental runs.
+type IterationCalculator interface {
+	Next(current int) (start, end int, done bool)
```
ygalblum, line 30: Why is `current` a parameter and not a field in the struct? The caller maintains `current` only for the sake of calling `Next`

pkg/burner/incremental.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	candidate := current + increase
+	if candidate > e.max {
+		candidate = e.max
+	}
+	return current, candidate, false
+}
+
+// RunIncrementalCreateJob executes incremental steps using the provided calculator.
+func RunIncrementalCreateJob(
```
ygalblum: Why is this a function and not a method of `JobExecutor`?

pkg/burner/incremental.go
```diff
@@ -0,0 +1,218 @@
...
+	ctx context.Context,
+	ex JobExecutor,
+	calculator IterationCalculator,
```
ygalblum, line 153: Why does it need to get the `IterationCalculator` as a parameter and not create it internally?

pkg/burner/incremental.go
```diff
@@ -0,0 +1,218 @@
...
+	stepDelay time.Duration,
```
ygalblum: The `stepDelay` is already in available through `ex.IncrementalLoad.StepDelay`

pkg/burner/utils.go
```diff
@@ -195,6 +195,8 @@ func (ex *JobExecutor) Verify() bool {
 		var objectsExpected int
 		if obj.RunOnce {
 			objectsExpected = obj.Replicas
+		} else if ex.IncrementalLoad != nil && (ex.IncrementalLoad.MaxIterations > 0 || ex.IncrementalLoad.MinIterations > 0) {
+			objectsExpected = obj.Replicas * int(math.Max(float64(ex.IncrementalLoad.MinIterations), float64(ex.IncrementalLoad.MaxIterations)))
```
ygalblum: Not sure I understood this calculation. Why is it the max of these two numbers? Don't you need to know all the different steps?

pkg/burner/executor.go
vishnuchalla: Good catch. This is not required, removed it.

pkg/burner/incremental.go
vishnuchalla: Done.

pkg/burner/incremental.go
vishnuchalla: This was intentional. If current were stored inside the calculator, the calculator would be, controlling execution flow, mutating internal state and becoming harder to reason about and reuse. So avoid that and provide user with flexibility to make calls based on scenarios, I have kept it this way.

pkg/burner/incremental.go
vishnuchalla: Make sense. Updated as suggested.

pkg/burner/incremental.go
vishnuchalla: Its again a separation of concern. So that the policy creation and code execution mechanism are isolated. So that the code is easily testable, extendable and maintainable in future.

pkg/burner/incremental.go
vishnuchalla: Changed as suggested.

pkg/burner/utils.go
vishnuchalla: Its derived from the logic [here](https://github.com/vishnuchalla/kube-burner/blob/incml_load/pkg/burner/incremental.go#L36,L46), further simplified it.

pkg/burner/incremental.go
ygalblum: 👍🏼

pkg/burner/incremental.go
ygalblum: 👍🏼

pkg/burner/utils.go
```diff
@@ -195,6 +195,8 @@ func (ex *JobExecutor) Verify() bool {
 		var objectsExpected int
 		if obj.RunOnce {
 			objectsExpected = obj.Replicas
+		} else if ex.IncrementalLoad != nil && (ex.IncrementalLoad.MaxIterations > 0 || ex.IncrementalLoad.MinIterations > 0) {
+			objectsExpected = obj.Replicas * ex.IncrementalLoad.MaxIterations
```
ygalblum: Sorry, but I still don't understand. This calculation assumes that only the max iterations is used. But IIUC that's not the case. The number of total iterations is the sum of the all the iteration steps.
For example, if `MinIterations` is 1, `MaxIterations` is 3 and `step` is 1. Then, the total number of iterations is `1+2+3=6` and not just `3`. What am I missing?

pkg/burner/utils.go
ygalblum: I saw that you've changed it. So, I'm resolving this comment and added a new comment instead explaining what I don't understand

pkg/burner/utils.go
vishnuchalla: Total number of iterations is going to be 3 which is same as max and the iterations progress from 1->2->3 with taking each step at a time which is of value 1.

ygalblum: Sorry, I don't understand. How can it be that the iterations are 1, 2 and 3 and the total number of iterations is 3. Maybe I'm misunderstanding the entire flow.

My understanding is that this is sort of a wrapper on top of job iterations. So, that the same job is executed first with 1 iteration, then with 2 and then with 3 and this check happens at the end of this loop, leading to a total of 6 iterations. What am I missing?

vishnuchalla: ohh no, numbers are correct. But the explanation is different. if min=1, max=3 and step=1, our aim to start at 1 and eventually get to 3. So we start with 1 iteration and once it is completed we keep the resources as is and then add step size 1 to create another iteration totaling to 2 on the cluster in terms of available resources and then add one more step after 2nd iteration objects are up and running to reach to 3. So at the end, in total we will end up having 3 iterations on the cluster in an incremental way.

ygalblum: OK, thanks for the explanation. I reread the code and now understand your point.
But, I have to say I wouldn't have understood that from the documentation. I think an example will be very helpful.

pkg/burner/executor.go
```diff
@@ -96,9 +101,58 @@ func newExecutor(configSpec config.Spec, kubeClientProvider *config.KubeClientPr
 	default:
 		log.Fatalf("Unknown jobType: %s", job.JobType)
 	}
+	ex.HealthCheckFunc = ex.defaultHealthCheck
 	return ex
 }
 
+// defaultHealthCheck performs Kubernetes-native health checks that should work across distributions.
+func (ex *JobExecutor) defaultHealthCheck(ctx context.Context) error {
```
rsevilla87: Why coding a new duplicated function if we have this https://github.com/kube-burner/kube-burner/blob/main/pkg/util/cluster_health.go#L12? already

docs/reference/configuration.md
```diff
@@ -144,6 +145,25 @@ Our configuration files strictly follow YAML syntax. To clarify on List and Obje
...
+|--------|-------------|------|---------|
+| `minIterations` | Minimum iterations to start with. If omitted, the job's `jobIterations` is used. | Integer | `jobIterations` |
+| `maxIterations` | Maximum iterations to reach. If omitted, no increase beyond `minIterations` is performed. | Integer | same as `minIterations` |
+| `stepDelay` | Delay between incremental steps (Go duration, e.g., `30s`). | Duration | `0s` |
+| `pattern.type` | Load pattern: `linear` or `exponential`. | String | `linear` |
+| `pattern.linear.minSteps` | Minimum number of steps for linear pattern. | Integer | `0` |
+| `pattern.linear.stepSize` | Fixed step size (iterations) for linear pattern. When set, steps will increment by this value. | Integer | `1` |
+| `pattern.exponential.base` | Base of the exponential increase. | Float | `2.0` |
+| `pattern.exponential.maxIncrease` | Maximum tolerable increase (absolute iterations) for an exponential bump. | Integer | `0` |
+| `pattern.exponential.warmupSteps` | Number of linear warmup steps before applying exponential increases. | Integer | `0` |
```
rsevilla87, line 162: Are we sure we need such amount of options?, is there a use case for all of them?

rsevilla87: Having a complex UX usually leads to an config over complexity that scare new users.

pkg/burner/executor.go
vishnuchalla: Thought it was only part of our downstream wrapper. Modified the existing one slightly. Thanks for the catch.

docs/reference/configuration.md
vishnuchalla: These are two simple options compared to upstream which is way too [complicated](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/api/types.go#L178). I have made it very clear with examples in documentation and also our tests have an example of usage. Let me know if there is anything else I can do to further simplify.

pkg/burner/utils.go
vishnuchalla: Agreed. Added more details in documentation.

docs/reference/configuration.md
```diff
@@ -165,6 +165,59 @@ Examples of valid configuration files can be found in the [examples folder](http
...
+If `warmupSteps = w > 0`, the first $w$ steps are increased more slowly (linear warmup), then exponential growth applies starting from the value reached at step $w$.
+
+After each step the runner performs the configured health check. If the health check fails the process stops early. Between successful steps the runner waits $\Delta t$ before applying the next $I_j$.
+
+#### Numeric examples (step-by-step)
+
+Linear (explicit):
+- Params: $m=10$, $N=50$, $s=10$, $\Delta t=30\text{s}$
+- Number of steps: $t=\left\lfloor\dfrac{50-10}{10}\right\rfloor+1=5$
+- Iterations per step: $I_1=10,\ I_2=20,\ I_3=30,\ I_4=40,\ I_5=50$
```
ygalblum: This line still reads like what I initially understood. That the first step has 10 iterations, then the second one has 20 and so on, leading to 10+20+30+40+50=150 iterations.
I'm wondering if defining the parameters the way I understand them won't be more intuitive to the users.

pkg/util/cluster_health.go
```diff
@@ -20,25 +22,54 @@ func ClusterHealthCheck(clientSet kubernetes.Interface) {
 
 func ClusterHealthyVanillaK8s(clientset kubernetes.Interface) bool {
 	var isHealthy = true
+
+	// 1) Check API server healthz
+	if discovery := clientset.Discovery(); discovery != nil {
```
ygalblum: What's the expected behaviour if `clientset.Discovery()` returns `nil`? Does silently disregarding the failure make sense here?

pkg/util/cluster_health.go
```diff
@@ -20,25 +22,54 @@ func ClusterHealthCheck(clientSet kubernetes.Interface) {
...
+		if rc := discovery.RESTClient(); rc != nil {
```
ygalblum: Same question

pkg/util/cluster_health.go
```diff
@@ -20,25 +22,54 @@ func ClusterHealthCheck(clientSet kubernetes.Interface) {
...
+			if data, err := rc.Get().AbsPath("/healthz").DoRaw(context.Background()); err != nil {
+				return false
+			} else {
+				// common healthy response contains "ok"
+				s := string(data)
+				if !strings.Contains(strings.ToLower(s), "ok") {
+					isHealthy = false
+					log.Errorf("apiserver health check failed: %s", s)
```
ygalblum: Why is failing to get `/healthz` enough to return `false`, but here the function continues?

pkg/util/cluster_health.go
```diff
@@ -20,25 +22,54 @@ func ClusterHealthCheck(clientSet kubernetes.Interface) {
...
+	}
+
+	// 2) Check node conditions
 	nodes, err := clientset.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
 	if err != nil {
 		log.Errorf("Error getting nodes: %v", err)
 		return false
 	}
-
+	var problems []string
```
ygalblum: Why do we need to accumulate `problems` at the end? We're just printing them out.

rsevilla87: yeah, the previous approach was correct and more simple IMO

pkg/util/cluster_health.go
vishnuchalla: Good point. Refactored the logic to avoid nested branching.

pkg/util/cluster_health.go
vishnuchalla: That is a quick check to do a ping on the apiserver. Refactored logic does an early return.

pkg/util/cluster_health.go
vishnuchalla: I just wanted to log all the errors in detail. And if the health check fails, we do a fatal and stop the execution. So should be fine IMHO.

docs/reference/configuration.md
vishnuchalla: Removed all the complicated math and updated the wording with simple examples.

docs/reference/configuration.md
```diff
@@ -165,58 +165,20 @@ Examples of valid configuration files can be found in the [examples folder](http
 !!! note
     Linear pattern falls back to a single step when `minSteps` is not provided and the implementation guards against division-by-zero when computing ranges. The runner will stop on any health-check error (script non-zero exit or built-in check failure).
 
-#### Math Explanation
+#### Incremental load behavior
 
-Let:
+The incremental load feature gradually increases the number of iterations from a configured minimum (`minIterations`) to a configured maximum (`maxIterations`) while retaining the resources created in each step. Two growth patterns are supported:
```
ygalblum: I'm sorry, I really don't understand how we're reading this differently.
To me, this paragraph says that the number of iterations will be increased from a minimum to maximum in steps. That means that it will first run the `minimum` number of iterations, then `minimum + step` iterations (in a loop) until it reaches the `maximum` number of iterations.
I don't see how this reads as it will run the `minimum` number and then the `step` (in loop) until reaching `maximum`

pkg/util/cluster_health.go
rsevilla87: It was already reporting all the unmet status conditions before, what is the difference here?

docs/reference/configuration.md
vishnuchalla: > I don't see how this reads as it will run the minimum number and then the step (in loop) until reaching maximum

I tried stressing this aspect by saying `retaining the resources created in each step` and by providing a simple example. But I am open to any suggestions you have on the rewording.

From the audience standpoint people might have different opinions/interpretations from the documentation until unless they try this out.

pkg/util/cluster_health.go
vishnuchalla: When I initially implemented this, I thought of logging all the errors in one big string. Later when I saw a exiting implementation I applied the similar updates to it. Either way it doesn't really matter on how these errors are logged. So restored the previous logic into this function.

docs/reference/configuration.md
ygalblum: `retaining the resources created in each step` doesn't say anything about the total number of iterations.

> From the audience standpoint people might have different opinions/interpretations from the documentation until unless they try this out.

I completely disagree. The documentation needs to be clear. Yes, sometimes it is not written well on the first try. But, it should not be left for interpretation when knowing it is ambiguous.

As for wording, I think `min` and `max` are incorrect terms to describe the implemented behaviour.  Furthermore, maybe defining the parameters this way are hard to explain to begin with. 
Maybe the correct way to address it is not to fix the documentation, but the implementation. Especially since this feature is not only about the incremental iteration, but also about adding a health check between steps. So, someone might want to keep the number of iterations constant, but with health checks. For example, run a total of 6 iterations, but run a health check every 2. In this case, the `step` is 0, but it's still provides value.
For this the parameters should be `startNumIterations`, `totalNumIterations` and `step` (allowing for 0, linear or exponential).

docs/reference/configuration.md
```diff
@@ -144,6 +145,40 @@ Our configuration files strictly follow YAML syntax. To clarify on List and Obje
...
+| `pattern.type` | Load pattern: `linear` or `exponential`. | String | `linear` |
+| `pattern.linear.minSteps` | Minimum number of steps for linear pattern. | Integer | `0` |
+| `pattern.linear.stepSize` | Fixed step size (iterations) for linear pattern. When set, steps will increment by this value. | Integer | `1` |
+| `pattern.exponential.base` | Base of the exponential increase. | Float | `2.0` |
+| `pattern.exponential.maxIncrease` | Maximum tolerable increase (absolute iterations) for an exponential bump. | Integer | `0` |
+| `pattern.exponential.warmupSteps` | Number of linear warmup steps before applying exponential increases. | Integer | `0` |
+| `healthCheckScript` | Optional shell script path (.i.e local/remote) executed after each incremental step. If omitted, a built-in API + node readiness check is used. | String | `""` |
+
+!!! note
+    Linear pattern falls back to a single step when `minSteps` is not provided and the implementation guards against division-by-zero when computing ranges. The runner will stop on any health-check error (script non-zero exit or built-in check failure).
```
ygalblum, line 166: Not sure I understand what you mean by `single step` here. If it is  not set, then kube burner should just use the start, total, and step size.

docs/reference/configuration.md
```diff
@@ -144,6 +145,40 @@ Our configuration files strictly follow YAML syntax. To clarify on List and Obje
...
+
+The incremental load feature gradually increases the number of iterations from a configured start (`startIterations`) to a configured total (`totalIterations`) while retaining the resources created in each step. Two growth patterns are supported:
+
+- Linear: iterations increase by a fixed amount each step (configured with `pattern.linear.stepSize`).
+- Exponential: iterations grow multiplicatively using `pattern.exponential.base`. An optional `pattern.exponential.warmupSteps` value can apply a few initial linear increases before exponential growth begins.
+
+After each increase the runner performs the configured health check and will stop early on failure. Between successful steps the runner waits the configured `stepDelay` before applying the next increase.
+
+Simple examples:
+- Linear: `startIterations=10`, `totalIterations=50`, `pattern.linear.stepSize=10` → 10, 20, 30, 40, 50 with constant increments of 10 iterations in each step.
```
ygalblum, line 178: I thought we agreed that stepSize is the acceleration of the number of iterations in each loop. I don't see how someone can read `10, 20, 30, 40, 50` and understand that there will be a total of 50 iterations

docs/reference/configuration.md
```diff
@@ -144,6 +145,40 @@ Our configuration files strictly follow YAML syntax. To clarify on List and Obje
...
+- Exponential: `startIterations=5`, `totalmaxIterations=100`, `pattern.exponential.base=2` → 5, 10, 20, 40, 80, 100 by simply multiplying the iterations by 2 in each step.
```
ygalblum, line 179: Same here

vishnuchalla:
Moving this to draft in favor of this: https://github.com/kube-burner/kube-burner/pull/1150

vishnuchalla:
Closing this to revisit later.

PR #1089, opened 2026-01-29
Adding space between measurement value and unit.

afcollins:
## Type of change

- Refactor

## Description

Without a space between measurement value and the units, double-click selects the unit, which is then not processed as a number and requires manual removal.

A space was added for service latency measurement before in #1050, I must have missed this one at that time.

I hope this is alright and doesn't interfere with any who might have created log-parsers.

- Related Issue #1030


rsevilla87:
FYI @ygalblum

ygalblum:
> FYI @ygalblum

👍🏼

PR #1096, opened 2026-01-30
Avoid log.Fatal in library packages

### LINKED ISSUE #1095: Library code uses log.Fatal, preventing graceful error handling
Some packages under `pkg/` use `log.Fatal` / `log.Fatalf` for error handling.
Since this is library code, these calls terminate the entire process, which
makes kube-burner harder to reuse as a library and complicates testing and
graceful shutdown.

Replacing these calls with proper error returns would allow errors to be
handled by callers while keeping the existing behavior and logging intact.

Nesar976:
This PR removes usage of `log.Fatal` / `log.Fatalf` from library code under
`pkg/`. Exiting the process from these packages makes kube-burner harder to
reuse as a library and complicates testing and graceful shutdown.

Errors are now returned to callers instead, while keeping the existing logging
and behavior intact.

Fixes #1095

### DIFF
--- a/hack/tag_name.sh
+++ b/hack/tag_name.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 if [[ -z $(git branch --show-current) ]]; then
-  git describe --tags --abbrev=0
+  git describe --tags --abbrev=0 2>/dev/null || echo "dev"
 else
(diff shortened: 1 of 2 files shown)


vishnuchalla:
Closing this as stale and not required at the moment.

PR #1102, opened 2026-02-01
fix(burner): fix objectOperations metric on early returns

### LINKED ISSUE #1101: Metric counter may be skipped on early error returns
Some code paths return early on error before incrementing: atomic.AddInt32(&ex.objectOperations, 1)

When this happens, failed attempts are not reflected in metrics, which can lead to under-counting and inaccurate reporting. This mainly affects metric accuracy and observability rather than functionality.

Nesar976:
## Type of change
- Bug fix

## Description
While working on the request creation flow, I noticed that the
`objectOperations` metric was not always being incremented when the
function returned early due to an error.

This change moves the metric increment to a deferred statement so it
is consistently recorded in all execution paths, avoiding missed
increments and preventing double counting.

Please let me know if any additional adjustments are needed.

Fixes #1101

### DIFF
--- a/pkg/burner/create.go
+++ b/pkg/burner/create.go
@@ -301,7 +301,6 @@ func (ex *JobExecutor) waitForCompletion(iterationStart, iterationEnd int, ns st
 }
 
 func (ex *JobExecutor) createRequest(ctx context.Context, gvr schema.GroupVersionResource, ns string, obj *unstructured.Unstructured, timeout time.Duration) {
-	defer atomic.AddInt32(&ex.objectOperations, 1)
 	var uns *unstructured.Unstructured
 	var err error
 	if log.GetLevel() == log.TraceLevel {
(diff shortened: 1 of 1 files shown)


pkg/burner/create.go
```diff
@@ -301,6 +301,7 @@ func (ex *JobExecutor) waitForCompletion(iterationStart, iterationEnd int, ns st
 }
 
 func (ex *JobExecutor) createRequest(ctx context.Context, gvr schema.GroupVersionResource, ns string, obj *unstructured.Unstructured, timeout time.Duration) {
+	defer atomic.AddInt32(&ex.objectOperations, 1)
```
rsevilla87: This is not valid, the requests can be retried due to RetryWithExponentialBackOff I think is better to place `atomic.AddInt32(&ex.objectOperations, 1)`  right after line 314

Nesar976:
Hi @rsevilla87 Thanks for pointing this out. I’ve removed the defer and moved atomic.AddInt32(&ex.objectOperations, 1) right after line 314 inside the RetryWithExponentialBackOff callback as suggested.

PR #987, opened 2025-09-11
Resource level churning

rsevilla87:
## Type of change

- Refactor
- New feature
- Optimization

## Description

Enable resource level churning along with a configuration reorg and code improvements.
cc @mcornea @mukrishn 

## Related Tickets & Documents

- Related Issue #
- Closes #984

PR #1105, opened 2026-02-01
Fix watcher stop errors being silently discarded

Aman-Cool:
## Fix: watcher stop errors were silently dropped

This PR fixes a regression where errors returned from `watcherManager.StopAll()` were ignored because the return value of `slices.Concat` was not assigned.

As a result, kube-burner could report success (exit code `0`) even when watcher shutdown failed, hiding real issues like informer cache sync timeouts and misleading CI pipelines.

### What changed
- Properly append watcher stop errors to the main error slice so failures are propagated.

### Why it matters
- Restores correct exit codes
- Prevents silent failures
- Brings back behavior that existed before PR #987

This is a minimal, one-line fix that restores a core correctness guarantee: failed operations must not be silently ignored.

### DIFF
--- a/pkg/burner/job.go
+++ b/pkg/burner/job.go
@@ -89,7 +89,7 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider, 
 		_, restConfig := kubeClientProvider.DefaultClientSet()
 		measurementsFactory := measurements.NewMeasurementsFactory(configSpec, metricsScraper.MetricsMetadata, additionalMeasurementFactoryMap)
 		jobExecutors = newExecutorList(configSpec, kubeClientProvider, embedCfg)
-		handlePreloadImages(jobExecutors, kubeClientProvider)
+		handlePreloadImages(ctx, jobExecutors, kubeClientProvider)
 		// Iterate job list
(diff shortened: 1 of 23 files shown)


Aman-Cool:
@rsevilla87 @ygalblum ,Fixes a regression where watcher shutdown errors were silently ignored, causing kube-burner to report success even when it shouldn’t.

pkg/burner/job.go
```diff
@@ -208,7 +207,7 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
 				measurementsInstance = nil
 			}
 			watcherStopErrs := watcherManager.StopAll()
-			slices.Concat(errs, watcherStopErrs)
+			errs = append(errs, watcherStopErrs...)
```
ygalblum, line 210: The real issue here is that the returned array was not stored. I'm trying to understand if the called function should change here.
@rsevilla87 is there a reason why we used `slices.Concat` and not `append` in the original code?

rsevilla87: Don't remember an specific reason now, it was probably a mistake

ygalblum (approved):

PR #1109, opened 2026-02-02
Handle cancellation and errors in async goroutines

### LINKED ISSUE #1108: Asynchronous goroutines lack cancellation and error propagation
Several goroutine-based workflows rely solely on sync.WaitGroup for synchronization, without any mechanism to propagate cancellation or surface execution errors. This can result in silent failures, orphaned goroutines, and unnecessary resource usage under failure or shutdown conditions. Adding explicit cancellation and error propagation would make these workflows safer and more predictable.

Nesar976:
This change updates async goroutine workflows to properly propagate context cancellation and surface execution errors. It prevents silent failures and ensures in-flight work stops cleanly when the parent operation is canceled, without altering existing behavior.

Fixes #1108

### DIFF
--- a/.github/workflows/test-k8s.yml
+++ b/.github/workflows/test-k8s.yml
@@ -38,12 +38,18 @@ jobs:
         k8s-version: ${{ fromJson(needs.fetch-latest-kind-node-tags.outputs.tags) }}
     steps:
 
-    - name: Check out code
+    - name: Checkout PR head
       uses: actions/checkout@v6
       with:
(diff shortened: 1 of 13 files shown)


rsevilla87:
Empty PR, the changes were probably addressed by another one, closing it

Nesar976:
Hi @rsevilla87 Thanks for the clarification. I might have missed that the issue was already addressed in another PR. Appreciate you pointing it out.

PR #1111, opened 2026-02-02
Cache compiled templates to avoid repeated runtime work

Nesar976:
This change moves template parsing and compilation out of hot execution paths and caches the compiled templates during initialization.
The cached templates are reused during metric scraping and alert evaluation, reducing unnecessary CPU overhead while keeping behavior and APIs unchanged.

Clean, minimal, and focused only on the performance issue.
Fixes #1100


vishnuchalla:
This optimization doesn't add any significant difference on templating logic. End the end they are all propagated in memory. Feel free to join community call if you still feel this is important.

PR #1116, opened 2026-02-03
Normalizing labels for node latency as a known issue

vishnuchalla:
## Type of change
Normalize labels for node latency measurement.

- Refactor
- Bug fix
- Optimization

## Description
Normalizes the labels for node latency measurement. This is a known issue with nodes changing their labels often. [similar fix](https://github.com/cloud-bulldozer/workers-scale/blob/849b2a909e9c2422cc5638bbfd170c54bc4ed030/workerscale/metrics.go#L98)

A test was run through our ocp-wrapper and using olm workload and this issue poped out.

- Closes # https://redhat-internal.slack.com/archives/C05CDC19ZKJ/p1770112625995489

## Testing
Tested and verified in local in a hacky way.
```
time="2026-02-03 14:10:33" level=info msg="map[beta_kubernetes_io/arch:amd64 beta_kubernetes_io/instance-type:m6a.xlarge beta_kubernetes_io/os:linux failure-domain_beta_kubernetes_io/region:us-west-1 failure-domain_beta_kubernetes_io/zone:us-west-1a kubernetes_io/arch:amd64 kubernetes_io/hostname:ip-10-0-0-125.us-west-1.compute.internal kubernetes_io/os:linux node-role_kubernetes_io/control-plane: node-role_kubernetes_io/master: node_kubernetes_io/instance-type:m6a.xlarge node_openshift_io/os_id:rhel topology_ebs_csi_aws_com/zone:us-west-1a topology_k8s_aws/zone-id:usw1-az1 topology_kubernetes_io/region:us-west-1 topology_kubernetes_io/zone:us-west-1a]" file="node_latency.go:218"
```

### DIFF
--- a/.github/workflows/test-k8s.yml
+++ b/.github/workflows/test-k8s.yml
@@ -38,12 +38,18 @@ jobs:
         k8s-version: ${{ fromJson(needs.fetch-latest-kind-node-tags.outputs.tags) }}
     steps:
 
-    - name: Check out code
+    - name: Checkout PR head
       uses: actions/checkout@v6
       with:
-        fetch-depth: 1
+        # Use the PR head SHA
         ref: ${{ github.event.pull_request.head.sha }}
+        # Fetch only the most recent commit of the PR
+        fetch-depth: 1
         persist-credentials: false
+
+    - name: Fetch base branch for comparison
+      # This tells Git about origin/main (or master) without downloading the whole repo
+      run: git fetch origin ${{ github.event.pull_request.base.ref }} --depth=1
 
     - name: Download kube-burner binary
       uses: actions/download-artifact@v7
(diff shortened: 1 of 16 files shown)


rsevilla87:
> I don't mind this change. But, why is it specific to node latency? The comment says that certain backends don't like the `.` separator. How does that not affect other measurements?

Because this "measurement" collects node labels.

vishnuchalla:
> > I don't mind this change. But, why is it specific to node latency? The comment says that certain backends don't like the `.` separator. How does that not affect other measurements?
> 
> Because this "measurement" collects node labels.

Yes, we saw mapping issues due to '.' in our ES from this metric. But I think its a good idea to put this function in utils and reuse.

PR #1122, opened 2026-02-04
Inject UUID variable to config file

rsevilla87:
## Type of change

- New feature

It's useful to have this variable available in the main config file

- Related Issue #
- Closes #

### DIFF
--- a/pkg/measurements/pvc_latency.go
+++ b/pkg/measurements/pvc_latency.go
@@ -129,7 +129,6 @@ func (p *pvcLatency) handleUpdatePVC(obj any) {
 		pm := value.(pvcMetric)
 		log.Tracef("handleUpdatePVC: PVC: [%s], Version: [%s], Phase: [%s]", pvc.Name, pvc.ResourceVersion, pvc.Status.Phase)
 		if pm.bound == 0 || pm.lost == 0 {
-			// https://pkg.go.dev/k8s.io/api/core/v1#PersistentVolumeClaimPhase
 			if pvc.Status.Phase == corev1.ClaimPending {
 				if pm.pending == 0 {
 					log.Debugf("PVC %s is pending", pvc.Name)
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -260,6 +260,7 @@ func ParseWithUserdata(uuid string, timeout time.Duration, configFileReader, use
 	}
 
 	inputData, err := getInputData(userDataFileReader, additionalVars)
+	inputData["UUID"] = uuid
```
vishnuchalla, line 263: Having this rendered into our configuration is fine. But do we really need this level of uniqueness at the moment?

rsevilla87: it's useful if you want for example to have
```
metricsDirectory: metrics-{{.UUID}}
```

vishnuchalla: Yeah, understood the usage. But not sure if we have many use-cases like that that require to be very unique in naming.

docs/reference/configuration.md
```diff
@@ -613,7 +613,7 @@ All object templates are injected with the variables below by default:
 - `Iteration`: Job iteration number.
 - `Replica`: Object replica number. Keep in mind that this number is reset to 1 with each job iteration.
 - `JobName`: Job name.
-- `UUID`: Benchmark UUID.
+- `UUID`: Benchmark UUID. (Can be also reference in the main configuration file)
```
```suggestion
- `UUID`: Benchmark UUID. (Can be also referenced in the main configuration file)
```

pkg/config/config.go
rsevilla87: At least two:
 
- Dynamic metricsDirectory w/o user intervention
- Workload re-execution w/o collision

The change is just one harmless and documented line

vishnuchalla: /lgtm

PR #1126, opened 2026-02-05
Fix file descriptor leak from workload readers

### LINKED ISSUE #1124: File descriptor leak due to unclosed reader returned by GetWorkloadReader()
GetWorkloadReader() returns a reader backed by a file handle that is never explicitly closed by the caller. During repeated runs or long-lived executions, this can lead to accumulated open file descriptors and eventually cause too many open files errors. I couldn’t find an existing issue tracking this behavior.

Nesar976:
Closes readers returned by GetWorkloadReader() at all call sites where they are backed by closable resources. This prevents file descriptor leaks during repeated or long-running executions while keeping existing behavior unchanged.

Fixes  #1124

### DIFF
--- a/.github/workflows/test-k8s.yml
+++ b/.github/workflows/test-k8s.yml
@@ -38,12 +38,18 @@ jobs:
         k8s-version: ${{ fromJson(needs.fetch-latest-kind-node-tags.outputs.tags) }}
     steps:
 
-    - name: Check out code
+    - name: Checkout PR head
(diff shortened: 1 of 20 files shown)


Nesar976:
@ygalblum Thanks for the review! Glad to see all checks passing.

Nesar976:
Hi @rsevilla87 ,
All checks are green now and the branch has been updated.
Let me know if you’d like anything else adjusted from my side.

PR #1128, opened 2026-02-06
Remove blocking file I/O from constructor

### LINKED ISSUE #1127: Blocking file I/O inside constructor makes object creation unsafe
Object initialization currently performs blocking file I/O (io.ReadAll) inside a constructor-like function. This makes object creation slow, harder to test, and risky in concurrent paths. Constructors should avoid heavy work and defer I/O to explicit initialization steps.

Nesar976:
Moves file reading out of the constructor and into explicit load paths. Object creation is now lightweight, easier to test, and safer for concurrent use. Behavior remains unchanged.

Fixes #1127

### DIFF
--- a/pkg/burner/job.go
+++ b/pkg/burner/job.go
@@ -19,7 +19,6 @@ import (
 	"errors"
 	"fmt"
 	"math"
-	"slices"
 	"sync"
 	"time"
 
@@ -208,7 +207,7 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider, 
 				measurementsInstance = nil
 			}
 			watcherStopErrs := watcherManager.StopAll()
-			slices.Concat(errs, watcherStopErrs)
+			errs = append(errs, watcherStopErrs...)
 			if jobExecutor.GC {
 				jobExecutor.gc(ctx, nil)
 			}
--- a/pkg/burner/object.go
+++ b/pkg/burner/object.go
@@ -74,20 +74,9 @@ func newObject(obj config.Object, mapper *restmapper.DeferredDiscoveryRESTMapper
 	}
 }
 
-	if obj.ObjectTemplate != "" {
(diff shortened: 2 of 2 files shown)


ygalblum (changes requested):
Why is this really better? It is not really a constructor (since GoLang doesn't have constructor). Caller must also call the `loadTemplate` function. I don't see the benefit in this change

Nesar976:
Hi @ygalblum This isn’t about Go having constructors, but about avoiding side effects during object creation. Doing blocking file I/O at init time makes creation slower, non-deterministic, and harder to reason about in concurrent paths. By moving I/O to an explicit load step, object creation stays cheap and predictable, failures are localized, and callers control when blocking work happens.

ygalblum:
You say that it's not about constructors and then explain that the issue is with the fact that it's a constructor.
In order for the object to work, loading has to be called. This means that you force the user to make an additional call for every created object. If they forget, things break. I don't see any real benefit in this change. There's nothing beneficial in having the object creation "cheap" and there are no side-affects in reading from I/O.

Nesar976:
@ygalblum Thanks for the clarification — I see your point about the added call and potential misuse.
I agree the change doesn’t provide enough benefit here to justify the API impact.
I’ll adjust the approach accordingly. Appreciate the review.

ygalblum:
> I agree the change doesn’t provide enough benefit here to justify the API impact.

This change basically imposes patterns from other languages, ones with constructors that cannot fail, onto a program that does not need it

> I’ll adjust the approach accordingly

Not sure there's anything needed here

Nesar976:
Hi @ygalblum Thanks for the detailed feedback. I agree that this change doesn’t provide enough benefit to justify the API impact in this case. I’ll drop this approach and close the PR accordingly. Appreciate the review and discussion.

ygalblum:
Closing the PR following the discussion.
@Nesar976 Thanks for your work

PR #1130, opened 2026-02-08
feat: support templated kind field by resolving REST mapping at runtime

Nesar976:
This PR adds support for templating the kind field in object templates by resolving the Kubernetes REST mapping at runtime, after the template has been rendered.

Previously, kube-burner attempted to resolve the REST mapping before template evaluation, which made it impossible to use templating expressions inside the kind field. As a result, workloads that dynamically generate different resource kinds (for example CRDs and CRs from a single job definition) could not be expressed cleanly.

With this change:
	•	The object template is rendered first.
	•	The resolved kind is then used to dynamically fetch the correct REST mapping.
	•	Object creation proceeds using the resolved GVK without requiring pre-expanded manifests.

This enables more flexible job definitions while preserving existing behavior for non-templated kinds.

•	Related Issue: #862

### DIFF
--- a/pkg/burner/create.go
+++ b/pkg/burner/create.go
@@ -140,8 +140,10 @@ func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterati
 		log.Debugf("Creating object replicas from iteration %d", i)
 		for objectIndex, obj := range ex.objects {
 			if obj.gvr == (schema.GroupVersionResource{}) {
-				// resolveObjectMapping may set ex.nsRequired to true if the object is namespaced but doesn't have a namespace specified
-				ex.resolveObjectMapping(obj)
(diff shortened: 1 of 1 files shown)


pkg/burner/create.go
```diff
@@ -234,7 +234,11 @@ func (ex *JobExecutor) replicaHandler(ctx context.Context, labels map[string]str
 			// The object's GVK might not have been resolvable during setupCreateJob() - perhaps the
 			// corresponding CRD might not be installed or the kube-apiserver isn't reachable at the moment.
 			_, err := ex.mapper.RESTMapping(gvk.GroupKind())
-
+			if err != nil {
```
vishnuchalla, line 237: Same block twice!

vishnuchalla:
Closing this as a no-op. Please feel free to join our community call to further discuss about it.

PR #1131, opened 2026-02-08
Optimize template compilation in hot paths

### LINKED ISSUE #1110: Templates Are Recompiled Repeatedly in Hot Paths
Templates are being parsed and compiled inside frequently executed code paths instead of being initialized once and reused. This adds unnecessary CPU overhead and increases latency during metric collection and alert evaluation.

At scale, this cost multiplies with workload size and can become a noticeable performance bottleneck.

Nesar976:
This change moves template parsing and compilation out of frequently executed paths and into profile initialization.

Alert expressions, alert descriptions, and metric queries are now compiled once when profiles are loaded and reused during evaluation and scraping. This avoids repeated template parsing inside hot loops, reducing unnecessary CPU overhead and latency, especially at scale.

The behavior and output remain unchanged; the update is strictly an internal performance optimization.
 
Fixes #1110

### DIFF
--- a/pkg/alerting/alert_manager.go
+++ b/pkg/alerting/alert_manager.go
@@ -45,15 +45,19 @@ const (
 	rcAlert                       = 3
 )
 
-// alertProfile expression list
-type alertProfile []struct {
+type alertDefinition struct {
 	// PromQL expression to evaluate
 	Expr string `yaml:"expr"`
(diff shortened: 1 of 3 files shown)


vishnuchalla:
This optimization doesn't add any significant difference on templating logic. End the end they are all propagated in memory. Feel free to join community call if you still feel this is important.

Issue #1008, opened 2025-10-22
gvr DeletionStrategy deletes namespaces created by kube-burner


samavasi:
Reading the docs/reference on DeletionStrategy, Global deletion strategy to apply, default or gvr (where default deletes entire namespaces and gvr deletes objects within namespaces)

To my understanding this strategy would target all objects created by jobs/objects of a kube-burner execution, but I see in the source code that after that cleanup there is a call to delete namespaces function also for gvr DeletionStrategy.

Also it seems that with gvr if I have kind: Job objects that have completed they dont get deleted, is it because they are in status completed or would it be expected that all objects created in any state would be cleaned up regardless?

rsevilla87:
Hey @samavasi , correct, gvr deletion strategy ends up by deleting the "emptied" namespaces, that works in that way by design.

wrt the issue with Jobs and the completed pods left, that was recently fixed, by adding the `DeletePropagationBackground` propagation policy when deleting the objects. https://github.com/kube-burner/kube-burner/blob/main/pkg/burner/namespaces.go#L54, the fix should be available in https://github.com/kube-burner/kube-burner/releases/tag/v2.0.0

Let me know if that works for you

sanjay7178:
@rsevilla87 While I was testing gvr strategy with kube-burner config.yaml , I figured out it initiates the gvr strategy to clean the objects within in namespaces and then again it cleans up the emptied namespaces with label `kube-burner-job=<job-name>` themselves .

Here's the problem trace : 
[...]

PR #1132, opened 2026-02-09
Fix gvr deletion strategy deleting kube-burner namespaces

Nesar976:
This PR fixes an issue where the `gvr` deletion strategy was deleting namespaces created by kube-burner.

According to the documentation and expected behavior:
- `default` deletion strategy deletes the entire namespace
- `gvr` deletion strategy should only delete objects within namespaces, not the namespace itself

The current behavior was incorrectly deleting namespaces even when `gvr` was selected.  
This change ensures that namespaces are preserved when using the `gvr` deletion strategy.

To prevent regressions, unit tests were added to explicitly verify:
- namespaces are **preserved** when `deletionStrategy = gvr`
- namespaces are **deleted** when `deletionStrategy = default`

All tests pass successfully.

Issue: #1008

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/prometheus/common v0.62.0
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/cobra v1.8.1
+	github.com/stretchr/testify v1.10.0
 	golang.org/x/exp v0.0.0-20240719175910-8a7402abbf56
 	golang.org/x/time v0.10.0
 	gonum.org/v1/gonum v0.15.1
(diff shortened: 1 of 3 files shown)


rsevilla87:
The current behaviour is intentional (delete namespaces after removing the child objects), you can update the docs accordingly to align the behavior with the documentation

Nesar976:
@rsevilla87  Thanks for the clarification.
Understood — if the current behavior (deleting the namespace after removing child objects) is intentional, then updating the documentation to reflect this makes sense.
I’ll follow up with a documentation update to align the docs with the actual behavior. Appreciate you pointing this out.

PR #1134, opened 2026-02-09
job: preserve watcher stop errors when aggregating errs

### LINKED ISSUE #1133: Watcher stop errors are lost due to ignored slices.Concat return value
slices.Concat(errs, watcherStopErrs) is called without assigning the returnedslice. Since slices.Concat returns a new slice, watcher stop errors are silently dropped and never reported, causing cleanup failures to be missed.

Nesar976:
Fixes a bug where watcher stop errors were dropped because the return value of
slices.Concat was not assigned. The aggregated error slice now correctly
includes watcher cleanup failures.

Fixes #1133


vishnuchalla:
already fixed

PR #1136, opened 2026-02-09
job: propagate runWaitList error

### LINKED ISSUE #1135: runWaitList error return value is ignored
runWaitList returns an error, but the return value is not checked or propagated.
As a result, failures during wait operations can be silently ignored, leading to
false success and missed error reporting.

Nesar976:
Fixes a case where the error returned by runWaitList was ignored, causing wait
failures to be silently dropped.

Fixes #1135


rsevilla87:
This PR is emtpy now @Nesar976

Nesar976:
Hi @rsevilla87 After merging main into this branch, the earlier commits no longer produce a diff. The original changes appear to have been incorporated into main, so the branch now matches the base.

Thanks for taking the time to review and approve this.

PR #1140, opened 2026-02-11
fix: prefer ingress IP over hostname for LoadBalancer services

### LINKED ISSUE #1139: LoadBalancer ingress uses hostname when IP is present in service latency handling
In handleCreateSvc, the LoadBalancer ingress handling uses ingress.Hostname when ingress.IP is present, and falls back to ingress.IP when it’s empty. This results in using the wrong address (or an empty one) for service latency checks.

This can lead to incorrect latency targets, unnecessary DNS resolution, and silently wrong metrics.

Expected behavior: prefer ingress.IP when available, otherwise fall back to ingress.Hostname, and never append empty values.

Location: pkg/measurements/service_latency.go

Nesar976:
Fixes the LoadBalancer ingress address selection in service latency handling.

When iterating over `svc.Status.LoadBalancer.Ingress`, the logic previously used
`ingress.Hostname` when an IP was present and fell back to `ingress.IP` when it
was empty. This resulted in incorrect or empty addresses being used for latency
checks.

The logic is corrected to:
- prefer `ingress.IP` when available
- fall back to `ingress.Hostname` only when IP is empty
- avoid appending empty values

The change is minimal and limited to the affected code path.
Fixes #1139

### DIFF
--- a/pkg/measurements/service_latency.go
+++ b/pkg/measurements/service_latency.go
@@ -140,9 +140,9 @@ func (s *serviceLatency) handleCreateSvc(obj any) {
 				case corev1.ServiceTypeLoadBalancer:
 					for _, ingress := range svc.Status.LoadBalancer.Ingress {
 						if ingress.IP != "" {
-							ips = append(ips, ingress.Hostname)
-						} else {
 							ips = append(ips, ingress.IP)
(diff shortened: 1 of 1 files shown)


vishnuchalla:
Invalid change for a load balancer type.

PR #1144, opened 2026-02-12
Fix latency skip annotation not being enforced

### LINKED ISSUE #1143: Latency skip annotation is logged but not enforced
KubeBurner detects the latency-related annotation and logs that latency is being discarded when the annotation value is set to "false".
However, latency processing still continues instead of stopping immediately.

This results in the annotation semantics being ignored.

Nesar976:
Description
When the latency skip annotation is set to "false", the code logs that latency is discarded but continues processing it.
This change adds an early return so latency processing stops immediately, enforcing the intended annotation semantics.

Fixes #1143


vishnuchalla:
Noop change

PR #1146, opened 2026-02-13
Fix RunOnce guard when iterationStart is non-zero

### LINKED ISSUE #1145: [BUG] RunOnce is silently skipped when iterationStart is not zero
RunOnce execution is currently guarded by an iteration index check that implicitly assumes iterations always start at 0.
When kube-burner runs in churn, recreate, or resume paths where iterationStart is non-zero, the RunOnce block is silently skipped even though it has never executed.

This results in missing one-time initialization logic without any error, warning, or log signal.

Nesar976:
Fixes RunOnce execution being tied to iteration == 0.
When execution resumes with a non-zero iterationStart, RunOnce was silently skipped.
This change makes RunOnce state-driven so it executes exactly once regardless of iteration offset.

Fixes #1145

### DIFF
--- a/.github/workflows/check-docs-links.yml
+++ b/.github/workflows/check-docs-links.yml
@@ -31,7 +31,7 @@ jobs:
         uses: lycheeverse/lychee-action@v2.7.0
         with:
           lycheeVersion: v0.20.1
-          args: --verbose --no-progress --exclude-loopback './docs/**/*.md'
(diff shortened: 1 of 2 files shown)


rsevilla87:
I do see changes not related with th PR title here

Nesar976:
Hi @rsevilla87 , thanks for the clarification due to past commits the unwanted chages have been appeared i will close this PR due to unwanted changes and i will raise a new PR for the issue

PR #1147, opened 2026-02-15
Downgrade transient status-not-found log from error to debug in verifyCondition

guchen11:
## Summary

When kube-burner creates custom resources (e.g. `VirtualMachine`, `DataVolume`), the `.status` field is not immediately populated by the controller. The first status check in `verifyCondition` (`waiters.go`) finds `!found` and logs at `level=error`, even though this is a normal transient state during object creation.

This produces noisy error lines like:
```
level=error msg="Error extracting or finding status in object VirtualMachine/vm-1: <nil>"
```

## Fix

Split the error handling in `verifyCondition`:
- `log.Errorf` only when there is an actual extraction error (`err != nil`) — return the error to stop retrying
- `log.Debugf` when status is simply not yet available (`!found`) — return `false, nil` to continue the retry loop

## Before
```go
status, found, err := unstructured.NestedMap(item.Object, "status")
if err != nil || !found {
    log.Errorf("Error extracting or finding status in object %s/%s: %v", ...)
    return false, err
}
```

## After
```go
status, found, err := unstructured.NestedMap(item.Object, "status")
if err != nil {
    log.Errorf("Error extracting status from object %s/%s: %v", ...)
    return false, err
}
if !found {
    log.Debugf("Status not yet available for %s/%s, retrying", ...)
    return false, nil
}
```

## Test Plan

- Tested with KubeVirt `VirtualMachine` creation using `kube-burner` on an OpenShift cluster
- Before: `level=error` logged on first status check for every newly created VM
- After: transient state handled silently at debug level, no more false error noise
- All existing wait/retry logic preserved — only the log level changes for the `!found` case

### DIFF
--- a/.github/workflows/test-k8s-ppc64le.yml
+++ b/.github/workflows/test-k8s-ppc64le.yml
@@ -26,7 +26,7 @@ jobs:
         cache: false
 
     - name: Install bats
-      uses: bats-core/bats-action@3.0.1
+      uses: bats-core/bats-action@4.0.0
       with:
         bats-version: 1.11.0
         support-install: false
--- a/.github/workflows/test-k8s.yml
+++ b/.github/workflows/test-k8s.yml
@@ -58,7 +58,7 @@ jobs:
         path: /tmp/
 
     - name: Install bats
-      uses: bats-core/bats-action@3.0.1
+      uses: bats-core/bats-action@4.0.0
       with:
         bats-version: 1.10.0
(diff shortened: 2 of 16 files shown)


ygalblum:
Thanks for the PR. I've been planning on doing this for a while now. However, before we can merge, please amend your commit and add a signoff. You can run:
```
git commit --amend --no-edit --signoff
git push --force
```

PR #1150, opened 2026-02-17
Incremental load feature cumulative version

vishnuchalla:
## Type of change

- Refactor
- New feature
- Optimization
- Documentation
- CI

## Description
Adding support for incremental load. 
* For both linear or exponential and relevant options.
* Adding support for custom health check script.
* All the metrics have a `incrementalLoadUUID` in their metadata.
* Runtime measurements are separate for each incremental run, same applies with prometheus metrics. But all have `incrementalLoadUUID` filed that helps us to identify that they are all part of same run.

- Related Issue # https://github.com/kube-burner/kube-burner/issues/1059

## Testing
Tested and verified in a local AWS cluster. Also added a CI test for the same.
Full logs: https://gist.github.com/vishnuchalla/6e6c2d869dd880206bd63e6977b6611e

Grafana Dashboard: https://grafana.rdu2.scalelab.redhat.com:3000/d/dff80doghob9cc/kube-burner-report-raw-uuid-incremental-load-copy?orgId=1&var-Datasource=QzcDu7T4z&var-uuid=904b42aa-a2df-4c44-822d-59fe82a09c24&var-incrementalLoadUUID=All&var-master=ip-10-0-26-88.us-west-2.compute.internal&var-worker=ip-10-0-19-109.us-west-2.compute.internal&var-infra=&var-latencyPercentile=P99&from=1772817552791&to=1772818245876

### Metrics in local
Example jobSummary of a step job containing incrementalLoadUUID
```
{
    "metricName": "jobSummary",
    "passed": true,
    "timestamp": "2026-03-06T17:20:38.878786518Z",
    "uuid": "904b42aa-a2df-4c44-822d-59fe82a09c24",
    "incrementalLoadUUID": "b0940795-5adc-49f3-83fd-4b53db416071",
    "version": "incremental-load-cummulative@1ef60366129100f44230a30c8bf4980cd7c4e556"
  },
```
Runtime measurements that have incrementalLoadUUID in metadata
```
  {
    "quantileName": "Ready",
    "uuid": "904b42aa-a2df-4c44-822d-59fe82a09c24",
    "P99": 2000,
    "P95": 2000,
    "P50": 1000,
    "min": 0,
    "max": 2000,
    "avg": 950,
    "timestamp": "2026-03-06T17:26:38.582931389Z",
    "metricName": "podLatencyQuantilesMeasurement",
    "jobName": "vchalla-test",
    "metadata": {
      "incrementalLoadUUID": "bccb30c2-3775-46a9-acf6-92a6fc24d101"
    }
  }
```
Prometheus metrics in child job that also have incrementalLoadUUID in metadata
```
  {
    "jobName": "vchalla-test",
    "labels": {
      "container": "cluster-baremetal-operator",
      "namespace": "openshift-machine-api",
      "node": "ip-10-0-65-229.us-west-2.compute.internal",
      "pod": "cluster-baremetal-operator-548b84b4b4-f8dmr"
    },
    "metadata": {
      "incrementalLoadUUID": "b0940795-5adc-49f3-83fd-4b53db416071"
    },
    "metricName": "containerCPU",
    "query": "(sum(irate(container_cpu_usage_seconds_total{name!=\"\",container!~\"POD|\",namespace=~\"openshift-.*|cilium|stackrox|calico.*|tigera.*\"}[2m]) * 100) by (container, pod, namespace, node)) > 0",
    "timestamp": "2026-03-06T17:21:38.879Z",
    "uuid": "904b42aa-a2df-4c44-822d-59fe82a09c24",
    "value": 0.05733399405358158
  }
```
Compressed version of collected metrics from local file system
[collected-metrics.zip](https://github.com/user-attachments/files/25801136/collected-metrics.zip)

### DIFF
--- a/pkg/burner/job.go
+++ b/pkg/burner/job.go
@@ -96,6 +96,14 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider, 
 			return
 		}
 
+		// Cleanup previous runs
+		for _, jobExecutor := range jobExecutors {
+			if jobExecutor.JobType == config.CreationJob && jobExecutor.Cleanup {
+				log.Infof("Cleaning up previous runs for job: %s", jobExecutor.Name)
+				jobExecutor.gc(ctx, nil)
+			}
+		}
+
 		// Iterate job list
 		var measurementsInstance *measurements.Measurements
 		var measurementsJobName string
@@ -125,10 +133,6 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider, 
 			}
 			log.Infof("Triggering job: %s", jobExecutor.Name)
 			if jobExecutor.JobType == config.CreationJob {
-				if jobExecutor.Cleanup {
-					log.Info("Cleaning up previous runs")
-					jobExecutor.gc(ctx, nil)
-				}
 				if config.IsChurnEnabled(jobExecutor.Job) {
 					log.Info("Churning enabled")
 					log.Infof("Churn cycles: %v", jobExecutor.ChurnConfig.Cycles)
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -1,0 +1,2 @@
+# External GitHub URLs that get rate-limited (429) during CI checks
+https://github.com/kubevirt/containerized-data-importer/blob/main/doc/datavolumes.md#conditions
--- a/docs/contributing/ai-policy.md
+++ b/docs/contributing/ai-policy.md
@@ -33,7 +33,7 @@ Using LLMs for coding is allowed, but responsibility remains with the contributo
 
 ### Requirements
 
-- Follow all project contribution guidance in [CONTRIBUTING.md](https://github.com/kube-burner/kube-burner/blob/main/CONTRIBUTING.md)
+- Follow all project contribution guidance in [CONTRIBUTING.md](../../CONTRIBUTING.md)
 - Keep changes focused, minimal, and easy to review
 - Match existing style, formatting, and structure
 - Remove unnecessary noise (unused files, generated metadata, unrelated edits)


vishnuchalla:
@jtaleric @afcollins appreciate your feedback here, being primary requesters and users of this feature.

docs/reference/configuration.md
```diff
@@ -144,6 +145,38 @@ Our configuration files strictly follow YAML syntax. To clarify on List and Obje
...
+
+The incremental load feature increases the number of iterations from a configured start (`startIterations`) to a configured total (`totalIterations`) in cumulative fashion. Two growth patterns are supported:
+
+- Linear: iterations increase by a fixed amount each step (configured with `pattern.linear.stepSize`).
+- Exponential: iterations grow multiplicatively using `pattern.exponential.base`. An optional `pattern.exponential.warmupSteps` value can apply a few initial linear increases before exponential growth begins.
+
+After each increase the runner performs the configured health check and will stop early on failure. Between successful steps the runner waits the configured `stepDelay` before applying the next increase.
+
+Simple examples:
+- Linear: `startIterations=10`, `totalIterations=50`, `pattern.linear.stepSize=10` → 10, 20, 30, 40, 50 with constant increments of 10 iterations in each step.
```
ygalblum: As before, this is unclear. What does `10, 20, 30, 40, 50` mean? 
There's no reason to be short here. Explain what would happen (same for exponential)

pkg/prometheus/prometheus.go
```diff
@@ -82,15 +86,15 @@ func (p *Prometheus) ScrapeJobsMetrics(jobList ...Job) error {
 				renderedQuery.Reset()
 				if metric.Instant {
 					if metric.CaptureStart {
-						docsToIndex[metric.MetricName+"-start"] = append(docsToIndex[metric.MetricName+"-start"], p.runInstantQuery(query, metric.MetricName+"-start", jobStart, eachJob)...)
+						docsToIndex[metric.MetricName+"-start"+"_"+uuid] = append(docsToIndex[metric.MetricName+"-start"+"_"+uuid], p.runInstantQuery(query, metric.MetricName+"-start", jobStart, eachJob)...)
```
afcollins: I'm not a fan of adding uuid in the file name. I thought a run's metrics were collected in a dir with the uuid in the file name already? e.g. collected-metrics-{uuid}/{metricsName}

afcollins:
Not seeing code related to it in the PR, but this elapsedTime calculation from the logs doesn't look correct:
```
msg="Job vchalla-test took 2s" file="job.go:203"
```

pkg/prometheus/prometheus.go
afcollins: Oh, I see it's a config-specific setting in the downstream tool. 🤔

afcollins: I would still rather have the uuid in the directory name and constant file names (e.g. 'containerCPU.json') than adding uuid to the basename of the file. Especially since this seems like this change would affect all jobs, not only incremental jobs.

pkg/burner/metadata.go
```diff
@@ -33,6 +33,8 @@ type JobSummary struct {
 	ElapsedTime         float64        `json:"elapsedTime"`
 	AchievedQps         float64        `json:"achievedQps,omitempty"`
 	UUID                string         `json:"uuid"`
+	ParentUUID          string         `json:"parentUUID,omitempty"`
```
rsevilla87: I think this approach can be simplified, the value of UUID can and must the be same across all documents of the benchmark, regardless they belong to the  incremental load stage or not.


Having a new field like `incrementalRunUUID` rather than `parentUUID` can simplify the overall code and making the representation and querying of the data easier. i.e: a user can pull all the metrics of the job by querying using a single uuid, and then it can break it down in stages by using the different values of `incrementalRunUUID`

In addition, with that field we won't need the `IncrementalRun` field, as `incrementalLoadUUID != ""` 
will mark the document as incrementalRun

rsevilla87: wdty @afcollins ?

afcollins: Hmm, I see. So introducing increments as a layer below the UUID instead of one above.  Then all increments will have the same UUID, but we can filter out a specific increment by incrementalLoadUUID should we want to do that.

I think in my own experience, I would still want to query each increment individually, which means analysis logic has to change to look for the new `incrementalLoadUUID`, instead of looking up each distinct UUID.

I can see how your suggestion simplifies and keeps a consistent meaning of "UUID" in the context of a kube-burner run. I am not sure whether this makes analysis easier.

pkg/prometheus/prometheus.go
vishnuchalla: Keeping the metrics structure as is for the sake of simplicity. Relevant dependency PR: https://github.com/cloud-bulldozer/go-commons/pull/63

vishnuchalla:
> Not seeing code related to it in the PR, but this elapsedTime calculation from the logs doesn't look correct:
> 
> ```
> msg="Job vchalla-test took 2s" file="job.go:203"
> ```

Fixed it

pkg/burner/metadata.go
vishnuchalla: Updated as per the suggestions. Please see description for low level details.

pkg/prometheus/prometheus.go
```diff
@@ -160,14 +160,25 @@ func (p *Prometheus) ReadProfile(location string, embedCfg *fileutils.EmbedConfi
 
 // Create metric creates metric to be indexed
 func (p *Prometheus) createMetric(query, metricName string, job Job, labels model.Metric, value model.SampleValue, timestamp time.Time, isInstant bool) metric {
+	uuid := p.UUID
+	if job.UUID != "" {
+		uuid = job.UUID
```
rsevilla87: What is job.UUID?

pkg/prometheus/prometheus.go
```diff
@@ -160,14 +160,25 @@ func (p *Prometheus) ReadProfile(location string, embedCfg *fileutils.EmbedConfi
...
+	}
+	metadata := map[string]any{}
+	for k, v := range p.metadata {
+		metadata[k] = v
+	}
+	if job.IncrementalLoadUUID != "" {
+		metadata["incrementalLoadUUID"] = job.IncrementalLoadUUID
+	}
```
rsevilla87, line 169: Why not
```suggestion
	if job.IncrementalLoadUUID != "" {
		p.metadata["incrementalLoadUUID"] = job.IncrementalLoadUUID
	}
```

?

pkg/burner/job.go
```diff
@@ -97,10 +97,15 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
 		}
 
 		// Iterate job list
-		for jobExecutorIdx, jobExecutor := range jobExecutors {
+		var measurementsInstance *measurements.Measurements
+		var measurementsJobName string
+		for _, jobExecutor := range jobExecutors {
+			jobIdx := len(executedJobs) // Track the index where we're appending
+			startJobIdx := jobIdx       // Will work for both incremental and normal jobs
 			executedJobs = append(executedJobs, prometheus.Job{
```
rsevilla87, line 113: I don't understand this change, why jobExecutorIdx is not valid anymore?

pkg/burner/job.go
```diff
@@ -331,8 +362,16 @@ func indexMetrics(uuid string, executedJobs []prometheus.Job, returnMap map[stri
 			if elapsedTime > 0 {
 				achievedQps = math.Round((float64(job.ObjectOperations)/elapsedTime)*1000) / 1000
 			}
+			if job.JobConfig.IncrementalLoad != nil && job.IncrementalLoadUUID == "" {
+				achievedQps = 0
+			}
+			jsUUID := uuid
+			if job.UUID != "" {
```
rsevilla87: What is job.UUID, why do we need it?

pkg/burner/job.go
vishnuchalla: Because a step job being part of incremental run is also a job and is appended to the list of executed jobs.

pkg/burner/job.go
vishnuchalla: Good catch, not required so removed it.

pkg/prometheus/prometheus.go
vishnuchalla: sorry, was a left over.

pkg/prometheus/prometheus.go
vishnuchalla: When there is no metadata we run into this error
```
time="2026-03-12 13:26:32" level=info msg="Indexing [0] documents from metric top2PrometheusCPU" file="prometheus.go:227"
time="2026-03-12 13:26:32" level=info msg="Indexing skipped due to 0 docs" file="prometheus.go:232"
panic: assignment to entry in nil map

goroutine 63 [running]:
github.com/kube-burner/kube-burner/v2/pkg/prometheus.(*Prometheus).createMetric(_, {_, _}, {_, _}, {{0x38701631, 0xee144eca5, 0x0}, {0x2539b901, 0xee144eca9, ...}, ...}, ...)
	/home/vchalla/myforks/kube-burner/pkg/prometheus/prometheus.go:163 +0xdd
github.com/kube-burner/kube-burner/v2/pkg/prometheus.(*Prometheus).parseVector(_, {_, _}, {_, _}, {{0x38701631, 0xee144eca5, 0x0}, {0x2539b901, 0xee144eca9, ...}, ...}, ...)
	/home/vchalla/myforks/kube-burner/pkg/prometheus/prometheus.go:113 +0x366
github.com/kube-burner/kube-burner/v2/pkg/prometheus.(*Prometheus).runInstantQuery(_, {_, _}, {_, _}, {_, _, _}, {{0x38701631, 0xee144eca5, ...}, ...})
	/home/vchalla/myforks/kube-burner/pkg/prometheus/prometheus.go:201 +0x235
github.com/kube-burner/kube-burner/v2/pkg/prometheus.(*Prometheus).ScrapeJobsMetrics(0xc000494360, {0xc000203008, 0x7, 0x2cd3440?})
	/home/vchalla/myforks/kube-burner/pkg/prometheus/prometheus.go:87 +0xac9
github.com/kube-burner/kube-burner/v2/pkg/burner.indexMetrics({_, _}, {_, _, _}, _, {{0xc0007c5600, 0x2, 0x2}, {0x0, ...}, ...}, ...)
	/home/vchalla/myforks/kube-burner/pkg/burner/job.go:391 +0x77a
github.com/kube-burner/kube-burner/v2/pkg/burner.Run.func1()
	/home/vchalla/myforks/kube-burner/pkg/burner/job.go:284 +0x2b0e
created by github.com/kube-burner/kube-burner/v2/pkg/burner.Run in goroutine 1
	/home/vchalla/myforks/kube-burner/pkg/burner/job.go:89 +0x60c
```

PR #1152, opened 2026-02-17
fix: declare binary dependency for images target

### LINKED ISSUE #1151: Make images does not declare binary dependency
The `images` Makefile target does not depend on the binary it packages.

On a clean checkout, `make images` can fail or use a stale binary, breaking reproducible builds and impacting CI and releases.

**Expected**
`make images` should always build the binary first.

Nesar976:
Declare the binary as a prerequisite of the `images` Makefile target.

This ensures `make images` works on a clean checkout and always packages the binary built from the current source, avoiding stale artifacts and improving build reproducibility.

Fix #1151


vishnuchalla:
Please feel free to join our community to call if you still wish to pursue this change.

PR #1154, opened 2026-02-18
fix: make test-k8s explicit dependency checks

### LINKED ISSUE #1153: [BUG] make test-k8s has implicit dependencies causing failures
Make test-k8s relies on undocumented implicit environment dependencies . Running it in a clean setup results in non-obvious test failures, affecting new contributors and CI reliability . These dependencies should be made explicit or validated before execution.

Nesar976:
make test-k8s now validates all required tools up front and fails fast with clear errors when dependencies are missing. This removes implicit assumptions, improves developer experience, and makes CI behavior more reliable.

Fixes #1153


vishnuchalla:
Closing this as a no-op. Please feel free to join our community call to further discuss about it.

PR #1160, opened 2026-02-21
perf: limit Makefile source discovery to cmd/ and pkg/

### LINKED ISSUE #1159: Makefile: Optimize source discovery to reduce build overhead
The current Makefile source discovery appears to recursively scan the repository when collecting Go sources. This can:
	1.  Slow down every make invocation
	2. Become significantly worse when vendor/ exists
	3. Risk hitting command-line argument limits on larger trees

This does not cause functional breakage, but it introduces avoidable build-time overhead and scalability concerns as the repository grows.

A more bounded and efficient source discovery approach (e.g., excluding vendor/, limiting directories, or using go list ) could improve build performance without changing behavior.

Nesar976:
## Type of change

 - Optimization

Limit Go source discovery in the Makefile to cmd/ and pkg/ instead of recursively scanning the entire repository.

Previously, find . -type f -name "*.go" traversed the full tree, including directories such as vendor/, which adds unnecessary overhead and can lead to argument expansion risks as the repository grows.

This change scopes discovery to the actual Go source directories without altering build behavior or targets.

Fix #1159

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ HARDENED_CGO_LDFLAGS = -linkmode=external -extldflags "-Wl,-z,relro,-z,now"
 
 GIT_COMMIT = $(shell git rev-parse HEAD)
 VERSION ?= $(shell hack/tag_name.sh)
-SOURCES := $(shell find . -type f -name "*.go")
+SOURCES := $(shell find cmd pkg -type f -name "*.go")
 BUILD_DATE = $(shell date '+%Y-%m-%d-%H:%M:%S')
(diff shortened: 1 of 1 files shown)


rsevilla87:
This is not a bug, we don't have vendor folder, and make this change can lead to issues in the future

Nesar976:
Hi @rsevilla87 Thanks for the clarification.
Understood — since there’s no vendor/ directory and this isn’t considered a bug, I agree that restricting discovery to specific directories could introduce maintenance issues if the project structure changes in the future.

PR #1166, opened 2026-02-23
Validate pre-load image pull status

### LINKED ISSUE #1165: [RFE] Harden image pre load to actually verify the image is pulled
Currently the image pre load spins off a daemon set to pull images of interest on every node. However, it doesn't actually verify that images have actually been pulled, instead it just waits for `preLoadPeriod` and cleans up the pre load namespaces. In case there are some connectivity issues to the image registry or the preLoadPeriod is set too low, the process may abort before the images are cached on nodes. This in turn leads to a significant skew in pod latency measurements: the pod ready latency may include the image pull duration on some nodes.

Harden the image pre loading and actually verify that container images are pulled on all nodes before proceeding to the test run stage.

N/A

**Additional context**
This has already caused some false positive regressions in the OCP pipeline - typically it's perceived as a short spike at the very beginning of a test:
Image

avasilevskii:
## Type of change

- Refactor
- Optimization

## Description
This PR improves the image pre-loading process: instead of simply waiting for `preLoadPeriod`, the kube-burner polls pre-load namespace events to make sure that container images are pulled on all nodes before proceeding to the test run stage. Once the number of pulled images hits the desired count, the pre-load is considered finished and the pre-load namespace clean-up starts:
```
time="2026-02-24 14:49:51" level=info msg="Pre-load: Creating DaemonSet using images [quay.io/kubevirt/fedora-with-test-tooling-container-disk:v0.59.2 quay.io/cloud-bulldozer/nginx:latest] in namespace preload-kube-burner-04343" file="pre_load.go:265"
time="2026-02-24 14:49:56" level=debug msg="Pre-load: DaemonSet scheduled on 2 nodes" file="pre_load.go:279"
time="2026-02-24 14:49:56" level=info msg="Pre-load: Waiting for images to be pulled on 2 nodes (timeout 2m0s)" file="pre_load.go:93"
time="2026-02-24 14:49:56" level=debug msg="Pre-load: 4/4 images pulled across 2 nodes" file="pre_load.go:113"
time="2026-02-24 14:49:56" level=info msg="Pre-load: All images pulled on 2 nodes" file="pre_load.go:115"
```
Furthermore, this change potentially enables parallel image pulls by moving pre-load images to main containers section.

**CHANGE OF BEHAVIOR:** If not all DS replicas are ready within `preLoadPeriod`, the test will be terminated! The default timeout value has been increased to 10m.

- Related Issue #1165 
- Closes #1165

### DIFF
--- a/pkg/burner/pre_load.go
+++ b/pkg/burner/pre_load.go
@@ -265,23 +265,21 @@ func createDSs(ctx context.Context, clientSet kubernetes.Interface, imageList []
 	}
 
 	log.Infof("Pre-load: Creating DaemonSet using images %v in namespace %s", imageList, preLoadNs)
-	_, err := clientSet.AppsV1().DaemonSets(preLoadNs).Create(ctx, &ds, metav1.CreateOptions{})
+	created, err := clientSet.AppsV1().DaemonSets(preLoadNs).Create(ctx, &ds, metav1.CreateOptions{})
 	if err != nil {
 		return 0, err
 	}
 	var desired int
 	err = wait.PollUntilContextCancel(ctx, preLoadPollInterval, true, func(ctx context.Context) (bool, error) {
-		dsList, err := clientSet.AppsV1().DaemonSets(preLoadNs).List(ctx, metav1.ListOptions{})
+		ds, err := clientSet.AppsV1().DaemonSets(preLoadNs).Get(ctx, created.Name, metav1.GetOptions{})
 		if err != nil {
 			log.Errorf("Error getting DaemonSet status in %s: %v", preLoadNs, err)
 			return false, nil
 		}
-		if len(dsList.Items) > 0 {
-			if d := int(dsList.Items[0].Status.DesiredNumberScheduled); d > 0 {
-				log.Debugf("Pre-load: DaemonSet scheduled on %d nodes", d)
-				desired = d
-				return true, nil
-			}
+		if d := int(ds.Status.DesiredNumberScheduled); d > 0 {
+			log.Debugf("Pre-load: DaemonSet scheduled on %d nodes", d)
+			desired = d
+			return true, nil
 		}
 		return false, nil
 	})


pkg/burner/pre_load.go
```diff
@@ -75,23 +80,92 @@ func preLoadImages(ctx context.Context, job JobExecutor, clientSet kubernetes.In
...
-	time.Sleep(job.PreLoadPeriod)
+	log.Infof("Pre-load: Waiting for images to be pulled on %d nodes (timeout %v)", desired, job.PreLoadPeriod)
+	if err := waitForImagePull(preloadCtx, clientSet, desired, len(imageList)); err != nil {
+		return fmt.Errorf("pre-load: %v", err)
+	}
 	// 5 minutes should be more than enough to cleanup this namespace
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
 	defer cancel()
-	util.CleanupNamespacesByLabel(ctx, clientSet, "kube-burner-preload=true")
+	util.CleanupNamespacesByLabel(ctx, clientSet, fmt.Sprintf("%s=true", preLoadNs))
```
rsevilla87: I think we can reuse the same context here, so we don't need to declare a new one, we can bump preLoadPeriod to 10m just in case then :) (as we discussed, the preload stage is mandatory and we should make sure it passes to prevent interferences)

pkg/burner/pre_load.go
```diff
@@ -75,23 +80,92 @@ func preLoadImages(ctx context.Context, job JobExecutor, clientSet kubernetes.In
...
 	return nil
 }
 
+func waitForImagePull(ctx context.Context, clientSet kubernetes.Interface, desired, imageCount int) error {
+	expectedTotal := desired * imageCount
+	ticker := time.NewTicker(preLoadPollInterval)
+	defer ticker.Stop()
+	for {
```
rsevilla87: To be inline with other parts of the code and be more k8s friendly, I wonder if we can use `wait.PollUntilContextTimeout()`, (similar to https://github.com/kube-burner/kube-burner/blob/main/pkg/burner/waiters.go#L138) rather than an infinite loop.

pkg/burner/pre_load.go
avasilevskii: Done

pkg/burner/pre_load.go
avasilevskii: Good point!
Since we already have an external context here, I used `wait.PollUntilContextCancel`.

pkg/burner/pre_load.go
```diff
@@ -189,21 +256,37 @@ func createDSs(ctx context.Context, clientSet kubernetes.Interface, imageList []
...
 	log.Infof("Pre-load: Creating DaemonSet using images %v in namespace %s", imageList, preLoadNs)
 	_, err := clientSet.AppsV1().DaemonSets(preLoadNs).Create(ctx, &ds, metav1.CreateOptions{})
 	if err != nil {
-		return err
+		return 0, err
 	}
-	return nil
+	var desired int
+	err = wait.PollUntilContextCancel(ctx, preLoadPollInterval, true, func(ctx context.Context) (bool, error) {
+		dsList, err := clientSet.AppsV1().DaemonSets(preLoadNs).List(ctx, metav1.ListOptions{})
```
rsevilla87: Just a minor nit, can you use Get() instead List(), I know we have to use list because the daemonset doesn't have a fixed name (it uses generateName), but I think we don't need it anymore since the namespace name is now dynamic

avasilevskii: Done

PR #1171, opened 2026-02-27
Fix RunOnce guard when iterationStart is non-zero

### LINKED ISSUE #1145: [BUG] RunOnce is silently skipped when iterationStart is not zero

Nesar976:
RunOnce execution was guarded by i == 0, assuming iterations always start at zero.
When iterationStart is non-zero (e.g., resume or churn paths), the RunOnce block was silently skipped.

This change makes RunOnce state-driven so it executes exactly once per object regardless of the iteration offset.

FIxes #1145

### DIFF
--- a/pkg/burner/create.go
+++ b/pkg/burner/create.go
@@ -159,10 +159,11 @@ func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterati
 			}
 			ex.objects[objectIndex].LabelSelector = kbLabels
 			if obj.RunOnce {
-				if i == 0 {
+				if !obj.hasRun {
 					// this executes only once during the first iteration of an object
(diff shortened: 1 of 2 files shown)


Nesar976:
Hi @rsevilla87 I’ve closed the previous PR and opened a new one that contains only the minimal changes required to fix this issue. The new PR is focused strictly on the RunOnce guard fix with no unrelated modifications. Please let me know if anything else needs adjustment.

rsevilla87:
i can different of 0 when churning, as it takes the value of iterationsStart, which is random in that case, closing this PR

PR #1176, opened 2026-03-02
Churning objects fixes and improvements

rsevilla87:
## Type of change

- Bug fix
- Optimization

## Description

Due to some issues found while running churn with CUDNs, improve the logic to support removing cluster-scoped objects and adding extra logging messages.

As part of this change, update the gc logic to remove cluster-scoped objects before namespaced objects, necessary to effectively garbage collect namespaces selected by CUDNs.

- Related Issue #
- Closes #

### DIFF
--- a/pkg/burner/create.go
+++ b/pkg/burner/create.go
@@ -509,7 +509,7 @@ func (ex *JobExecutor) churnObjects(ctx context.Context) {
 				numToChurn := int(math.Max(float64(ex.ChurnConfig.Percent*len(objectList.Items)/100), 1))
 				randStart := rand.Intn(len(objectList.Items) - numToChurn + 1)
 				objectsToDelete := objectList.Items[randStart : numToChurn+randStart]
-				log.Debugf("Deleting %d %s", numToChurn, obj.Kind)
+				log.Infof("Deleting %d %s", numToChurn, obj.Kind)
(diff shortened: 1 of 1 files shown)


pkg/burner/create.go
```diff
@@ -490,32 +491,38 @@ func (ex *JobExecutor) churnObjects(ctx context.Context) {
...
+				objectList, err = ex.dynamicClient.Resource(obj.gvr).List(ctx, metav1.ListOptions{LabelSelector: labels.FormatLabels(labelSelector)})
 				if err != nil {
 					log.Errorf("Error listing objects: %v", err)
 					continue
 				}
+				log.Debugf("Total %s listed: %d, churning %d%%", obj.Kind, len(objectList.Items), ex.ChurnConfig.Percent)
 				numToChurn := int(math.Max(float64(ex.ChurnConfig.Percent*len(objectList.Items)/100), 1))
 				randStart := rand.Intn(len(objectList.Items) - numToChurn + 1)
 				objectsToDelete := objectList.Items[randStart : numToChurn+randStart]
+				log.Debugf("Deleting %d %s", numToChurn, obj.Kind)
```
vishnuchalla: Rest of the changes look good. Can we make this an Infof? If this was there, CUDN bug would have been a bit simpler to detect

PR #1178, opened 2026-03-02
Add UUID to local-indexing dir

afcollins:
## Type of change

- New feature

## Description

This change ensures the UUID is added to metricsDirectory for local indexing.

If the UUID was already added to path via job config, it is not added again.

Some examples:

```

// UUID added when base directory path is configured
metricsEndpoints:
  - indexer:
      metricsDirectory: /tmp/kube-burner
      type: local

time="2026-03-02 10:43:26" level=info msg="File /tmp/kube-burner-358d4634-d161-4309-9d86-15fb80bd8252/podLatencyMeasurement-kubelet-density.json created with 1 documents" file="base_measurement.go:159"

// When UUID is added through template configuration, it is not added again
metricsEndpoints:
  - indexer:
      metricsDirectory: /tmp/kube-burner-{{.UUID}}
      type: local


time="2026-03-02 10:42:03" level=info msg="File /tmp/kube-burner-91c99ff0-e0c6-4d2c-b02a-d67436e3c8e7/podLatencyMeasurement-kubelet-density.json created with 1 documents" file="base_measurement.go:159"

// When no directory is specified, the default 'collected-metrics' is used, now with uuid appended after a '-'
metricsEndpoints:
  - indexer:
      type: local

time="2026-03-02 10:44:40" level=info msg="File collected-metrics-2b51a62c-6c04-4138-961f-089ebe330782/podLatencyMeasurement-kubelet-density.json created with 1 documents" file="base_measurement.go:159"
```

- Related Issue #
- Closes #

### DIFF
--- a/pkg/util/metrics/metrics.go
+++ b/pkg/util/metrics/metrics.go
@@ -54,9 +54,9 @@ func ProcessMetricsScraperConfig(scraperConfig ScraperConfig) Scraper {
 	}
 	// MetricsEndpoint has preference over the configuration file
 	// if --metrics-directory is set, retain the value that is overwritten by UnmarshalYAML default
-	var passedMetricsDirectory string
+	var metricsDirectoryFlag string
 	if len(scraperConfig.ConfigSpec.MetricsEndpoints) > 0 {
-		passedMetricsDirectory = scraperConfig.ConfigSpec.MetricsEndpoints[0].MetricsDirectory
+		metricsDirectoryFlag = scraperConfig.ConfigSpec.MetricsEndpoints[0].MetricsDirectory
 	}
 	if scraperConfig.MetricsEndpoint != "" {
 		scraperConfig.ConfigSpec.MetricsEndpoints = DecodeMetricsEndpoint(scraperConfig.MetricsEndpoint)
@@ -70,11 +70,13 @@ func ProcessMetricsScraperConfig(scraperConfig ScraperConfig) Scraper {
 				indexerAlias = metricsEndpoint.Alias
 			}
 			if metricsEndpoint.Type == indexers.LocalIndexer {
-				if passedMetricsDirectory != "" {
-					metricsEndpoint.MetricsDirectory = passedMetricsDirectory
-				}
 				if metricsEndpoint.MetricsDirectory == "collected-metrics-{{.UUID}}" {
-					metricsEndpoint.MetricsDirectory = fmt.Sprintf("collected-metrics-%s", scraperConfig.ConfigSpec.GlobalConfig.UUID)
+					if metricsDirectoryFlag != "" && metricsDirectoryFlag != "collected-metrics-{{.UUID}}" {
+						metricsEndpoint.MetricsDirectory = metricsDirectoryFlag
+					} else {
+						// Default value from config.go unmarshalYAML
+						metricsEndpoint.MetricsDirectory = fmt.Sprintf("collected-metrics-%s", scraperConfig.ConfigSpec.GlobalConfig.UUID)
+					}
 				}
 			}
 			log.Infof("📁 Creating %s indexer: %s", metricsEndpoint.Type, indexerAlias)


afcollins:
It's true that we don't need it, but we already append the UUID to the log file name by default. This is an extension of that, to ensure that by default we aren't clobbering the same directory with repeated runs.

This is more to set a standard before other other PRs that might change the local indexing structure.

rsevilla87:
> It's true that we don't need it, but we already append the UUID to the log file name by default. This is an extension of that, to ensure that by default we aren't clobbering the same directory with repeated runs.
> 
> This is more to set a standard before other other PRs that might change the local indexing structure.

I see, I have a suggestion to align with the log file naming format then, why don't we update the default value of metricsDirectory, from`collected-metrics` to `metrics-<uuid>` or `collected-metrics-<uuid>` 

https://github.com/kube-burner/kube-burner/blob/main/pkg/config/config.go#L67

I think that is a more clear and intuitive way to obtain a similar outcome

afcollins:
> I see, I have a suggestion to align with the log file naming format then, why don't we update the default value of metricsDirectory, from`collected-metrics` to `metrics-<uuid>` or `collected-metrics-<uuid>`
> 
> https://github.com/kube-burner/kube-burner/blob/main/pkg/config/config.go#L67
> 
> I think that is a more clear and intuitive way to obtain a similar outcome

Yes, this is where I wanted to make the change, but I did not find the UUID available in that method and wiring it in seemed difficult. Hence, Claude suggested to add in the place I made the change.

vishnuchalla:
> > I see, I have a suggestion to align with the log file naming format then, why don't we update the default value of metricsDirectory, from`collected-metrics` to `metrics-<uuid>` or `collected-metrics-<uuid>`
> > https://github.com/kube-burner/kube-burner/blob/main/pkg/config/config.go#L67
> > I think that is a more clear and intuitive way to obtain a similar outcome
> 
> Yes, this is where I wanted to make the change, but I did not find the UUID available in that method and wiring it in seemed difficult. Hence, Claude suggested to add in the place I made the change.

Why not just have UUID as a global variable in this file and use it unmarshal methods. It's constant anyways.

vishnuchalla:
`golangci-lint run --fix` and `gofmt -w file.go` on all the files please?

afcollins:
It works! Cool. Moved the uuid assignment a little earlier and then we can set it by default. Removed the other changes.  Thanks for the guidance.

> golangci-lint run --fix and gofmt -w file.go on all the files please?

`golangci-lint run --fix` on config.go shows 0 issues.
```
% golangci-lint run --fix                                    
0 issues.
```

Also, this isn't run with `make lint` ? 

`gofmt` doesn't result in any changes.

Let me know if there's something I am missing?

afcollins:
Hmm. Failure on index subcommand. An initial look I see init and index call much of the same code but slightly different. Will look into how to refactor to ensure consistency.

afcollins:
Not sure about taking on a larger refactor. I pushed a simple fix, but it duplicates some logic, but it works with --uuid argument as well. And updated the help output.

with `--uuid`:
```
% $KUBE_BURNER index --start 1764568800 --end 1764569000 -u http://localhost:9090 -m ../.
./metrics-profiles/metrics.yaml --uuid=abc123                                                                                                                                                                                                    
time="2026-03-17 18:24:00" level=info msg="MetricsDirectory collected-metrics-abc123" file="kube-burner.go:382"
time="2026-03-17 18:24:00" level=info msg="📁 Creating local indexer: indexer-0" file="metrics.go:67"
time="2026-03-17 18:24:00" level=info msg="👽 Initializing prometheus client with URL: http://localhost:9090" file="prometheus.go:45"
time="2026-03-17 18:24:00" level=info msg="🔍 Endpoint: http://localhost:9090; profile: ../../metrics-profiles/metrics.yaml start: 2025-11-30T22:00:00-08:00 end: 2025-11-30T22:03:20-08:00; job: kube-burner-indexing, metricsClosing: " file="prometheus.go:67"
time="2026-03-17 18:24:01" level=info msg="Indexing [1348] documents from metric API99thLatency" file="prometheus.go:233"
time="2026-03-17 18:24:01" level=info msg="File collected-metrics-abc123/API99thLatency.json now contains 4044 documents" file="prometheus.go:238"
```

without `--uuid=`:
```
% $KUBE_BURNER index --start 1764568800 --end 1764569000 -u http://localhost:9090 -m ../../metrics-profiles/metrics.yaml
time="2026-03-17 18:24:04" level=info msg="MetricsDirectory collected-metrics-6f55efcb-e324-4cc8-83d8-ba35ae60e6fc" file="kube-burner.go:382"
time="2026-03-17 18:24:04" level=info msg="📁 Creating local indexer: indexer-0" file="metrics.go:67"
time="2026-03-17 18:24:04" level=info msg="👽 Initializing prometheus client with URL: http://localhost:9090" file="prometheus.go:45"
time="2026-03-17 18:24:04" level=info msg="🔍 Endpoint: http://localhost:9090; profile: ../../metrics-profiles/metrics.yaml start: 2025-11-30T22:00:00-08:00 end: 2025-11-30T22:03:20-08:00; job: kube-burner-indexing, metricsClosing: " file="prometheus.go:67"
time="2026-03-17 18:24:04" level=info msg="Indexing [1348] documents from metric API99thLatency" file="prometheus.go:233"
time="2026-03-17 18:24:04" level=info msg="File collected-metrics-6f55efcb-e324-4cc8-83d8-ba35ae60e6fc/API99thLatency.json now contains 1348 documents" file="prometheus.go:238"
```

# The Test Failure

Found a solution. It's a change to the logic but it works for init and index, so I will argue that it is cleaner that way. Updated docs and help as well.

pkg/util/metrics/metrics.go
```diff
@@ -56,6 +57,13 @@ func ProcessMetricsScraperConfig(scraperConfig ScraperConfig) Scraper {
 	if scraperConfig.MetricsEndpoint != "" {
 		scraperConfig.ConfigSpec.MetricsEndpoints = DecodeMetricsEndpoint(scraperConfig.MetricsEndpoint)
 	}
+	for i, ep := range scraperConfig.ConfigSpec.MetricsEndpoints {
```
rsevilla87: At the moment we don't support variables within in metricsEndpoints files, I think is better to remove this to be more consistent.

We can work on full variable support in a follow-up pr

afcollins: Ok. It's not a variable from the template though. It only comes from config.go:67  and for some reason although GlobalConfig.UUID is set it is not being picked up, even though UnmarshallYAML is being called from indexCmd.

It passes all the tests.

I'm not sure how else to wire it in 🤔 

So it is not being defined in MetricsEndpoint file, but if the command has both --metrics-directory and -e , then the absent field in the endpoint file is causing the default to be chosen. Do we want the command argument to have the highest priority?

afcollins: So, I took this to mean to not do templating with a general ReplaceAll at this time.

Because of the default value from UnmarshallYAML overriding the command-line option, I updated it to retain this value.

Added test and output here to show this behavior: https://gist.github.com/afcollins/75215061c7f239a3a556de8ab7d9f693

cmd/kube-burner/kube-burner.go
```diff
@@ -377,11 +377,11 @@ func indexCmd() *cobra.Command {
...
 			} else {
 				indexer.IndexerConfig = indexers.IndexerConfig{
-					Type:             indexers.LocalIndexer,
-					MetricsDirectory: metricsDirectory,
-					TarballName:      tarballName,
+					Type:        indexers.LocalIndexer,
+					TarballName: tarballName,
 				}
 			}
+			indexer.MetricsDirectory = metricsDirectory
```
afcollins, line 384: The field is available on Indexer and not specific to IndexerConfig. When --es-server is set, this is skipped, which fails this [test case](https://github.com/kube-burner/kube-burner/blob/main/test/test-k8s.bats#L168) that has "all the flags".

This change preserves the test-expected behavior.

pkg/util/metrics/metrics.go
afcollins: Now I am more confused.. Can you clarify what you mean by "we don't support variables within metricsEndpoints files"? I see them in the test metrics-endpoints: https://github.com/kube-burner/kube-burner/blob/main/test/k8s/metrics-endpoints.yaml#L8-L15

Edit: I understand these are env substitutions that happen from yaml decoder.

PR #1193, opened 2026-03-19
Add global hooks support for before/after all jobs execution

venkataanil:
Allow running hooks that are not tied to any specific job. Global hooks are configured in the 'global' section and execute before any job starts or after all jobs complete.

This helps networking workloads to setup test envrionment (like external server, port forwarding, frr) before running the jobs and avoid over dependency on measurement code to acheieve the same


rsevilla87 (commented):
Why do we need this?

are beforeJobExecution and afterJobExecution hooks not enough?

venkataanil:
> Why do we need this?
> 
> are beforeJobExecution and afterJobExecution hooks not enough?

As you have obseved, all networking tests uses multiple jobs and resources created in previous jobs are used in later jobs. If we want to have some scripts which should create some resources at the start of the test and then tear down at the end of the test, job based hoos not helpful.

rsevilla87:
> > Why do we need this?
> > are beforeJobExecution and afterJobExecution hooks not enough?
> 
> As you have obseved, all networking tests uses multiple jobs and resources created in previous jobs are used in later jobs. If we want to have some scripts which should create some resources at the start of the test and then tear down at the end of the test, job based hoos not helpful.

Then why don't you use _hooks.beforeJobExecution_ in the first job and _hooks.afterJobExecution_ in the last one. For your particular case, we don't need these code changes IIUC.

venkataanil:
> > > Why do we need this?
> > > are beforeJobExecution and afterJobExecution hooks not enough?
> > 
> > 
> > As you have obseved, all networking tests uses multiple jobs and resources created in previous jobs are used in later jobs. If we want to have some scripts which should create some resources at the start of the test and then tear down at the end of the test, job based hoos not helpful.
> 
> Then why don't you use _hooks.beforeJobExecution_ in the first job and _hooks.afterJobExecution_ in the last one. For your particular case, we don't need these code changes IIUC.

That is one solution I can go with but already the code in the network workload templates became very complex (many jobs, interdependency between jobs, complex go templating) and users have to track the jobs and be in the context. This is like a global and simple setup() and tearDown() kind of concepts for the entire test run.

docs/reference/configuration.md
```diff
@@ -405,6 +409,15 @@ Hooks are configured as a list under the `hooks` field in a job:
 
 The `when` field specifies at which stage the hook should execute:
 
+**Global Hook Stages** (configured in `global.hooks`):
+
+| Stage                    | Description                                           |
+|--------------------------|-------------------------------------------------------|
+| `beforeAllJobs`          | Before any job starts (after image preloading)        |
+| `afterAllJobs`           | After all jobs complete and metrics are indexed       |
```
```suggestion
| `setup`          | Before any job or measurement starts        |
| `teardown`           | Called after the benchmark execution regardless of test success or failure       |
```

docs/reference/configuration.md
```diff
@@ -442,41 +455,80 @@ The `when` field specifies at which stage the hook should execute:
...
 #### Example Configuration
 
+**Global hooks example:**
+
+```yaml
+global:
+  hooks:
+    # Run setup script before any job starts
+    - cmd: ["/scripts/cluster-setup.sh"]
+      when: beforeAllJobs
```
venkataanil, line 587: @rsevilla87 do you suggest to replace beforeAllJobs in this PR (in all files) with setup and afterAllJobs with teardown?

rsevilla87: yeah, I think that sounds better.

pkg/burner/job.go
```diff
@@ -91,6 +91,17 @@ func Run(configSpec config.Spec, kubeClientProvider *config.KubeClientProvider,
 		_, restConfig := kubeClientProvider.DefaultClientSet()
 		measurementsFactory := measurements.NewMeasurementsFactory(configSpec, metricsScraper.MetricsMetadata, additionalMeasurementFactoryMap)
 		jobExecutors = newExecutorList(configSpec, kubeClientProvider, embedCfg)
+
```
vishnuchalla, line 94: Should this be after preload?

docs/reference/configuration.md
vishnuchalla: @venkataanil changes look good except for this comment. Mind doing a quick update? so that we can get this in

vishnuchalla:
@venkataanil mind doing a quick update on this PR?

vishnuchalla:
Closing this as stale. Had a follow raised already: https://github.com/kube-burner/kube-burner/pull/1304

PR #1206, opened 2026-04-08
Fix image preload for VMIs

rsevilla87:
## Type of change

- Bug fix

## Description

Image preload wasn't working correctly for VMIs, it was timing out as demonstrated at https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/periodic-ci-openshift-eng-ocp-qe-perfscale-ci-main-metal-4.22-nightly-x86-daily-virt-6nodes/2041788597949960192/artifacts/daily-virt-6nodes/openshift-qe-virt-density-nomount/build-log.txt

This is because the ImageID is not being set (the _image_ field does not show these issues), I'm not sure why at the moment, it's possible that the cause is the image format used by kubevirt doesn't allow that.

Also adding an override command to the preloaded images in the DS to prevent running the default container entrypoint.

- Related Issue #
- Closes #

### DIFF
--- a/pkg/burner/pre_load.go
+++ b/pkg/burner/pre_load.go
@@ -153,7 +153,7 @@ func countPulledImages(ctx context.Context, clientSet kubernetes.Interface, seen
 				continue
 			}
 			pullCount++
-			if status.ImageID != "" {
+			if status.Image != "" {
 				seen[podName+"/"+status.Name] = struct{}{}
 				pulledCount++
 			}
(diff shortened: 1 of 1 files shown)


mukrishn (approved):
/lgtm

pkg/burner/pre_load.go
```diff
@@ -270,6 +270,7 @@ func createDSs(ctx context.Context, clientSet kubernetes.Interface, imageList []
 		ds.Spec.Template.Spec.Containers = append(ds.Spec.Template.Spec.Containers, corev1.Container{
 			Name:            fmt.Sprintf("pull-%d", i),
 			Image:           image,
+			Command:         []string{"override", "command"},
```
vishnuchalla, line 273: Why do we need this?

rsevilla87: To prevent running the default container entrypoint, which is something that shouldn't be executed w/o control

PR #1214, opened 2026-04-17
fix(errors): update EnhanceYAMLParseError test assertions to match format

### LINKED ISSUE #1212: [BUG] `EnhanceYAMLParseError` unit test fails on main format string mismatch
# Bug Description

```
Version: main
Go Version: go1.26.1
OS/Arch: darwin arm64
```

## **Describe the bug**

I was setting up the project locally and ran the tests, and noticed that `TestErrors` in `pkg/errors` is failing on the current `main` branch. The issue is in the `EnhanceYAMLParseError` test case for `yaml.TypeError`.

Looks like the format string in `errors.go` (line 33) was updated at some point but the test wasn't updated along with it two things are off:

- The function now outputs `"failed to parse config file: config.yml: \n..."` (with colons), but the test still checks for `"failed to parse config file config.yml"` (no colons)
- The function joins multiple errors using newlines (`\n`), while the test checks for semicolons (`; `)

## **To Reproduce**

2. Run:
   ```bash
   go test ./pkg/errors/... -v -run TestErrors
   ```
3. Test fails with 5 passed, 1 failed

## **Expected behavior**

The test should pass. I think the test assertions just need to be updated to match the current function output, but happy to go either way if the maintainers prefer changing the function format instead.

## **Screenshots or output**

```
• [FAILED] [0.000 seconds]
Error Enhancement EnhanceYAMLParseError when error is yaml.TypeError [It] should enhance with line and column information

  [FAILED] Expected
      <string>: failed to parse config file: config.yml: 
      line 5: invalid syntax
      line 10: unknown field
  to contain substring
      <string>: failed to parse config file config.yml
  In [It] at: pkg/errors/errors_test.go:48

Ran 6 of 6 Specs in 0.001 seconds
FAIL! -- 5 Passed | 1 Failed | 0 Pending | 0 Skipped
```

## **Additional context**

Here's where the mismatch is:

`pkg/errors/errors.go` line 33:
```go
return fmt.Errorf("failed to parse config file: %s: \n%s", filename, strings.Join(errorDetails, "\n"))
```

`pkg/errors/errors_test.go` lines 48-49:
```go
Expect(result.Error()).Should(ContainSubstring("failed to parse config file config.yml"))
Expect(result.Error()).Should(ContainSubstring("line 5: invalid syntax; line 10: unknown field"))
```

I can open a PR to fix this just let me know if you'd prefer updating the test to match the current format, or changing the function output back to match the test expectations.

Abhishek9639:
## Type of change

- [x] Bug fix

## Description

Fixes the `EnhanceYAMLParseError` test case that was failing on `main`. 

The format string in `errors.go` (line 33) was updated previously to include colons and join errors with newlines, but the test assertions were never updated to reflect this change. The test was still looking for the old format (no colons) and expected the errors to be joined with semicolons.

This PR updates the test assertions to match the actual function output, adding the missing colons and splitting the error checks so they work correctly with the new newline separator.

- Closes #1212

### DIFF
--- a/pkg/errors/errors_test.go
+++ b/pkg/errors/errors_test.go
@@ -45,8 +45,9 @@ var _ = Describe("Error Enhancement", func() {
 				result := errors.EnhanceYAMLParseError("config.yml", yamlErr)
 
 				Expect(result).To(HaveOccurred())
-				Expect(result.Error()).Should(ContainSubstring("failed to parse config file config.yml"))
-				Expect(result.Error()).Should(ContainSubstring("line 5: invalid syntax; line 10: unknown field"))
+				Expect(result.Error()).Should(ContainSubstring("failed to parse config file: config.yml:"))
+				Expect(result.Error()).Should(ContainSubstring("line 5: invalid syntax"))
+				Expect(result.Error()).Should(ContainSubstring("line 10: unknown field"))
 			})
 		})
 


Abhishek9639:
Hey @ygalblum,
PR is up for this. Just updated the test assertions to match the current function output. 
Let me know if it looks good.
Thanks

ygalblum (approved):

ygalblum:
Thanks

PR #1219, opened 2026-04-18
fix: nil pointer dereference in NewIterationCalculator when pattern config is missing

Abhishek9639:
## Type of change
 
- [x] Bug fix

## Description

I was reading through the incremental load code and noticed that `NewIterationCalculator` in `pkg/burner/incremental.go` would panic at startup in two places if the pattern sub-block is missing from the config.

For the exponential pattern, `MaxIncrease` and `WarmupSteps` were being read from `cfg.Pattern.Exponential` before the nil check so if someone sets `pattern.type: exponential` without an `exponential:` block, kube-burner crashes immediately. Fixed by moving the nil check first and using safe defaults (`base=2.0`, `maxIncrease=0`, `warmupSteps=0`).

For the linear pattern, the nil guard only wrapped `StepSize` but `cfg.Pattern.Linear.MinSteps` was accessed just outside it, causing the same crash when the `linear:` block is absent. Extended the guard to cover the full calculation block.

Also fixed a stale test assertion in `pkg/errors/errors_test.go` that was already failing on `main` the format string in `EnhanceYAMLParseError` had been updated but the test expectation wasn't kept in sync with it.

- No upstream issue tracked for this yet discovered while reading the codebase.

### DIFF
--- a/pkg/burner/incremental.go
+++ b/pkg/burner/incremental.go
@@ -53,24 +53,30 @@ func NewIterationCalculator(ex JobExecutor) IterationCalculator {
 	}
 	if cfg.Pattern.Type == config.ExponentialPattern {
 		base := 2.0
-		maxInc := cfg.Pattern.Exponential.MaxIncrease
-		warmup := cfg.Pattern.Exponential.WarmupSteps
-		if cfg.Pattern.Exponential != nil && cfg.Pattern.Exponential.Base > 0 {
-			base = cfg.Pattern.Exponential.Base
+		maxInc := 0
+		warmup := 0
+		if cfg.Pattern.Exponential != nil {
+			if cfg.Pattern.Exponential.Base > 0 {
+				base = cfg.Pattern.Exponential.Base
+			}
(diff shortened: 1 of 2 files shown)


Abhishek9639:
Hi @ygalblum,
Just wanted to flag this for review when you get a chance. It's a small fix moved some nil checks that were in the wrong order in NewIterationCalculator. 
Happy to make any changes if needed
Thanks

Abhishek9639:
Hi @ygalblum,
The check-docs failure is a transient 504 Gateway Timeout from GitHub's servers on pre-existing links in the docs not related to this PR at all (we only changed pkg/burner/incremental.go and pkg/errors/errors_test.go). 
Could a maintainer re-run the check-docs job? 
Thanks

Abhishek9639:
Hi @ygalblum,
image

vishnuchalla:

Abhishek9639:
Hey @vishnuchalla
Thanks

PR #1221, opened 2026-04-19
fix: guard nil Linear/Exponential config in NewIterationCalculator

### LINKED ISSUE #1220: [BUG] Nil pointer panic in `NewIterationCalculator` when linear pattern config is missing
# Bug Description

```
Version: main
Go Version: go1.26.1
OS/Arch: darwin arm64
```

## **Describe the bug**

While going through the incremental load code in `pkg/burner/incremental.go`, I found what looks like another nil pointer panic in the linear pattern branch of `NewIterationCalculator`.

If `pattern.type` is `linear` (or just not `exponential`, since the linear path is the `else` branch) but the `linear` config block is omitted, `cfg.Pattern.Linear` will be nil. The code has a nil guard, but it only covers the `StepSize` field access. The `MinSteps` field is read two lines later, completely outside the guard:

```go
// pkg/burner/incremental.go (simplified)
} else {
    step := 1
    if cfg.Pattern.Linear != nil && cfg.Pattern.Linear.StepSize > 0 {
        step = cfg.Pattern.Linear.StepSize
    }
    totalSteps := int(math.Ceil(float64(totalIt-startIt) / float64(step)))
    if cfg.Pattern.Linear.MinSteps > 0 && totalSteps < cfg.Pattern.Linear.MinSteps { // panics here if Linear is nil
        remaining := totalIt - startIt
        if remaining <= 0 {
            step = totalIt
        } else {
            step = int(math.Ceil(float64(remaining) / float64(cfg.Pattern.Linear.MinSteps)))
        }
    }
}
```

So the nil check on `cfg.Pattern.Linear` guards `StepSize`, but then `MinSteps` is dereferenced unconditionally right after. If someone omits the `linear` block entirely, it crashes at that second access.

## **To Reproduce**

2. Create a job config with `pattern.type: linear` but without the `linear` sub-block:
   ```yaml
   incrementalLoad:
     startIterations: 10
     totalIterations: 100
     pattern:
       type: linear
       # no linear block here
   ```
3. Run kube-burner with this config
4. It panics at startup during calculator initialization

## **Expected behavior**

If the `linear` block is absent, kube-burner should fall back to a default step size of `1` and skip the `MinSteps` adjustment entirely. A nil dereference crash for a missing optional config block is unexpected the field is marked `omitempty` in the struct tags, so omitting it should be a valid use case.

## **Screenshots or output**

I traced this through the code rather than running it end-to-end (spinning up a cluster just to hit this would be overkill), but based on the code path it would panic at the `MinSteps` dereference with something like:

```
goroutine 1 [running]:
github.com/kube-burner/kube-burner/v2/pkg/burner.NewIterationCalculator(...)
        pkg/burner/incremental.go:73
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x...]

goroutine 1 [running]:
runtime/debug.Stack()
        /usr/local/go/src/runtime/debug/stack.go:24 +0x5b
github.com/kube-burner/kube-burner/v2/pkg/burner.NewIterationCalculator(0xc0001c4000)
        /home/user/kube-burner/pkg/burner/incremental.go:73 +0x2a1
exit status 2
```

## **Additional context**

One way to handle this would be moving the `MinSteps` access inside the existing nil guard alongside `StepSize`, something like:

```go
} else {
    step := 1
    if cfg.Pattern.Linear != nil {
        if cfg.Pattern.Linear.StepSize > 0 {
            step = cfg.Pattern.Linear.StepSize
        }
        totalSteps := int(math.Ceil(float64(totalIt-startIt) / float64(step)))
        if cfg.Pattern.Linear.MinSteps > 0 && totalSteps < cfg.Pattern.Linear.MinSteps {
            remaining := totalIt - startIt
            if remaining <= 0 {
                step = totalIt
            } else {
                step = int(math.Ceil(float64(remaining) / float64(cfg.Pattern.Linear.MinSteps)))
            }
        }
    }
    if step <= 0 {
        step = 1
    }
    return &linearCalculator{start: startIt, total: totalIt, step: step}
}
```

This is roughly the same pattern I'd expect for the exponential branch. There's a related issue in the exponential path (filed separately) where a similar nil guard ordering problem exists might make sense to fix both in one PR if that's easier.

Happy to open a PR if this looks like the right direction.

Abhishek9639:
Fixes #1220

## Type of change
- [x] Bug fix

## Description

While reviewing the incremental load logic in `pkg/burner/incremental.go`, I noticed that `cfg.Pattern.Linear.MinSteps` was being accessed outside the nil guard that protects `StepSize`. If a user sets `pattern.type: linear` but omits the `linear` sub-block entirely, `cfg.Pattern.Linear` is nil and the code panics at the `MinSteps` dereference on startup.

The fix moves `totalSteps` and the `MinSteps` check inside the `cfg.Pattern.Linear != nil` block, so omitting the config safely falls back to `step=1` with no crash.

I also fixed the same ordering issue in the exponential branch, where `MaxIncrease` and `WarmupSteps` were read before the nil check as mentioned in the issue, it made sense to fix both in one PR.

Added unit tests covering nil config for both patterns, normal StepSize behaviour, MinSteps recalculation, and the zero-total edge case.

- Related issue #1220
- Closes #1220

### DIFF
--- a/pkg/burner/incremental.go
+++ b/pkg/burner/incremental.go
@@ -53,24 +53,29 @@ func NewIterationCalculator(ex JobExecutor) IterationCalculator {
 	}
 	if cfg.Pattern.Type == config.ExponentialPattern {
 		base := 2.0
-		maxInc := cfg.Pattern.Exponential.MaxIncrease
-		warmup := cfg.Pattern.Exponential.WarmupSteps
-		if cfg.Pattern.Exponential != nil && cfg.Pattern.Exponential.Base > 0 {
-			base = cfg.Pattern.Exponential.Base
+		var maxInc, warmup int
+		if cfg.Pattern.Exponential != nil {
+			if cfg.Pattern.Exponential.Base > 0 {
+				base = cfg.Pattern.Exponential.Base
+			}
+			maxInc = cfg.Pattern.Exponential.MaxIncrease
+			warmup = cfg.Pattern.Exponential.WarmupSteps
 		}
 		return &exponentialCalculator{start: startIt, total: totalIt, base: base, maxIncrease: maxInc, warmup: warmup, stepNo: 0}
 	} else {
 		step := 1
 		if cfg.Pattern.Linear != nil {
-			step = cfg.Pattern.Linear.StepSize
-		}
-		totalSteps := int(math.Ceil(float64(totalIt-startIt) / float64(step)))
-		if cfg.Pattern.Linear.MinSteps > 0 && totalSteps < cfg.Pattern.Linear.MinSteps {
-			remaining := totalIt - startIt
-			if remaining <= 0 {
-				step = totalIt
-			} else {
-				step = int(math.Ceil(float64(remaining) / float64(cfg.Pattern.Linear.MinSteps)))
+			if cfg.Pattern.Linear.StepSize > 0 {
+				step = cfg.Pattern.Linear.StepSize
+			}
+			totalSteps := int(math.Ceil(float64(totalIt-startIt) / float64(step)))
+			if cfg.Pattern.Linear.MinSteps > 0 && totalSteps < cfg.Pattern.Linear.MinSteps {
+				remaining := totalIt - startIt
+				if remaining <= 0 {
+					step = totalIt
+				} else {
+					step = int(math.Ceil(float64(remaining) / float64(cfg.Pattern.Linear.MinSteps)))
+				}
 			}
 		}
 		if step <= 0 {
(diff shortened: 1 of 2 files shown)


Abhishek9639:
Hi @ygalblum or @rsevilla87,
Spotted this while reading through the incremental load code. The nil guard was there but only half-covered `MinSteps` was still unprotected. Bundled the exponential fix too since it was the same issue. Tests added for both.
Thanks

Abhishek9639:
Hi @ygalblum,
image

vishnuchalla:

Abhishek9639:
Hey @vishnuchalla
Thanks

PR #1222, opened 2026-04-19
fix: guard nil exponential and linear config in NewIterationCalculator

### LINKED ISSUE #1217: [BUG] Nil pointer panic in `NewIterationCalculator` when exponential pattern config is missing
# Bug Description

```
Version: main
Go Version: go1.26.1
OS/Arch: darwin arm64
```

## **Describe the bug**

I was reading through the incremental load code and spotted what looks like a runtime panic waiting to happen in `NewIterationCalculator` inside `pkg/burner/incremental.go`.

If you set `pattern.type: exponential` but don't include the `exponential` config block, `cfg.Pattern.Exponential` will be nil. The problem is that lines 56-57 already dereference it to read `MaxIncrease` and `WarmupSteps`, and the nil check only shows up on line 58 by which point the crash has already occurred. The guard is essentially unreachable in the nil case.

```go
// pkg/burner/incremental.go lines 54-61
if cfg.Pattern.Type == config.ExponentialPattern {
    base := 2.0
    maxInc := cfg.Pattern.Exponential.MaxIncrease  // panics here if Exponential is nil
    warmup := cfg.Pattern.Exponential.WarmupSteps  // and here
    if cfg.Pattern.Exponential != nil && cfg.Pattern.Exponential.Base > 0 {
        base = cfg.Pattern.Exponential.Base
    }
    ...
}
```

Took me a moment to notice because the nil check is right there just a couple of lines too late.

## **To Reproduce**

2. Create a job config with `pattern.type: exponential` but without the `exponential` sub-block:
   ```yaml
   incrementalLoad:
     startIterations: 10
     totalIterations: 100
     pattern:
       type: exponential
       # no exponential block here
   ```
3. Run kube-burner with this config
4. It panics immediately at startup

## **Expected behavior**

If the `exponential` block is missing, kube-burner should either use safe defaults (`base=2.0`, `maxIncrease=0`, `warmup=0`) or return a validation error upfront. A silent nil dereference crash isn't great UX here.

## **Screenshots or output**

I confirmed this by reading the code rather than running it end-to-end (setting up a full cluster just to trigger this felt overkill), but based on the code path it would panic at `incremental.go:56` with something like:

```
goroutine 1 [running]:
github.com/kube-burner/kube-burner/v2/pkg/burner.NewIterationCalculator(...)
        pkg/burner/incremental.go:56
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x...]

goroutine 1 [running]:
runtime/debug.Stack()
        /usr/local/go/src/runtime/debug/stack.go:24 +0x5b
github.com/kube-burner/kube-burner/v2/pkg/burner.NewIterationCalculator(0xc0001c4000)
        /home/user/kube-burner/pkg/burner/incremental.go:56 +0x1a4
exit status 2
```

## **Additional context**

One way to fix this would be moving the nil check above the field reads:

```go
if cfg.Pattern.Type == config.ExponentialPattern {
    base := 2.0
    maxInc := 0
    warmup := 0
    if cfg.Pattern.Exponential != nil {
        if cfg.Pattern.Exponential.Base > 0 {
            base = cfg.Pattern.Exponential.Base
        }
        maxInc = cfg.Pattern.Exponential.MaxIncrease
        warmup = cfg.Pattern.Exponential.WarmupSteps
    }
    return &exponentialCalculator{start: startIt, total: totalIt, base: base, maxIncrease: maxInc, warmup: warmup, stepNo: 0}
}
```

There's also a similar issue a few lines down in the linear branch `cfg.Pattern.Linear.MinSteps` on line 68 is accessed outside the nil guard that wraps lines 64-66. Might make sense to fix both at once if you agree.

Happy to open a PR if this looks right to you.

Abhishek9639:
Fixes #1217

## Type of change

- Bug fix

## Description

`NewIterationCalculator` in `pkg/burner/incremental.go` had two nil pointer dereference panics:

1. **Exponential branch** - `cfg.Pattern.Exponential.MaxIncrease` and `.WarmupSteps` were read before the nil check, so setting `pattern.type: exponential` without an `exponential:` sub-block caused an immediate runtime panic.

2. **Linear branch** - `cfg.Pattern.Linear.MinSteps` was accessed outside the `if cfg.Pattern.Linear != nil` guard, triggering the same panic when the `linear:` sub-block was absent.

Both are fixed by setting safe defaults (`base=2.0`, `maxIncrease=0`, `warmup=0`, `step=1`) and reading all config fields only inside their respective nil guards. Unit tests are added using Ginkgo/Gomega covering nil configs, full configs, zero-base fallback, MinSteps adjustment, and empty pattern type.

- Closes #1217

### DIFF
--- a/pkg/burner/incremental.go
+++ b/pkg/burner/incremental.go
@@ -53,24 +53,30 @@ func NewIterationCalculator(ex JobExecutor) IterationCalculator {
 	}
 	if cfg.Pattern.Type == config.ExponentialPattern {
 		base := 2.0
-		maxInc := cfg.Pattern.Exponential.MaxIncrease
-		warmup := cfg.Pattern.Exponential.WarmupSteps
-		if cfg.Pattern.Exponential != nil && cfg.Pattern.Exponential.Base > 0 {
-			base = cfg.Pattern.Exponential.Base
+		maxInc := 0
+		warmup := 0
+		if cfg.Pattern.Exponential != nil {
+			if cfg.Pattern.Exponential.Base > 0 {
+				base = cfg.Pattern.Exponential.Base
+			}
+			maxInc = cfg.Pattern.Exponential.MaxIncrease
+			warmup = cfg.Pattern.Exponential.WarmupSteps
 		}
 		return &exponentialCalculator{start: startIt, total: totalIt, base: base, maxIncrease: maxInc, warmup: warmup, stepNo: 0}
 	} else {
 		step := 1
 		if cfg.Pattern.Linear != nil {
-			step = cfg.Pattern.Linear.StepSize
-		}
-		totalSteps := int(math.Ceil(float64(totalIt-startIt) / float64(step)))
-		if cfg.Pattern.Linear.MinSteps > 0 && totalSteps < cfg.Pattern.Linear.MinSteps {
-			remaining := totalIt - startIt
-			if remaining <= 0 {
-				step = totalIt
-			} else {
-				step = int(math.Ceil(float64(remaining) / float64(cfg.Pattern.Linear.MinSteps)))
+			if cfg.Pattern.Linear.StepSize > 0 {
+				step = cfg.Pattern.Linear.StepSize
+			}
+			totalSteps := int(math.Ceil(float64(totalIt-startIt) / float64(step)))
+			if cfg.Pattern.Linear.MinSteps > 0 && totalSteps < cfg.Pattern.Linear.MinSteps {
+				remaining := totalIt - startIt
+				if remaining <= 0 {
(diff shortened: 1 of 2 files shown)


Abhishek9639:
Hey @ygalblum,
Would love to get a review on this when you get a chance. Pretty straightforward nil guard fix happy to adjust anything if 
the approach doesn't look right to you. 
Thanks

Abhishek9639:
Hi @ygalblum,
image

vishnuchalla:

Abhishek9639:
Hey @vishnuchalla
Thanks

PR #1225, opened 2026-04-22
fix(measurements): gracefully ignore uninitialized timestamps to avoid false 0ms latency ingestion

Kayd-06:
## Type of change

- [x] Bug fix

## Description

Fixes a silent mathematical underflow vulnerability across the `pkg/measurements/` metric suite that was inadvertently ingesting `0ms` metrics for unused/untracked workflow phases.

**Why it's needed:**
When a Kube-burner test suite runs through phases (such as `ContainersReady`, `DVBound`, `Scheduling`, etc.), if a phase completes instantaneously or if an underlying volume/pod condition is bypassed, the phase's tracking timestamps remain uninitialized (`0001-01-01`).

When evaluating metric generation, `.Sub(m.Timestamp)` is called against these uninitialized phases, which previously resulted in a massive negative number underflow. The existing handler caught this `<0` logic and explicitly clamped the result to exactly `0ms`. This ultimately forced downstream reporting logic (such as Elasticsearch) to skew heavily by treating completely absent deployment routines as having taken `0ms`. 

**Implementation Details:**
- Checks `m.[timestamp].IsZero()` (or explicit `0` int64 structures) globally inside `normalizeMetrics()`.
- Uninitialized phases are now cleanly bypassed to a discrete `-1` value.
- `getLatency()` metric mappings explicitly strip `-1` latency phases, ensuring they are entirely ignored during Elasticsearch dictionary indexing rather than outputting false `0ms` fields.

**Modified Components:**
- `vmim_latency.go`
- `pod_latency.go`
- `datavolume_latency.go`
- `pvc_latency.go`
- `job_latency.go`

### DIFF
--- a/pkg/measurements/datavolume_latency.go
+++ b/pkg/measurements/datavolume_latency.go
@@ -264,26 +264,38 @@ func (dv *dvLatency) normalizeMetrics() float64 {
 		// v2 latencies are currently under AB testing which blindly trust kubernetes as source of
 		// truth and will prevent us from those over 1s delays as well as <0 cases.
 		errorFlag := 0
-		m.DVBoundLatency = int(m.dvBound.Sub(m.Timestamp).Milliseconds())
-		if m.DVBoundLatency < 0 {
-			log.Tracef("DVBoundLatency for DataVolume %v falling under negative case. So explicitly setting it to 0", m.Name)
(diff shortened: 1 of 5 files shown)


vishnuchalla:

PR #1226, opened 2026-04-23
Fix nil pointer dereference in parseMatrix churn metric tagging

Abhishek9639:
## Type of change

- Bug fix

## Description

This PR fixes a potential nil pointer dereference panic in `pkg/alerting/alert_manager.go`.

When checking if an alert falls within the churn window, the logic was verifying that `churnStart != nil` but was omitting the check for `churnEnd != nil` before dereferencing `*churnEnd`. If `churnEnd` happens to be `nil` while `churnStart` is not, the application would panic at runtime.

This issue has been resolved by adding `churnEnd != nil` to the conditional check. I've also added a comprehensive Ginkgo test suite for `parseMatrix` to ensure the logic works properly under various `nil` scenarios and to prevent future regressions.

### DIFF
--- a/pkg/alerting/alert_manager.go
+++ b/pkg/alerting/alert_manager.go
@@ -197,7 +197,7 @@ func parseMatrix(value model.Value, uuid, description string, metadata any, seve
 				Description: renderedDesc.String(),
 				MetricName:  alertMetricName,
 			}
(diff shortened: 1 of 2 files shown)


Abhishek9639:
Hey @ygalblum,
PR is up for review. Let me know if the test coverage looks good to you or if you'd like to see any other scenarios covered.
Thanks

vishnuchalla (approved):
lgtm

PR #1236, opened 2026-05-06
[RFE] Add repeatEveryNIterations config option (#1241)

venkataanil:
## Type of change

- New feature

[RFE] https://github.com/kube-burner/kube-burner/issues/1241 

## Description

This feature allows cluster-scoped objects to be shared across multiple
namespaces, which is useful for resources like EgressIP, PersistentVolume,
ClusterRole, ClusterUserDefinedNetwork, and MultiNetworkPolicy.

When namespacesPerObject=N, the object is created once per N iterations.

Why not use multiple jobs?
--------------------------
While cluster-scoped objects could be created in a separate job, this
breaks critical single-job features:

1. Churn: Namespace churning deletes and recreates namespaces along with
   their associated cluster-scoped objects. With separate jobs, the churn
   job cannot manage the lifecycle of objects created by another job.

2. Incremental Load: Gradually scaling up iterations requires coordinated
   creation of both namespaced and cluster-scoped objects. Separate jobs
   cannot synchronize this scaling.

3. Cleanup/GC: Garbage collection operates per-job. Cluster-scoped objects
   in a separate job won't be cleaned up with their associated namespaces.

With namespacesPerObject, everything stays in one job, preserving churn,
incremental load, verification, and cleanup behavior.

Example: With namespacesPerObject=2 and jobIterations=10:
  - EgressIP created at iterations: 0, 2, 4, 6, 8 (5 total)
  - Namespaces 0,1 share egressip-0; namespaces 2,3 share egressip-1; etc.
  - Churn correctly deletes/recreates namespace pairs with their EgressIP

Changes:
- Add NamespacesPerObject field to Object struct with default value of 1
- Add buildTemplateData() helper for consistent template variable handling
- Add validation to ensure consistent namespacesPerObject values within a job
- Update object creation to skip iterations based on namespacesPerObject
- Align namespace churn to namespacesPerObject boundaries to prevent
  partial deletion of shared objects
- Add deleteSharedObjectsForIterations() for proper cleanup during churn
- Update Verify() to calculate expected object count correctly
- Update preload to use the same template variables
- Add integration test for the feature

 Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>
 Signed-off-by: venkataanil <vkommadi@redhat.com>

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -239,7 +239,7 @@ Each object element supports the following parameters:
 | `wait`                 | Wait for object to be ready                                       | Boolean | true    |
 | `waitOptions`          | Customize [how to wait](#object-wait-options) for object to be ready     | Object  | {}       |
 | `runOnce`              | Create or delete this object only once during the entire job    | Boolean | false   |
-| `SharingNamespacesCount`  | Controls how many namespaces share a single instance of an object. See [SharingNamespacesCount](#SharingNamespacesCount)  | Integer | 1   |
+| `repeatEveryNIterations`  | Controls how often to create an object (once per N iterations). See [repeatEveryNIterations](#repeatEveryNIterations)  | Integer | 1   |
 
 !!! warning
     Kube-burner is only able to wait for a subset of resources, unless `waitOptions` are specified.
@@ -267,18 +267,19 @@ The following object types have built-in waiters:
 !!! info
     Find more info about the waiters implementation in the `pkg/burner/waiters.go` file
 
-### SharingNamespacesCount
+### repeatEveryNIterations
 
-The `SharingNamespacesCount` parameter controls how many namespaces share a single instance of a cluster-scoped object. This is useful for cluster-scoped objects (like `ClusterRoles` or `PersistentVolume`) that should be shared across multiple namespaces.
-Without this, each iteration creates a namespace and a cluster-scoped resource. Thus we have one-to-one mapping for the namespace and cluster-scoped resource i.e a cluster-scoped resource is not shared across the namespaces. When SharingNamespacesCount is set for a cluster-scoped resource, an iteration creates a namespace, skips creating cluster-scoped resource if it is created in the previous iterations and then the namespace scoped object templates in that iteration has to use the corresponding previously created cluster-scoped resource, resulting the sharing of the cluster-scoped resource.
+The `repeatEveryNIterations` parameter controls how often an object is created - once per N iterations instead of every iteration. This is useful for objects (cluster-scoped or namespaced) that should be shared across multiple iterations.
 
-As some iterations skip creating the cluster scoped resource, next iterations need to adjust the .Iteration var while passing to the template. For example, in the below table, iteration 1 skips creating ClusterRole and then in iteration 2 adjusts the .Iteration to 1 so that clusterrole-{.Iteration} will be rendered as clusterrole-1
+Without this, each iteration creates every object. With `repeatEveryNIterations` set, an object is created only when the iteration number is a multiple of N. Other iterations skip creating that object and can reference the previously created one.
 
-The main intention of bringing this feature is to fit cluster-scoped object and the namespaces when shared into a sing job, so that churn and incremental features work. kube-burner adjusts churn boundaries and churn starting iterations according to the mapping between the cluster scoped resource and namespaces.
+As some iterations skip creating the object, the `.Iteration` template variable is adjusted accordingly. For example, with `repeatEveryNIterations: 2`, iteration 1 skips creating the object and uses `.Iteration = 0`, so `clusterrole-{{.Iteration}}` renders as `clusterrole-0`.
 
-Below table illustrates what kube-burner does in each iteration to create ClusterRole (cluster-scoped object) and RoleBinding, when a ClusterRole is shared among 2 RoleBinding (or namepsaces).
+This feature enables sharing objects across iterations within a single job, allowing churn and incremental features to work correctly. kube-burner adjusts churn boundaries and starting iterations according to the repeat interval.
 
-With SharingNamespacesCount: 2 set for ClusterRole (a cluster scoped object) and jobIterations: 10:
+Below table illustrates what kube-burner does in each iteration to create ClusterRole and RoleBinding, when a ClusterRole is shared among 2 RoleBindings.
+
+With repeatEveryNIterations: 2 set for ClusterRole and jobIterations: 10:
 
 | Iteration | Namespace | ClusterRole Created | RoleBinding Created |
 |--------------|---------------------------------------------------------|---------|---------|
@@ -302,18 +303,18 @@ This feature is helpful for:
 #### Template variables
 
 ```yaml
-  # sharing-namespaces-count-test.yml
+  # repeat-every-n-iterations-test.yml
   jobs:
-    - name: sharing-namespaces-count-test
+    - name: repeat-every-n-iterations-test
       namespace: test-ns
       objects:
-        # ClusterRole created once per 2 namespaces (shared across test-ns-0/test-ns-1, test-ns-2/test-ns-3)
+        # ClusterRole created once per 2 iterations (shared across test-ns-0/test-ns-1, test-ns-2/test-ns-3)
         - objectTemplate: objectTemplates/clusterrole.yml
-          SharingNamespacesCount: 2
+          repeatEveryNIterations: 2
         # RoleBinding in each namespace, referencing the shared ClusterRole
         - objectTemplate: objectTemplates/rolebinding.yml
           inputVars:
-            clusterRoleSnc: 2 # Must match ClusterRole's SharingNamespacesCount
+            repeatN: 2 # Must match ClusterRole's repeatEveryNIterations
             namespace: test-ns
 
   # rolebinding.yml
@@ -324,9 +325,9 @@ This feature is helpful for:
   roleRef:
     apiGroup: rbac.authorization.k8s.io
     kind: ClusterRole
-    # Calculate: Iteration / clusterRoleSnc
-    # e.g., iteration 1 with clusterRoleSnc=2 → 1/2 = 0 → clusterrole-0
-    name: clusterrole-{{ div .Iteration .clusterRoleSnc }} # Same for test-ns-0 and test-ns-1
+    # Calculate: Iteration / repeatN
+    # e.g., iteration 1 with repeatN=2 → 1/2 = 0 → clusterrole-0
+    name: clusterrole-{{ div .Iteration .repeatN }} # Same for test-ns-0 and test-ns-1
   subjects:
     - kind: ServiceAccount
       name: default
@@ -346,10 +347,11 @@ This feature is helpful for:
 
 #### Validation
 
-All objects in a job must have consistent SharingNamespacesCount values:
-  - All objects can have SharingNamespacesCount: 1 (default)
+All objects in a job must have consistent repeatEveryNIterations values:
+  - All objects can have repeatEveryNIterations: 1 (default)
   - Or all non-default values must be the same (e.g., all 2)
-  - Mixing SharingNamespacesCount: 2 and SharingNamespacesCount: 3 is not allowed
+  - Mixing repeatEveryNIterations: 2 and repeatEveryNIterations: 3 is not allowed
+  - repeatEveryNIterations > 1 cannot be used with churn mode `objects`. Use churn mode `namespaces` instead
 
 ### Object wait Options
 
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -239,6 +239,7 @@ Each object element supports the following parameters:
 | `wait`                 | Wait for object to be ready                                       | Boolean | true    |
 | `waitOptions`          | Customize [how to wait](#object-wait-options) for object to be ready     | Object  | {}       |
 | `runOnce`              | Create or delete this object only once during the entire job    | Boolean | false   |
+| `namespacesPerObject`  | Controls how many namespaces share a single instance of an object. See [namespacesPerObject](#namespacesPerObject)  | Integer | 1   |
 
 !!! warning
     Kube-burner is only able to wait for a subset of resources, unless `waitOptions` are specified.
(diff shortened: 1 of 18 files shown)


rsevilla87:
I don't understand this change, regardless of the implementation details, the same can be done by declaring multiple jobs.

venkataanil:
> I don't understand this change, regardless of the implementation details, the same can be done by declaring multiple jobs.

I have re-worked on this PR. Initially I have created this PR for "stages" feature. But I removed code related to stages feature in this PR in favor of @vishnuchalla https://github.com/kube-burner/kube-burner/pull/1238 

This PR now has only "namespacesPerObject" feature.

This is very useful for resources like EgressIP, PersistentVolume,
ClusterRole, ClusterUserDefinedNetwork, and MultiNetworkPolicy.

I have added why "namespacesPerObject" feature is required in the commit message now i.e

```
Why not multiple jobs?
While cluster-scoped objects could be created in a separate job, this
breaks critical single-job features:

Churn: Namespace churning deletes and recreates namespaces along with
their associated cluster-scoped objects. With separate jobs, the churn
job cannot manage the lifecycle of objects created by another job.

Incremental Load: Gradually scaling up iterations requires coordinated
creation of both namespaced and cluster-scoped objects. Separate jobs
cannot synchronize this scaling.

Cleanup/GC: Garbage collection operates per-job. Cluster-scoped objects
in a separate job won't be cleaned up with their associated namespaces.

With namespacesPerObject, everything stays in one job, preserving churn,
incremental load, verification, and cleanup behavior.
```

For CUDN churn feaure, @mohit-sheth needed to add lot of code in https://github.com/kube-burner/kube-burner-ocp/pull/436 and also a change in https://github.com/kube-burner/kube-burner/pull/1240

When I need to implement churn or incremental load for MultiNetworkPolicy https://github.com/kube-burner/kube-burner-ocp/pull/194 I need to also go through similar complex changes.

Till now we are creating same number of cluster-scoped objects as the job iterations if it is part of the same job. This allows sharing the object across namespaces with single job.

@rsevilla87 @vishnuchalla @mohit-sheth @jtaleric fyi ^

venkataanil:
As we have a test added, we can test the feature like below

export USE_EXISTING_CLUSTER=yes
export KUBECONFIG=/root/mno/kubeconfig

KUBE_BURNER=/root/test2/kube-burner/bin/amd64/kube-burner bats --filter "namespacesPerObject" -F pretty -T --show-output-of-passing-tests test-k8s.bats

pkg/burner/create.go
```diff
@@ -463,6 +569,8 @@ func (ex *JobExecutor) churnNamespaces(ctx context.Context) []error {
 		}
 		// 1 hour timeout to delete namespace
 		cleanupCtx, cancel := context.WithTimeout(context.Background(), time.Hour)
+		// Delete shared cluster-scoped objects that serve the churned namespaces
+		ex.deleteSharedObjectsForIterations(cleanupCtx, randStart, numToChurn+randStart)
```
venkataanil: This deletes CUDNs and namespaces in parallel, so we can avoid 
https://github.com/kube-burner/kube-burner-ocp/pull/436/changes#diff-17f50a1adc4b3d8988e1cdb993b2adfa16f864c1116d15898834c5c5c9bddc3aR61

docs/reference/configuration.md
```diff
@@ -266,6 +267,147 @@ The following object types have built-in waiters:
...
+      objects:
+        # One PV per 2 namespaces
+        - objectTemplate: persistentvolume.yml
+          namespacesPerObject: 2
+
+        # Each namespace gets its own PVC referencing the shared PV
+        - objectTemplate: persistentvolumeclaim.yml
+          inputVars:
+            pvNamespacesPerObject: 2
+            namespace: pv-test
```
rsevilla87: This example makes no sense, you can get a similar behavior with just 2 iterations and no namespacesPerObject stuff

venkataanil:
@rsevilla87 @vishnuchalla This PR addresses a kube-burner limitation with cluster scoped resources for churn and incremental processing. 

Without this, a cluster scoped resource is tied to a single namespace i.e in 1:1 ratio. The purpose of cluster scoped resource is to be shared among namespaces, however we are restricting one cluster scoped resource to one namespace. It doesn't make sense to have workloads where a cluster scoped resources is not shared among namespaces.

Without this, user has to split cluster scoped resource into another job with different iterations (forced to use 2 jobs, one job for cluster scoped resources and another for namespace scoped resoruces) when a cluster scoped resource is shared across namespaces. 

For incremental load, user has to write a custom script which iterates through these 2 jobs ( @jtaleric can you please share your script which you are using for incremental load).

For churn, user has to:
1) add 4 jobs i.e 
    a) 2 jobs for creation at the start
    b) and then additional 2 jobs for churn https://github.com/kube-burner/kube-burner-ocp/pull/436/changes#diff-05ea7a77d4e30ed7c4e79f071ecec8516905182a7e5062a0c61feadd994d41d3
2) along  with a custom churn logic which is similar to what we have in kube-burner https://github.com/kube-burner/kube-burner-ocp/pull/436/changes#diff-1a9c80eb6a49d5cd4440269ef1a45e631ce1f9a6f49e897eeceaaeb99c039576R134 
3) additionally some changes in kube-burner https://github.com/kube-burner/kube-burner/pull/1240 to support this outside kube-burner code 

When other workloads like multi network policy, egress IP and PVC needs to test churn & incremental processing, they also need to come up with similar additional complex scripts.

Its better to address this limitation natively in kube-burner instead of going through work around complex scripts.

The code mainly 
1) during creation, adjusts iteration number to the cluster scoped object when it is shared
2) during churn, algins starting iteration and number of iterations according to the namespacesPerObject

docs/reference/configuration.md
venkataanil: Intention is to share PV among PVC.  User has to properly define persistentvolumeclaim.yml to share a PV among PVC.

You can run test/k8s/kube-burner-namespaces-per-object.yml which shares a cluster scoped test/k8s/objectTemplates/clusterrole.yml among namespaces using test/k8s/objectTemplates/rolebinding.yml
This test  has churn enabled, so you can test churn and cluster scoped resource sharing.

docs/reference/configuration.md
```diff
@@ -239,6 +239,7 @@ Each object element supports the following parameters:
 | `wait`                 | Wait for object to be ready                                       | Boolean | true    |
 | `waitOptions`          | Customize [how to wait](#object-wait-options) for object to be ready     | Object  | {}       |
 | `runOnce`              | Create or delete this object only once during the entire job    | Boolean | false   |
+| `namespacesPerObject`  | Controls how many namespaces share a single instance of an object. See [namespacesPerObject](#namespacesPerObject)  | Integer | 1   |
```
vishnuchalla: How about calling this sharingNamespaces or sharedByNamespaces or SharingNamespacesCount? Something that clarifies that the object is being shared by namespaces instead of having namespaces assigned to the object.

docs/reference/configuration.md
```diff
@@ -266,6 +267,147 @@ The following object types have built-in waiters:
 !!! info
     Find more info about the waiters implementation in the `pkg/burner/waiters.go` file
 
+### namespacesPerObject
```
vishnuchalla: Also we need to clarify that iterations doesn’t apply to these shared resources

docs/reference/configuration.md
```diff
@@ -266,6 +267,147 @@ The following object types have built-in waiters:
...
+      resources: ["pods"]
+      verbs: ["get"]
+```
+
+#### Validation
+
+All objects in a job must have consistent namespacesPerObject values:
+  - All objects can have namespacesPerObject: 1 (default)
+  - Or all non-default values must be the same (e.g., all 2)
+  - Mixing namespacesPerObject: 2 and namespacesPerObject: 3 is not allowed
```
vishnuchalla: Then in this case, do you think we should move this under group options or to the job level? This enforcement can be error prone for users.

pkg/burner/create.go
```diff
@@ -52,6 +52,50 @@ type churnDeletedObject struct {
 	gvr    schema.GroupVersionResource
 }
 
+// getMaxNamespacesPerObject returns the maximum NamespacesPerObject value across all objects
+func (ex *JobExecutor) getMaxNamespacesPerObject() int {
```
vishnuchalla: This wouldn’t be required if we move it to job level or group level

pkg/burner/create.go
```diff
@@ -172,6 +216,13 @@ func (ex *JobExecutor) RunCreateJob(ctx context.Context, iterationStart, iterati
 					log.Debugf("RunOnce set to %s, so creating object once", obj.ObjectTemplate)
 					ex.replicaHandler(ctx, kbLabels, obj, ns, i, &wg)
 				}
+			} else if obj.NamespacesPerObject > 1 {
+				// Only create when iteration is a multiple of namespacesPerObject
```
vishnuchalla: We should have a validation check for this, IMHO

test/k8s/kube-burner-namespaces-per-object.yml
```diff
@@ -0,0 +1,31 @@
...
+    verifyObjects: true
+    errorOnVerify: true
+    churnConfig:
+      duration: 10s
+      percent: 20
+      delay: 2s
+      mode: namespaces
+      
+    objects:
+      # ClusterRole created once per 2 namespaces (shared across npo-test-0/npo-test-1, npo-test-2/npo-test-3)
```
vishnuchalla: Explaination in the docs should be sufficient to understand this feature. Let’s remove these comments.

pkg/burner/create.go
```diff
@@ -423,6 +474,35 @@ func (ex *JobExecutor) churnNamespaces(ctx context.Context) []error {
 	}
 	numToChurn := int(math.Max(float64(ex.ChurnConfig.Percent*len(jobNamespaces.Items)/100), 1))
 
+	// Get max namespacesPerObject for alignment
```
vishnuchalla: Same here on the comments. Looks too much wordy.

pkg/burner/pre_load.go
```diff
@@ -192,7 +192,24 @@ func isOwnedByDaemonSet(ownerReferences []metav1.OwnerReference, daemonSetName s
 func getJobImages(job JobExecutor) ([]string, error) {
 	var imageList []string
 	for _, object := range job.objects {
-		renderedObj, err := util.RenderTemplate(object.objectSpec, object.InputVars, util.MissingKeyZero, job.functionTemplates)
+		// Build template data with same variables as normal rendering
+		npo := object.NamespacesPerObject
```
vishnuchalla: Not sure why preload is even involved in these changes. Why do we need to adjust the job iteration to 0 here?

pkg/burner/create.go
```diff
@@ -423,6 +474,35 @@ func (ex *JobExecutor) churnNamespaces(ctx context.Context) []error {
...
+	//       - etc.
+	//
+	//   If you churn 3 namespaces (ns-0, ns-1, ns-2), you'd delete:
+	//     - Complete pair for egressIP-0 (ns-0 + ns-1) ✓
+	//     - Half of egressIP-1 (only ns-2, but ns-3 survives) ✗ BROKEN!
+	//
+	//   So we round up numToChurn from 3 to 4, ensuring complete pairs are churned.
+	if maxNpo > 1 {
+		// Align numToChurn to be a multiple of maxNpo (round up)
+		numToChurn = ((numToChurn + maxNpo - 1) / maxNpo) * maxNpo
```
vishnuchalla: To simplify this, can we have this logic modularised into a function?

docs/reference/configuration.md
venkataanil: SharingNamespacesCount makes sense :) I will use that.

docs/reference/configuration.md
venkataanil: sure

docs/reference/configuration.md
venkataanil: No. 
Grouing is required for only CUDN as we need an orderly creation for CUDN. For all other shared cluster scoped resources like multinetworkpolicy, egress ip, pvc, clusterrole we don't need grouping/order. 

Regarding moving it to job, assume if the user wants only one cluster role and 2 cudn. To acheive this with this PR, he won't define namespacesPerObject for cluster role and set it to 2 for cudn. So I want it to be defined as part of the object.

We need more math to allow 2 namespacesPerObject for cudn and 3 for clusterrole. So for simplicity I kept this as a restriction and a validation for now.

pkg/burner/create.go
venkataanil: Same explanation https://github.com/kube-burner/kube-burner/pull/1236/changes#r3296171770

pkg/burner/create.go
venkataanil: can you please explain?

pkg/burner/create.go
venkataanil: I will move it to a function.

pkg/burner/create.go
venkataanil: Should I remove the comment?

pkg/burner/pre_load.go
venkataanil: preload has failed when I ran the test, so I have added it. I will relook into it.

test/k8s/kube-burner-namespaces-per-object.yml
venkataanil: Sure vishnu

pkg/burner/create.go
venkataanil: I looked into moving to a new function. However the code to be part of the new function will be small and not re-used as well. So not going with the new function.

pkg/burner/pre_load.go
rsevilla87, line 1: I don't think we need to modify this file

pkg/burner/create.go
```diff
@@ -52,6 +52,50 @@ type churnDeletedObject struct {
 	gvr    schema.GroupVersionResource
 }
 
+// getMaxSharingNamespacesCountAcrossAllObjects returns the maximum SharingNamespacesCount value across all objects
+func (ex *JobExecutor) getMaxSharingNamespacesCountAcrossAllObjects() int {
```
rsevilla87: This check is not needed, we're already checking that all objects in the the job have the same the same sharingNamespacesCount at validateSharingNamespacesCount()

pkg/burner/create.go
rsevilla87: This functionality should also work with workloads with objects with `SharingNamespacesCount = 1`, 

Please use the function namespaces.CleanupNonNamespacedResourcesByLabel() rather than adding a new one that basically mimics that one, this function receives a the object and labelSelctor as argument

pkg/burner/create.go
```diff
@@ -52,6 +52,50 @@ type churnDeletedObject struct {
...
+		}
+	}
+	return maxSnc
+}
+
+// deleteSharedObjectsForIterations deletes cluster-scoped shared objects
+// that were created for iterations in the given range
+func (ex *JobExecutor) deleteSharedObjectsForIterations(ctx context.Context, iterationStart, iterationEnd int) {
+	for _, obj := range ex.objects {
+		if obj.SharingNamespacesCount <= 1 {
```
rsevilla87: As I noted in the calling function, we should allow churning cluster-scoped objects with `SharingNamespacesCount = 1`

docs/reference/configuration.md
```diff
@@ -239,6 +239,7 @@ Each object element supports the following parameters:
 | `wait`                 | Wait for object to be ready                                       | Boolean | true    |
 | `waitOptions`          | Customize [how to wait](#object-wait-options) for object to be ready     | Object  | {}       |
 | `runOnce`              | Create or delete this object only once during the entire job    | Boolean | false   |
+| `SharingNamespacesCount`  | Controls how many namespaces share a single instance of an object. See [SharingNamespacesCount](#SharingNamespacesCount)  | Integer | 1   |
```
```suggestion
| `sharingNamespacesCount`  | Controls how many namespaces share a single instance of an object. See [SharingNamespacesCount](#SharingNamespacesCount)  | Integer | 1   |
```

rsevilla87: I don't like much the variable name, I find it confusing, objects are not actually "shared" across namespaces. Can you replace by something like `repeatEveryNIterations`

pkg/burner/create.go
venkataanil: sure

docs/reference/configuration.md
venkataanil: @rsevilla87 @vishnuchalla similar to iterationsPerNamespace, can we rename it to iterationsPerObject?

iterationsPerNamespace: N iterations share 1 namespace
iterationsPerObject: N iterations share 1 cluster-scoped object

venkataanil: Cluade gave me these alternatives along with iterationsPerObject

  - objectCreationInterval - "create object at intervals of N iterations"
  - iterationStride - technical but clear
  - namespaceGroupSize - if thinking namespace-centric

pkg/burner/create.go
```diff
@@ -329,7 +406,7 @@ func (ex *JobExecutor) createRequest(ctx context.Context, gvr schema.GroupVersio
 		if ns != "" {
 			uns, err = ex.dynamicClient.Resource(gvr).Namespace(ns).Create(ctx, obj, metav1.CreateOptions{})
 		} else {
-			if !ex.nsChurning {
+			if !ex.nsChurning || allowClusterScopedChurn {
```
rsevilla87: We want to support cluster-scoped objects churning regardless they have sharingNamespacesCount >1 or not

mohit-sheth (dismissed):
Tested this PR, works well
thanks @venkataanil

rsevilla87 (approved):

PR #1237, opened 2026-05-06
Add label-based skip for pod latency measurement

venkataanil:
Introduce kube-burner.io/skip-pod-latency label to exclude specific pods from pod latency tracking. Pods with this label set to "true" will tracked by the podLatency measurement.

### DIFF
--- a/pkg/measurements/pod_latency.go
+++ b/pkg/measurements/pod_latency.go
@@ -115,6 +115,11 @@ func (p *podLatency) handleCreatePod(obj any) {
 		return
 	}
 	podLabels := pod.GetLabels()
+	// Skip pod latency tracking if the skip label is set
+	if value, ok := podLabels[config.KubeBurnerLabelSkipPodLatency]; ok && value == "true" {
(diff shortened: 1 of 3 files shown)


rsevilla87 (approved):
lgtm

pkg/measurements/pod_latency.go
```diff
@@ -115,6 +115,11 @@ func (p *podLatency) handleCreatePod(obj any) {
 		return
 	}
 	podLabels := pod.GetLabels()
+	// Skip pod latency tracking if the skip label is set
+	if value, ok := podLabels[config.KubeBurnerLabelSkipPodLatency]; ok && value == "true" {
```
rsevilla87, line 119: Goconst linter is complaning, I find overkill to define a constant just for this, but honestly it seems the more elegant way. wdyt?

vishnuchalla (changes requested):
I like this idea, but i think this should be implemented in a generic way so that not only for pods it applies to any other resource where we want to skip the measurements.

pkg/measurements/pod_latency.go
afcollins: I like having a constant to keep the labels in a single place together

PR #1239, opened 2026-05-07
Prevent execution error on high measurement error rate

rsevilla87:
## Type of change

- Optimization

## Description

The user can debug latency errors afterwards, kube-burner shouldn't fail the execution when observing them as is not a tool issue.

This behavior can prevent the execution of other subsequent steps like additional tests or regression checking.

- Related Issue #
- Closes #

### DIFF
--- a/pkg/measurements/base_measurement.go
+++ b/pkg/measurements/base_measurement.go
@@ -125,7 +125,7 @@ func (bm *BaseMeasurement) StopMeasurement(normalizeMetrics func() float64, getL
 		log.Infof("%s: %v 99th: %v ms max: %v ms avg: %v ms", bm.JobConfig.Name, pq.QuantileName, pq.P99, pq.Max, pq.Avg)
 	}
 	if errorRate > 10.00 {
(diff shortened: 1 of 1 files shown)


pkg/measurements/base_measurement.go
```diff
@@ -125,7 +125,7 @@ func (bm *BaseMeasurement) StopMeasurement(normalizeMetrics func() float64, getL
 		log.Infof("%s: %v 99th: %v ms max: %v ms avg: %v ms", bm.JobConfig.Name, pq.QuantileName, pq.P99, pq.Max, pq.Avg)
 	}
 	if errorRate > 10.00 {
-		return fmt.Errorf("something is wrong with system under test. %v latencies error rate was: %.2f", bm.MeasurementName, errorRate)
+		log.Warnf("%v latencies error rate was: %.2f", bm.MeasurementName, errorRate)
```
vishnuchalla, line 128: Are we sure about this? It has helped us identify issues sometimes. With this switch to warning we might end up overlooking some benchmarks.

vishnuchalla (approved):
lgtm

PR #1243, opened 2026-05-18
Fix pre-load falsely reporting images as pulled

mcornea:
## Type of change

- Bug fix

## Description

The pre-load check used status.Image to determine whether a container image had been pulled. This field is populated by the kubelet as soon as the container status entry is created, even while the image is still being pulled (Waiting/ContainerCreating), causing pre-load to declare success within seconds for multi-GB images.

Replace the check with isImagePulled(), which inspects container state instead: Running, Terminated, or RestartCount > 0 means the image was pulled. For KubeVirt container disk images where the container command fails at create time (CreateContainerError), the Waiting reason itself proves the pull succeeded, distinguishing it from image-pull-related reasons (ContainerCreating, ErrImagePull, ImagePullBackOff).

### DIFF
--- a/pkg/burner/pre_load.go
+++ b/pkg/burner/pre_load.go
@@ -153,7 +153,7 @@ func countPulledImages(ctx context.Context, clientSet kubernetes.Interface, seen
 				continue
 			}
 			pullCount++
-			if status.Image != "" {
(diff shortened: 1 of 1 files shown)


vishnuchalla (approved):
lgtm

PR #1248, opened 2026-05-28
Optimize deletion code by using DeleteCollection

rsevilla87:
## Type of change

- Optimization

## Description

By using `DeleteCollection` we can optmizt the code, avoiding Listing and Deleting objects

- Related Issue #
- Closes #


vishnuchalla (approved):
lgtm. Verified the [usage](https://github.com/kubernetes/kubernetes/blob/2f5e22e8bb66836f767715d99a83662250968bf5/staging/src/k8s.io/client-go/dynamic/simple.go#L216)

PR #1249, opened 2026-05-28
Remove achieved qps

rsevilla87:
## Type of change

- New feature

## Description

This functionality doesn't work well and  has demonstrated not to be very useful as is.

- Related Issue #
- Closes #


vishnuchalla (approved):
+1 on this. Can be removed as it is not adding much value.

PR #1252, opened 2026-05-28
fix(burner): prevent panic in churnObjects with empty object list

Golevka2001:
## Type of change
- Bug fix

## Description
When namespace deletion is blocked by admission webhooks (e.g., Kyverno policies), leftover objects from previous runs may not match the expected label selector. This causes churnObjects to receive an empty object list, leading to a slice bounds out of range panic when calculating deletion percentages with the formula `math.Max(percent*0/100, 1)`.

This change adds a guard clause to check if the object list is empty before attempting any slice operations. When no objects are found matching the label selector, the churn cycle is gracefully skipped with a warning log instead of panicking.

### DIFF
--- a/pkg/burner/create.go
+++ b/pkg/burner/create.go
@@ -524,6 +524,10 @@ func (ex *JobExecutor) churnObjects(ctx context.Context) {
 					return timeI.Before(&timeJ)
 				})
 				log.Debugf("Total %s listed: %d, churning %d%%", obj.gvr.Resource, len(objectList.Items), ex.ChurnConfig.Percent)
+				if len(objectList.Items) == 0 {
+					log.Warnf("No %s found with label selector %s, skipping churn cycle", obj.gvr.Resource, labels.FormatLabels(labelSelector))
+					continue
+				}
 				percent := int(math.Max(float64(ex.ChurnConfig.Percent*len(objectList.Items)/100), 1))
 				objectsToDelete := objectList.Items[:percent]
 				log.Infof("Deleting %d %s", len(objectsToDelete), obj.gvr.Resource)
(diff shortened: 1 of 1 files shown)


Golevka2001:
## Steps to Reproduce

### 1. Install Kyverno

Install Kyverno in the cluster:

```bash
kubectl create -f https://github.com/kyverno/kyverno/releases/download/v1.16.2/install.yaml
```

### 2. Create a Namespace Deletion Blocking Policy

Create a Kyverno ClusterPolicy that blocks namespace deletion.

[block-namespace-deletion.yaml](https://github.com/user-attachments/files/28354558/block-namespace-deletion.yaml)

```bash
kubectl apply -f block-namespace-deletion.yaml
```

### 3. Run kube-burner Without the Policy (Optional)

First, run kube-burner successfully to create the initial namespace and objects. This step ensures objects are created cleanly on the first run.

[test.yaml](https://github.com/user-attachments/files/28354615/test.yaml)

[pod.yaml](https://github.com/user-attachments/files/28354623/pod.yaml)


```bash
kube-burner init -c test.yaml
```

### 4. Reproduce the Panic

Run kube-burner again. This time, the GC phase will fail to delete the namespace due to the Kyverno policy, leaving residual objects. The subsequent job creation will encounter "already exists" errors, and the churn phase will panic when trying to slice an empty object list.

```bash
kube-burner init -c test.yaml
```

### Expected Behavior

The command should panic with:

```
panic: runtime error: slice bounds out of range [:1] with capacity 0
```

The panic occurs in `pkg/burner/create.go:528` in the `churnObjects` function when attempting to slice `objectList.Items[:1]` on an empty list.

### Cleanup

To clean up after testing:

```bash
kubectl delete clusterpolicy block-namespace-deletion
kubectl delete namespace test-ns --force --grace-period=0
kubectl delete -f https://github.com/kyverno/kyverno/releases/download/v1.16.2/install.yaml
```

Golevka2001:
[panic.log](https://github.com/user-attachments/files/28354838/panic.log)

rsevilla87 (approved):
Thanks for the fix!,lgtm

PR #1258, opened 2026-06-19
Fix broken refs in udn-density-l3 example

afcollins:
## Type of change

- Bug fix

## Description

Example structure differs from the config file.


afcollins:
This one could probably use more love to sync with the downstream workload, to be more upstream-friendly

cc @mohit-sheth

vishnuchalla:
@afcollins mind doing a quick rebase?

vishnuchalla (approved):
lgtm

PR #1262, opened 2026-06-25
Defer cluster-scoped objects garbage collection

### LINKED ISSUE #1260: [BUG] improve GC code to clean up namespaces and cluster scoped objects in parallel
# Bug Description

If we see CUDN templates https://github.com/kube-burner/kube-burner-ocp/blob/main/cmd/config/cudn-density/cudn-density.yml#L213 , we are defining a new job for deleting CUDNs. 

```go-template
{{ if .GC }}
  # Job 4: Delete workload namespaces (kills pods, NADs start terminating)
  # Uses waitForDeletion: false because namespaces will be stuck in Terminating
  # until CUDNs are deleted in Job 5 (NAD finalizers block namespace deletion).
  - name: cudn-density-cleanup-namespaces
    jobType: delete
    waitForDeletion: false
    objects:
      - kind: Namespace
        apiVersion: v1
        labelSelector: {kube-burner.io/job: cudn-density-create-namespaces}

  # Job 5: Delete CUDNs (releases NAD finalizers, namespaces finish terminating)
  - name: cudn-density-cleanup-cudns
    jobType: delete
    waitForDeletion: true
    objects:
      - kind: ClusterUserDefinedNetwork
        apiVersion: k8s.ovn.org/v1
        labelSelector: {kube-burner.io/job: {{ $cudnJobName }} }
{{ end }}

``` 

The requirement is to delete the CUDN (cluster scoped object) and namespace in parallel. We can enhance our GC code like we have in our churn code to delete both in parallel

https://github.com/kube-burner/kube-burner/blob/main/pkg/burner/create.go#L884-L885

rsevilla87:
This pull request refactors the resource cleanup logic in the `JobExecutor`'s `gc` method to improve the order and clarity of resource deletion. The main change is separating the deletion of resources with predefined namespaces from cluster-scoped resources, and ensuring that resources with predefined namespaces are deleted first.

Resource cleanup logic refactor:

* Changed the order of deletion in the `gc` method of `JobExecutor` to delete resources with predefined namespaces before deleting cluster-scoped resources, improving clarity and potentially avoiding dependency issues. (`pkg/burner/job.go`, [[1]](diffhunk://#diff-bc606f2fd54d740859086073970bd68ea15b70016af24d3ac7a446741932b1edL493-R493) [[2]](diffhunk://#diff-bc606f2fd54d740859086073970bd68ea15b70016af24d3ac7a446741932b1edR524-R536)

With this change when using the default deletion strategy the sequence is:

1. gc namespaces
2. gc objects with predefined ns
3. gc cluster-scoped-resources

With the gvr strategy is:
1. gc objects in namespaces
2.  gc namespaces
3. gc objects with predefined ns
3. gc cluster-scoped-resources

closes: #1260


vishnuchalla (approved):
lgtm

PR #1285, opened 2026-07-31
Pprof replicas

rsevilla87:
This pull request introduces a new `replicas` field for pprof target configuration, allowing users to limit the number of pod instances from which profiling data is collected. T

**Key changes:**

#### Feature: Limit pprof collection with `replicas`

* Added an optional `replicas` field to the `PProftarget` struct in `types.go`, allowing users to specify how many matching pods to collect profiling data from; `0` (default) means all pods are included.
* Updated the logic in `pprof.go` to sort matching pods by name and select only the first N pods according to the `replicas` value, logging when limiting is applied. [[1]](diffhunk://#diff-8f29187dcd65305ccf3c06a9baaa4aa9635fa84b159631ea17c7052b73f1dd74R160) [[2]](diffhunk://#diff-8f29187dcd65305ccf3c06a9baaa4aa9635fa84b159631ea17c7052b73f1dd74L168-R186)
* Updated documentation in `docs/measurements/index.md` to describe the new `replicas` field, including usage notes and configuration examples for different components. [[1]](diffhunk://#diff-2253f73e9477212fbf76e586b07956f9e70b15c804dca04c0014dff104f1e8e6R1023-R1024) [[2]](diffhunk://#diff-2253f73e9477212fbf76e586b07956f9e70b15c804dca04c0014dff104f1e8e6R1047-R1054) [[3]](diffhunk://#diff-2253f73e9477212fbf76e586b07956f9e70b15c804dca04c0014dff104f1e8e6R1079)
* Added `replicas` to example measurement configurations, including `kube-burner-measure.yml`, to demonstrate limiting collection to a subset of pods.

### DIFF
--- a/pkg/measurements/pprof.go
+++ b/pkg/measurements/pprof.go
@@ -159,8 +159,7 @@ func (p *pprof) copyCerts() error {
 func (p *pprof) getPods(target types.PProftarget) ([]corev1.Pod, error) {
 	var pods []corev1.Pod
 	// If DaemonSet is deployed and no explicit label selector is provided, use DaemonSet pods
-	pprofNodeTargets := p.getPprofNodeTargets(target)
-	if pprofNodeTargets != nil { // Node target
+	if pprofNodeTargets := p.getPprofNodeTargets(target); pprofNodeTargets != nil { // Node target
 		labelSelector := labels.Set(pprofNodeTargets).String()
 		podList, err := p.ClientSet.CoreV1().Pods(types.PprofNamespace).List(context.TODO(),
 			metav1.ListOptions{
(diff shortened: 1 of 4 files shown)


rsevilla87:
cc @Sandeepyadav93 @venkataanil

rsevilla87:
PTAL when you get time @vishnuchalla

pkg/measurements/pprof.go
```diff
@@ -156,23 +157,39 @@ func (p *pprof) copyCerts() error {
...
+		}
+		pods = podList.Items
 	}
-	labelSelector := labels.Set(target.LabelSelector).String()
-	podList, err := p.ClientSet.CoreV1().Pods(target.Namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: labelSelector})
-	if err != nil {
-		return []corev1.Pod{}, fmt.Errorf("error listing pods labeled with %s: %v", labelSelector, err)
+	if target.Replicas > 0 && target.Replicas < len(pods) {
+		log.Debugf("Limiting %s pprof collection to %d/%d instances", target.Name, target.Replicas, len(pods))
+		sort.Slice(pods, func(i, j int) bool {
```
vishnuchalla, line 187: Why do we want to sort the pods based on name here? Can't we just pick a random on in the list.

rsevilla87: We need to sort them to ensure we always pick the same pods for each interval

vishnuchalla (approved):
lgtm

PR #1286, opened 2026-07-31
Add tokenFile support for Prometheus token refresh

Sandeepyadav93:
## Description

The Prometheus bearer token is set once at startup and never refreshed. For Some providers like GKE/Azure - long-running workloads (>1 hour), the token can expire mid-run causing 401 errors. This adds a tokenFile config field that points to a file re-read on every Prometheus HTTP request, allowing an external process to rotate the token without restarting kube-burner.

- Add tokenFile field to MetricsEndpoint config and Auth struct
- Define PrometheusClient interface (Query/QueryRange) so the Client field can hold either the go-commons static-token client or the new token-file-aware client
- Add tokenFileTransport RoundTripper that reads the token from disk on each request
- Add --token-file CLI flag to index and check-alerts commands
- Wire tokenFile through workloads helper for kube-burner-ocp consumers
- Add unit tests for tokenFileTransport

Explanation of how it works:-

1. kube-burner (pkg/prometheus/prometheus.go) calls prometheus.NewClientWithTokenFile(...) — that's a one-time setup
2. go-commons (prometheus/prometheus.go) stores the tokenFile path in the authTransport struct
3. go-commons authTransport.RoundTrip() — this is the key. It implements http.RoundTripper, which means it runs on every HTTP request the Prometheus client makes

Every time kube-burner does a Query() or QueryRange(), Go's HTTP client calls RoundTrip(), which does:

  if bat.tokenFile != "" {
      tokenBytes, err := os.ReadFile(bat.tokenFile)
      token := strings.TrimSpace(string(tokenBytes))
      req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
  }

  So the file is read fresh on every single Prometheus API call — not cached. That's how the external refresh script's updated token gets picked up automatically.

## Type of change

- New feature

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.25.0
 require (
 	dario.cat/mergo v1.0.1
 	github.com/Masterminds/sprig/v3 v3.3.0
-	github.com/cloud-bulldozer/go-commons/v2 v2.3.2
+	github.com/cloud-bulldozer/go-commons/v2 v2.3.8
 	github.com/google/uuid v1.6.0
 	github.com/itchyny/gojq v0.12.16
 	github.com/kubernetes-csi/external-snapshotter/client/v4 v4.2.0
--- a/go.sum
+++ b/go.sum
@@ -106,8 +106,8 @@ github.com/chzyer/readline v1.5.1/go.mod h1:Eh+b79XXUwfKfcPLepksvw2tcLE/Ct21YObk
 github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1/go.mod h1:Q3SI9o4m/ZMnBNeIyt5eFwwo7qiLfzFZmjNmxjkiQlU=
 github.com/chzyer/test v1.0.0/go.mod h1:2JlltgoNkt4TW/z9V/IzDdFaMTM2JPIi26O1pF38GC8=
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
-github.com/cloud-bulldozer/go-commons/v2 v2.3.2 h1:Lw6N4tlp+5oIgOQ5Uk4uQLSpXRqK7i2nPEQYmhjypR0=
-github.com/cloud-bulldozer/go-commons/v2 v2.3.2/go.mod h1:UJt/nJdrpuSQjWNQyou1OmDP1DEFHZx7frNxnNawloY=
+github.com/cloud-bulldozer/go-commons/v2 v2.3.8 h1:q9HdNFTnlbCdA6/y6UGd1jg0HImd6j8K4y9CFFrzQsg=
+github.com/cloud-bulldozer/go-commons/v2 v2.3.8/go.mod h1:UJt/nJdrpuSQjWNQyou1OmDP1DEFHZx7frNxnNawloY=
 github.com/cncf/udpa/go v0.0.0-20191209042840-269d4d468f6f/go.mod h1:M8M6+tZqaGXZJjfX53e64911xZQV5JYwmTeXPW+k8Sc=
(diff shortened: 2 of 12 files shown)


pkg/prometheus/prometheus.go
```diff
@@ -27,11 +32,80 @@ import (
...
+	return v, err
+}
+
+func (c *tokenFilePrometheusClient) QueryRange(query string, start, end time.Time, step time.Duration) (model.Value, error) {
+	r := apiv1.Range{Start: start, End: end, Step: step}
+	v, _, err := c.api.QueryRange(context.TODO(), query, r)
+	return v, err
+}
+
+func newTokenFileClient(url, tokenFile, username, password string, tlsSkipVerify bool) (PrometheusClient, error) {
```
rsevilla87: SImilar to the original prometheus implementation I think we should move this code to go-commons

Sandeepyadav93: Thanks Raul for you suggestion, I will iterate on it.

Sandeepyadav93: Trying with https://github.com/cloud-bulldozer/go-commons/pull/71

go.mod
```diff
@@ -158,3 +158,5 @@ replace (
 	k8s.io/client-go => k8s.io/client-go v0.34.3
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20250710124328-f3f2b991d03b
 )
+
+replace github.com/cloud-bulldozer/go-commons/v2 => github.com/Sandeepyadav93/go-commons/v2 v2.0.0-20260803121945-a7240b4a1818
```
Sandeepyadav93: These changes are for test only till https://github.com/cloud-bulldozer/go-commons/pull/71 are merged

go.sum
```diff
@@ -71,6 +71,8 @@ github.com/Microsoft/go-winio v0.6.1/go.mod h1:LRdKpFKfdobln8UmuiYcKPot9D2v6svN5
 github.com/NYTimes/gziphandler v1.1.1/go.mod h1:n/CVRwUEOgIxrgPvAQhUUr9oeUtvrhMomdKFjzJNB0c=
 github.com/PuerkitoBio/purell v1.1.1/go.mod h1:c11w/QuzBsJSee3cPx9rAFu61PvFxuPbtSwDGJws/X0=
 github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578/go.mod h1:uGdkoq3SwY9Y+13GIhn11/XLaGBb4BfwItxLd5jeuXE=
+github.com/Sandeepyadav93/go-commons/v2 v2.0.0-20260803121945-a7240b4a1818 h1:fiiVXugHgifKrINQtzZZwlxHCQfaMiBDmP+JjX/pwYg=
```
Sandeepyadav93: These changes are for test only till https://github.com/cloud-bulldozer/go-commons/pull/71 are merged

pkg/prometheus/prometheus.go
Sandeepyadav93: @rsevilla87 Thanks for the suggestion! I've moved the token file logic to go-commons and opened a PR there: https://github.com/cloud-bulldozer/go-commons/pull/71

This PR is now simplified to consume NewClientWithTokenFile from go-commons instead of implementing custom transport/client types here.

Keeping this in draft since we need the go-commons PR to merge first — currently using my fork via a replace directive in go.mod for testing, which will be reverted once go-commons is released.

Also added documentation for the new tokenFile field and --token-file CLI flag.

Could you please take another look when you get a chance?

Sandeepyadav93:
@vishnuchalla PTAL, I don't think the failing job test-k8s v1.34.8 is related to my PR, I am seeing the same failure on some other PRs.

vishnuchalla (approved):
lgtm

PR #1291, opened 2026-08-04
Use bash rather than sh for hooks

rsevilla87:
/bin/sh on Ubuntu 24 is dash, not bash

* Changed the command used in `prepareCommand` in `pkg/burner/hooks.go` to use `/bin/bash` instead of `/bin/sh` for running embedded scripts, improving compatibility with Bash-specific syntax.

Should fix CI issues observed at https://github.com/kube-burner/kube-burner-ocp/pull/480

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -540,7 +540,7 @@ The `when` field specifies at which stage the hook should execute:
 
 #### Embedded Script Support
 
-When using kube-burner-ocp or any application with an embedded filesystem, hooks can reference scripts stored in the embedded `scripts` directory.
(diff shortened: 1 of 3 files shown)


docs/reference/configuration.md
```diff
@@ -566,28 +568,28 @@ jobs:
     jobType: create
     jobIterations: 100
     namespace: workload-ns
-    
+
     hooks:
       # Background monitoring hook - runs throughout deployment
-      - cmd: ["/bin/bash", "/scripts/monitor-resources.sh"]
+      - cmd: ["/bin/bash", "/usr/local/bin/monitor-resources.sh"]
```
vishnuchalla, line 574: Is this change expected?

rsevilla87: yeah, it replaces the reference to scripts by another path to avoid confusion, the scripts directory is configurable, I noted it in another commit.

vishnuchalla (approved):
lgtm

PR #1295, opened 2026-08-19
Avoid etcd defrag during incremental load test

venkataanil:
etcd defragmentation during load testing is causing etcd leader elections and bringing down the api server. All other components like kubelet, crio and ovn are down due to this and CVO is reporting as unstable cluster. So we disable auto defrag before running the incremental load and manualy defrag before each incremental step.

## Type of change

- Refactor
- New feature
- Bug fix
- Optimization
- Documentation
- CI

- Related Issue #
- Closes #

### DIFF
--- a/pkg/burner/incremental.go
+++ b/pkg/burner/incremental.go
@@ -175,6 +175,37 @@ func (ex *JobExecutor) RunIncrementalCreateJob(
 	originalRunID := ex.runid
 	originalIterations := ex.JobIterations
 
+	// Initialize etcd defragmentation manager if enabled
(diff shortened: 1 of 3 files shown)


afcollins:
I'm not sure about this. If defrag is disrupting a workload, I think the fragile part of the workload that gets disrupted is the bug.

Also, etcd-operator is openshift specific: `Group:    "operator.openshift.io"`

vishnuchalla:
@venkataanil I believe this is just an experiment as you mentioned in one of our meetings. Should we consider this to be reviewed?

PR #1300, opened 2026-08-30
Restoring beforeChurn implementation

vishnuchalla:
## Type of change

- Refactor
- New feature
- Bug fix
- Optimization
- Documentation
- CI

## Description
Restore batch churn implementation as it supports workload than need chaos to be created during churning a workload.

This will be useful for special case where we have a lot of chaos going on the cluster during the churn phase.

## Testing
Tested and verified in local in integration with ocp-wrapper.
[kube-burner-ocp-6deb8d11-3af4-4aad-877b-47e479b759e1.log](https://github.com/user-attachments/files/31614863/kube-burner-ocp-6deb8d11-3af4-4aad-877b-47e479b759e1.log)

Able to reproduce spiky behavior in api-server thread usage
Screenshot 2026-08-30 at 10 18 23 AM

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -509,6 +509,7 @@ The `when` field specifies at which stage the hook should execute:
 | `beforeJobExecution`     | Before the job starts creating or processing objects                                                 |
 | `onEachIteration`        | At the start of each iteration of a create job                                                       |
 | `afterJobExecution`      | After the job finishes creating or processing objects (before churn, if enabled)                     |
+| `beforeChurn`            | Before churn starts on create jobs with churn enabled                                                |
 | `afterChurn`             | After churn completes on create jobs with churn enabled                                              |
 | `beforeCleanup`          | After the job finishes, before job pause and optional garbage collection                             |
 | `afterCleanup`           | After per-job garbage collection (`gc: true`), or after global GC when `gcMetrics` is enabled        |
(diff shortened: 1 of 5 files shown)


vishnuchalla:
cc: @sdodson

rsevilla87 (changes requested):
One of the reasons of not having this hook is that it's redundant with the `afterJobExecution` stage, this hook happens just between the regular job execution and the churning stage.

vishnuchalla:
> One of the reasons of not having this hook is that it's redundant with the `afterJobExecution` stage, this hook happens just between the regular job execution and the churning stage.

Yes, but there is a error check [here](https://github.com/vishnuchalla/kube-burner/blob/main/pkg/burner/job.go#L180-L182) that might skip churn phase. Also in case of churn jobs,  should afterJob hook run after the churn and object verification is completed? 🤔 . Some where around [here](https://github.com/vishnuchalla/kube-burner/blob/main/pkg/burner/job.go#L203)

rsevilla87:
> > One of the reasons of not having this hook is that it's redundant with the `afterJobExecution` stage, this hook happens just between the regular job execution and the churning stage.
> 
> Yes, but there is a error check [here](https://github.com/vishnuchalla/kube-burner/blob/main/pkg/burner/job.go#L180-L182) that might skip churn phase. 

That makes sense IMO, if the job times out in its regular stage, we should abort the execution regardless what comes next.

>Also in case of churn jobs, should afterJob hook run after the churn and object verification is completed? 🤔 . Some where around [here](https://github.com/vishnuchalla/kube-burner/blob/main/pkg/burner/job.go#L203)

We have the `afterChurn` hook for that, it doesn't include object verification though.

rsevilla87 (approved):

PR #1303, opened 2026-09-01
Add outputFile support for background hooks

rsevilla87:
Adds an outputFile field to Hook that redirects stdout/stderr of a background hook to a file


vishnuchalla (approved):
lgtm

PR #1309, opened 2026-09-22
Add baselinePhase option to vmiLatency measurement

arpsharm:
## Summary

- Adds an opt-in `baselinePhase` field to the `Measurement` config struct
- When set to `"vmiPending"`, `normalizeMetrics()` computes all phase latencies relative to the VMI Pending timestamp instead of `VM.CreationTimestamp`
- Workloads that don't set `baselinePhase` are completely unaffected

## Problem

Workloads using `runStrategy: Halted` with `dataVolumeTemplates` (e.g. windows-bootstorm) create the VM, clone the disk (~320s), then call `virtctl start`. Since `vmiLatency` uses `VM.CreationTimestamp` as the baseline, `VMReadyLatency` = clone time + boot time (~340s) — measuring storage performance, not boot performance. Early phases like `VMIPending` produce NaN.

## Solution

\`\`\`yaml
measurements:
  - name: vmiLatency
    baselinePhase: vmiPending
\`\`\`

All `.Sub(m.Timestamp)` calls become `.Sub(baseline)`, where `baseline` is `m.vmiPending` when configured.

## Test results

69 VMs, `runStrategy: Halted`, DV cloning, 8+ runs on bare metal:

| Metric | Before (default) | After (vmiPending) |
|---|---|---|
| VMReady P99 | 353,000 ms | 19,798–22,073 ms |
| VMIRunning P99 | NaN | 20,500–22,000 ms |
| VMIPending P99 | NaN | 0 ms |

## Files changed

- `pkg/measurements/types/types.go` — new `BaselinePhase` field
- `pkg/measurements/vmi_latency.go` — baseline logic in `normalizeMetrics()`


vishnuchalla (approved):
lgtm

